PR #39, opened 2025-02-28
fix(duckdb): try fix typecheck

LemonNekoGH:


packages/drizzle-duckdb-wasm/src/driver.ts
```diff
@@ -123,6 +123,9 @@ export function drizzle<
     if (typeof connection === 'string')
       return constructByDSN(connection, drizzleConfig) as any
 
+    if (typeof connection === 'undefined')
+      throw new Error('connection option is required')
+
```
LemonNekoGH, line 128: Can I throw Error here?

nekomeowww (approved):

packages/drizzle-duckdb-wasm/src/driver.ts
sumimakito: Yes, I think so

nekomeowww: Yes.

sumimakito (approved):

PR #41, opened 2025-03-02
feat(stage-web): load and save live2d model

LemonNekoGH:
## Description

Now it can load and save other live2D model, but blink and motions broken, I will fix it soon.

## Additional context

Can I use pinia to let data cross `apps/stage-web/src/components/Widgets/Live2DSettings.vue` and `packages/stage-ui/src/components/Live2D/Model.vue`? Or should I use props and emit to do this?

### DIFF
--- a/packages/stage-ui/src/utils/live2d-zip-loader.ts
+++ b/packages/stage-ui/src/utils/live2d-zip-loader.ts
@@ -71,7 +71,7 @@ function createFakeSettings(files: string[]): ModelSettings {
   settings.name = modelName;
 
   // provide this property for FileLoader
-  (settings as any)._objectURL = `DontTouchMe://${settings.url}`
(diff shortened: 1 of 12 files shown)


packages/stage-ui/src/utils/live2d-zip-loader.ts
```diff
@@ -1,8 +1,81 @@
...
+            '': motions.map(motion => ({ File: motion })),
+          }
+        : undefined,
+    },
+  })
+
+  settings.name = modelName;
+
+  // provide this property for FileLoader
+  (settings as any)._objectURL = `DontTouchMe://${settings.url}`
```
nekomeowww:
```suggestion
  (settings as any)._objectURL = `example://${settings.url}`
```

nekomeowww (approved):

PR #50, opened 2025-03-07
refactor!: use xsschema/ext-providers instead of typeschema/providers

kwaa:


apps/stage-web/src/utils/jsonFormat.ts
```diff
@@ -39,24 +39,8 @@ ${JSON.stringify(await toJSONSchema(schema))}`))
...
-
-Validation errors:
-${validation.issues.map(issue => `- ${issue.message}`).join('\n')}
-
-Please provide a corrected response that matches the schema:
-${JSON.stringify(await toJSONSchema(schema))}`))
-
-    const response = await call(options, schema)
-    return parseJSONFormat(response, options, schema, JSON.stringify(parsedContent), validation.issues.map(i => i.message).join(', '))
+    // TODO: print validation issues
```
kwaa, line 42: xsschema doesn't return `issues`, so I added a TODO here.

nekomeowww: ok

nekomeowww (approved):

PR #63, opened 2025-03-09
chore(stage-web): make the live2d shadow also follow theme color

LemonNekoGH:


apps/stage-web/src/styles/themes.css
```diff
@@ -10,5 +10,17 @@
   --airi-theme-primary-800: #a70a3e;
   --airi-theme-primary-900: #8e0d3b;
   --airi-theme-primary-950: #50011b;
+
   --theme-colors-hue: 354.31;
+  --theme-colors-chroma: calc(0.18 + (cos(var(--theme-colors-hue) * 3.14159265 / 180) * 0.04));
```
junkwarrior87, line 15: i think it can use `* 1deg` here (in replacement of `* 3.14 / 180`)

LemonNekoGH: No it will make color too bright

junkwarrior87: hmm.. i tried locally, didn't see any difference

junkwarrior87: > hmm.. i tried locally, didn't see any difference

seems related to monitor.. weird, technically they're equivalent

sumimakito (approved):
Tested and LGTM.

OT: Do we also want to make the two buttons below follow the primary color?

Screenshot

junkwarrior87:
> Tested and LGTM.
> 
> OT: Do we also want to make the two buttons below follow the primary color?
> 
> Screenshot

I've opened https://github.com/moeru-ai/airi/pull/67 regarding this

PR #74, opened 2025-03-12
feat(stage-web): theme presets

luoling8192:
CleanShot 2025-03-13 at 01 11 05@2x


LemonNekoGH (changes requested):
You can use [culori.js](https://culorijs.org/) to convert color spaces

luoling8192:
> You can use [culori.js](https://culorijs.org/) to convert color spaces

Thank you! I have used it.

PR #87, opened 2025-03-19
fix(stage-ui): extract colors

junkwarrior87:
make the buttons cursor-pointer, and also fixed the issue that the extracted color does not take effect when RGB is on

did some refactoring for some common components (partially, will continue on following PRs)


nekomeowww (approved):
Thanks.

PR #96, opened 2025-03-25
style(stage-web): fix word break

LemonNekoGH:


apps/stage-web/src/components/Widgets/ChatHistory.vue
```diff
@@ -77,11 +77,17 @@ onTokenLiteral(async () => {
...
+            <div v-if="message.content" class="markdown-content break-words" text="base <sm:xs" v-html="process(message.content as string)" />
             <div v-else />
           </div>
         </div>
       </div>
     </div>
   </div>
 </template>
+
+<!-- <style scoped>
```
luoling8192: If it is unnecessary, should we remove it?

LemonNekoGH: 👌 removed

PR #97, opened 2025-03-25
feat(package): @moeru-ai/ccc

luoling8192:

PR #98, opened 2025-03-25
feat: airi card

luoling8192:
Depends #97

CleanShot 2025-03-30 at 03 04 04@2x


packages/stage-ui/src/stores/modules/airi-card.ts
```diff
@@ -0,0 +1,242 @@
...
+      agents: existingExtension.agents ?? {},
+    }
+  }
+
+  function newAiriCard(card: Card | ccv3.CharacterCardV3): AiriCard {
+    // Handle ccv3 format if needed
+    if ('data' in card) {
+      const ccv3Card = card as ccv3.CharacterCardV3
+      return {
+        name: ccv3Card.data.name,
```
luoling8192, line 137: It can move to ccc later

packages/stage-ui/src/stores/chat.ts
```diff
@@ -63,10 +62,10 @@ export const useChatStore = defineStore('chat', () => {
   }
 
   const messages = ref<Array<Message | ErrorMessage>>([
-    SystemPromptV2(
-      t('prompt.prefix'),
-      t('prompt.suffix'),
-    ),
+    {
+      role: 'system',
+      content: systemPrompt.value.replace(/\{\{user\}\}/g, 'user'),
```
nekomeowww, line 67: 想想模板字符串能不能有库通用化了吧...

nekomeowww (approved):

PR #100, opened 2025-03-26
feat(stage-web): new settings animation

LemonNekoGH:


apps/stage-web/src/components/IconAnimation.vue
```diff
@@ -0,0 +1,75 @@
...
+  return isAnimating.value ? props.isReverse ? props.iconSize : 25 : props.isReverse ? 25 : props.iconSize
+})
+
+const textColor = computed(() => {
+  return isAnimating.value ? props.isReverse ? props.textColor : 'text-white' : props.isReverse ? 'text-white' : props.textColor
+})
+</script>
+
+<template>
+  <div pointer-events-none fixed left-0 top-0 h-full w-full>
```
junkwarrior87: ~1. add `z--1` back (you can see the visual differences in /modules page, etc.)~
  update: i know what's the problem, the transition should be on top of the UI but the final icon should not, so applying z--1 here will be wrong. let's leave this first, i'll fix it on further PRs
2. `left-0 top-0 h-full w-full` --> `inset-0`
3. `pointer-events-none` inherits, so just adding for the parent div will be ok, no need to add them for all children

apps/stage-web/src/pages/settings/appearance/index.vue
```diff
@@ -142,9 +145,16 @@ const dark = useDark()
...
-    <div text="40" i-lucide:paintbrush />
-  </div>
+  <IconAnimation
+    v-if="showAnimationComponent"
+    :duration="1000"
+    :started="showIconAnimation"
+    :is-reverse="true"
+    :icon="animationIcon"
+    :icon-size="12"
+    position="calc(100vw - 9.5rem), calc(100vh - 9.5rem)"
```
junkwarrior87: use `100dvh` instead (same as all below)

apps/stage-web/src/components/IconAnimation.vue
```diff
@@ -0,0 +1,75 @@
...
+const opacity = computed(() => {
+  return isAnimating.value ? props.isReverse ? 0 : 1 : props.isReverse ? 1 : 0
+})
+
+const size = computed(() => {
+  return isAnimating.value ? props.isReverse ? props.iconSize : 25 : props.isReverse ? 25 : props.iconSize
+})
+
+const textColor = computed(() => {
+  return isAnimating.value ? props.isReverse ? props.textColor : 'text-white' : props.isReverse ? 'text-white' : props.textColor
```
junkwarrior87: this is too lengthy (and you'll need to repeat the values every time), consider using `xor`?
e.g.
```js
return (isAnimating.value ^ props.isReverse) ? 'text-white' : props.textColor
```

additionally, you might consider using this way:
```js
const from = { position: ..., opacity: ..., ... }
const to = { ... }

const cssProps = computed(() => (isAnimating.value ^ props.isReverse) ? from : to)
```
to further reduce repeated calculations

apps/stage-web/src/pages/settings/index.vue
```diff
@@ -1,8 +1,92 @@
...
 
+import IconAnimation from '../../components/IconAnimation.vue'
+
 const router = useRouter()
+const showIconAnimation = ref(false)
+const iconAnimation = ref<InstanceType<typeof IconAnimation>>()
+const resolveAnimation = ref<() => void>()
+const { t } = useI18n()
+
+// 添加动画相关的状态
```
nekomeowww: Try not to use Chinese as comments as much as possible.

apps/stage-web/src/pages/settings/index.vue
```diff
@@ -1,8 +1,92 @@
...
+
+const removeBeforeEach = router.beforeEach(async (_, __, next) => {
+  await new Promise<void>((resolve) => {
+    resolveAnimation.value = resolve
+  })
+  removeBeforeEach()
+  next()
+})
+
+const settings = [
```
nekomeowww: Since `t` is involved, use `computed` here.

apps/stage-web/src/components/IconAnimation.vue
```diff
@@ -0,0 +1,75 @@
...
+  if (newVal) {
+    requestAnimationFrame(() => {
+      isAnimating.value = true
+    })
+  }
+})
+
+const centerPosition = computed(() => {
+  const size = props.started ? 25 : props.iconSize
+  return `calc(50vw - ${size / 2}rem), calc(50vh - ${size / 2}rem)`
```
nekomeowww: Should `vw` to `dvw`?

apps/stage-web/src/pages/settings/index.vue
junkwarrior87: or move t inside the component props

packages/stage-ui/src/stores/settings.ts
```diff
@@ -29,6 +29,7 @@ export const useSettings = defineStore('settings', () => {
   const live2dMotionMap = useLocalStorage<Record<string, string>>('settings/live2d/motion-map', {})
 
   const disableTransitions = useLocalStorage('settings/disable-transitions', true)
+  const useIconAnimation = useLocalStorage('settings/use-icon-animation', true)
```
junkwarrior87: use `disableIconAnimation` to align with the existing code logic, otherwise you'll see on the appearance page that the icon behaviors are flipped

apps/stage-web/src/pages/settings/providers/index.vue
```diff
@@ -50,9 +59,16 @@ const { allProvidersMetadata } = storeToRefs(providersStore)
       :configured="provider.configured"
     />
   </div>
-  <div text="neutral-200/50 dark:neutral-600/20" pointer-events-none fixed bottom-0 right-0 z--1 translate-x-10 translate-y-10>
-    <div text="40" i-lucide:brain />
-  </div>
+  <IconAnimation
+    v-if="showAnimationComponent && settingsStore.useIconAnimation"
```
junkwarrior87: cannot directly do this (i.e. disable the whole icon component when disabling the setting),
otherwise you'll realize that when disabling icon transitions, the lower-right icon just simply disappeared

nekomeowww (approved):

PR #125, opened 2025-04-10
fix(stage-tamagotchi): app icon

typed-sigterm:
## Description

As the title says.


LemonNekoGH (approved):

LemonNekoGH:
We can also change the tauri identifier to `ai.moeru.airi-tamagotchi`

typed-sigterm:
Note that changing identifier will cause the old and new versions are considered as two separate apps, need to be explained in the changelog.

nekomeowww:
> Note that changing identifier will cause the old and new versions are considered as two separate apps, need to be explained in the changelog.

No problem, can definitely specify this out when rolling out next release. Thanks for noticing us!

apps/stage-tamagotchi/src-tauri/icons/ios/AppIcon-20x20@1x.png
kwaa, line 1: Many low-res icons here, and I don't think modern devices use that size. Is it possible to remove them?

apps/stage-tamagotchi/src-tauri/icons/android/mipmap-hdpi/ic_launcher.png
kwaa, line 1: same here

apps/stage-tamagotchi/src-tauri/icons/ios/AppIcon-20x20@1x.png
typed-sigterm: These sizes are (now or in the past) required by Apple to be included, and small-sized icons will be used for notifications, settings, or tabbars. The current specification is [here](https://developer.apple.com/design/human-interface-guidelines/app-icons). If compatibility is not a concern, it's okay to delete them.

kwaa: > These sizes are (now or in the past) required by Apple to be included, and small-sized icons will be used for notifications, settings, or tabbars. The current specification is [here](https://developer.apple.com/design/human-interface-guidelines/app-icons). If compatibility is not a concern, it's okay to delete them.

https://developer.apple.com/design/human-interface-guidelines/app-icons#iOS-iPadOS-app-icon-sizes

120, 180, 167, 152, 80, 58, 87, 76, 114

but——

> You can let the system automatically scale down your 1024x1024 px app icon to produce all other sizes, or — if you want to customize the appearance of the icon at specific sizes — you can supply multiple versions such as the following.

So just providing a 1024x1024 icon should be fine.

nekomeowww (approved):

PR #177, opened 2025-05-26
fix(ui-transitions): build

LemonNekoGH:


packages/ui-transitions/package.json
```diff
@@ -37,6 +37,7 @@
     "@unocss/reset": "^66.1.2",
     "@vitejs/plugin-vue": "^5.2.4",
     "unplugin-vue-router": "^0.12.0",
+    "vite": "^6.3.5",
```
nekomeowww, line 40: Wait, why? Shouldnt we remove this?

nekomeowww (approved):

PR #186, opened 2025-05-30
feat(stage-tamagotchi): allowing click through and fade away when hovering (currently macOS only)

nekomeowww:
## Description

As demo

https://github.com/user-attachments/assets/9f51c99c-f7ab-4653-9022-62ac001a1d9b

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,12 +95,18 @@ name = "app"
 version = "0.4.27"
 dependencies = [
  "log",
+ "objc2 0.6.1",
+ "objc2-foundation 0.3.1",
+ "serde",
+ "serde_json",
  "tauri",
  "tauri-build",
  "tauri-plugin-log",
  "tauri-plugin-mcp",
  "tauri-plugin-os",
  "tauri-plugin-prevent-default",
+ "tokio",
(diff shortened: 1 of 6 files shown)


apps/stage-tamagotchi/src/composables/window-shortcuts.ts
```diff
@@ -53,4 +68,47 @@ export function useWindowShortcuts() {
...
+    }
+  }
+
+  function handleMouseMove(event: MouseEvent) {
+    window.mouseX = event.clientX
+    window.mouseY = event.clientY
+  }
+
+  onMounted(() => {
+    window.addEventListener('keydown', handleKeyDown)
```
LemonNekoGH: Maybe we can use `vueuse`?

nekomeowww: Ok. And wait, it seems I fotgot to delete some of the code here, typecheck will fail.

nekomeowww: Thanks for pointing out. The change to this file wasn't necessary

PR #187, opened 2025-05-30
feat(stage-tamagotchi): add click through support (previous macOS only, now both supported)

nekomeowww:
https://github.com/user-attachments/assets/2618b1ed-2f8c-4142-ba3d-93528f95e105

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,9 +23,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      # Rust
-      - uses: dtolnay/rust-toolchain@nightly
-      - uses: mozilla-actions/sccache-action@v0.0.3
+      # # Rust
+      # - uses: dtolnay/rust-toolchain@nightly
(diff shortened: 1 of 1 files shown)


apps/stage-tamagotchi/src-tauri/src/lib.rs
```diff
@@ -180,11 +60,26 @@ async fn start_monitor_for_clicking_through(window: tauri::Window) -> Result<(),
...
+        // Only allow disabling click-through when:
+        // 1. Cursor is OUTSIDE the window AND
+        // 2. Modifier key is pressed
+        let should_be_click_through = cursor_inside && !modifier_pressed;
+
+        if should_be_click_through {
+          let _ = set_cursor_inside(&window, true);
+        } else {
+          let _ = set_cursor_inside(&window, false);
+        }
```
sumimakito: nit: Maybe we can make this more compact

```suggestion
        let _ = set_cursor_inside(&window, cursor_inside && !modifier_pressed); // `true` to enable clicking through
```

PR #193, opened 2025-06-03
"Proxy object could not be cloned." - Fix proposal

### LINKED ISSUE #192: "Proxy object could not be cloned." when sending more than 1 message
### Describe the bug

Hi !
I tried to install the project and run in on my own hardware, which seemed to work nicely, but I stumbled across an error when doing some testing.
When I send a first message, I get an answer but if I try to write a second message, I immediately get the error "**Proxy object could not be cloned**".

Here's a screenshot of this problem :
Image

### System Info

```Shell
lj50@lj50-Tour:~$ npx envinfo --system --binaries --browsers

  System:
    OS: Linux 6.8 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 12.68 GB / 30.50 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 18.19.1 - /usr/bin/node
    npm: 9.2.0 - /usr/bin/npm
    pnpm: 10.11.1 - /usr/local/bin/pnpm
```
Navigator is Firefox 138.0.4, May 17, 2025

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/CONTRIBUTING.md).

### Contributions

LJ5O:
Hi !
After some extensive debugging, I found the problem causing the error I encountered in #192 and was able to write a possible fix to it.

## Description

In https://github.com/moeru-ai/airi/blob/2dbeccc62c1d635856a444c0ac919902d578e38e/packages/stage-ui/src/stores/llm.ts#L21-L28, we are waiting for a `streamText` object, an answer from the LLM. This object is generated by the package `@xsai/stream-text`.

By looking at the code from this package, I can see the [following](https://github.com/moeru-ai/xsai/blob/3e165f3fd02a293e9a49685757a3f205d1521d62/packages/stream-text/src/index.ts#L165-L173) : 
```ts
const stepOne = async (options: StreamTextOptions): RecursivePromise<void> => {
    const step: StreamTextStep = {
      choices: [],
      finishReason: 'error',
      messages: structuredClone(options.messages),
      stepType: 'initial',
      toolCalls: [],
      toolResults: [],
    }
```

When instantiating a new `streamText`, `messages` are passed to `structuredClone()` function. Since `messages` is an object containing vue3 proxies somewhere, it fails ( *Proxy object could not be cloned* )

The solution I found is to use the function `toRaw()` provided by *vue* on every properties of `messages` using a convenient recursive function I added, `toRawObject()`.
That way, `structuredClone()` can be used without any trouble.

## Linked Issues

Fixes #192 

## Additional context

I think this is somewhat related to https://stackoverflow.com/questions/71075490/how-to-make-a-structuredclone-of-a-proxy-object

Before the fix : 
Capture d’écran du 2025-06-03 14-25-56
After : 
Capture d’écran du 2025-06-03 14-28-27

> Sorry if my explanations are not very easy to understand, not very used to write PR yet, but I would be happy to discuss further about this PR if that could help !

### DIFF
--- a/packages/stage-ui/src/stores/llm.ts
+++ b/packages/stage-ui/src/stores/llm.ts
@@ -6,6 +6,7 @@ import { streamText } from '@xsai/stream-text'
 import { defineStore } from 'pinia'
 
 import { debug, mcp } from '../tools'
+import { toRawObject } from '../utils'
 
 export const useLLM = defineStore('llm', () => {
   async function stream(model: string, chatProvider: ChatProvider, messages: Message[], options?: {
@@ -21,7 +22,7 @@ export const useLLM = defineStore('llm', () => {
     return await streamText({
       ...chatProvider.chat(model),
       maxSteps: 10,
-      messages,
+      messages: toRawObject(messages),
       headers,
       tools: [
         ...await mcp(),
--- a/packages/stage-ui/src/utils/index.ts
+++ b/packages/stage-ui/src/utils/index.ts
@@ -1,2 +1,3 @@
 export * from './eye-motions'
 export * from './iterator'
+export * from './toRawObject'
--- a/packages/stage-ui/src/utils/toRawObject.ts
+++ b/packages/stage-ui/src/utils/toRawObject.ts
@@ -1,0 +1,27 @@
+import { toRaw } from 'vue'
+
+export function toRawObject<T>(obj: T): T {
+  /*
+  This function can be used to convert an object into a Raw object. It will recussively convert all the values inside it.
+  The returned object will be a Raw object with all its compatible values converted into Raw objects.
+  */
+  if (typeof obj !== 'object' || obj === null) {
+    // Seems to be a primitive value or undefined
+    return obj
+  }
+
+  // Thanks to vue's toRaw function, we can only keep the basic values, removing useless Proxies
+  const raw = toRaw(obj)
+
+  if (Array.isArray(raw)) { // In case we are working with an array, stored data can be checked as well
+    return raw.map(item => toRawObject(item))
+  }
+
+  // We're not working on an array, so we can check the defined properties of the "custom" object
+  const cleanObj = {} as T
+  for (const key in raw) {
+    cleanObj[key] = toRawObject(raw[key])
+  }
+
+  return cleanObj // And voilà, a nice raw object that can be easily structuredClone()-ed
+}


LemonNekoGH:
🤔 We have already called `toRaw` in chat store:

https://github.com/moeru-ai/airi/blob/2dbeccc62c1d635856a444c0ac919902d578e38e/packages/stage-ui/src/stores/chat.ts#L136-L142

LJ5O:
Was it working on your side ? Maybe my setup is in cause 🤔

Well, I tried to run
```ts
      const newMessages = messages.value.slice(0, messages.value.length - 1).map((msg) => {
        if (msg.role === 'assistant') {
          const { slices: _, ...rest } = msg // exclude slices
          const a = toRaw(rest)
          return structuredClone(a) // FAILS HERE : Error sending message: DOMException: Proxy object could not be cloned.
        }
        const a = toRaw(msg)
        return structuredClone(a)
      })
```
Which didn't worked. `toRaw()` removes the `Proxy` only on the outer side of a given object, so maybe another one is hidden somewhere.

I played around, displaying the content of `rest` in the console, and found something interesting : **`rest.tool_results` is also a `Proxy`**, which is not removed by the first `toRaw` call. We have to call `toRaw` precisely on this property.

So, we just need to add the following : 
```ts
        if (msg.role === 'assistant') {
          const { slices: _, ...rest } = msg // exclude slices
          rest.tool_results = toRaw(rest.tool_results) // ADDED FIX
          return toRaw(rest)
```
I'll also revert what I did previously, this single line is more efficient that looping through the whole object

nekomeowww:
Thanks for submitting the issue along with the fixes. 

Let me take a look into this. Either `toRaw` and `toValues` will do the trick but I kind of unsure about which one fits and suits the best for this scenario.

LemonNekoGH (approved):
LGTM, thanks, I really forgot to call `toRaw` for tool call results.

nekomeowww:
Oh ok.

nekomeowww (approved):
Thanks.

PR #198, opened 2025-06-07
fix(unocss): add %alpha placeholder for complex color expressions

Menci:
## Description

UnoCSS's color-oriented class names like `bg-primary-100/60` could contain an color opacity like `60` to be applied to the color. It's implemented by adding an alpha part like `/ 0.6` to the color expressions.

When the expressions are complex like `color-mix()`, UnoCSS can't parse them and add the alpha part correctly. `/60` for those colors will have no effect (i.e. generates the same CSS rules as the class with no `/60` suffix).

The solution is adding `%alpha` placeholder for the color expressions for UnoCSS to replace directly.

### DIFF
--- a/uno.config.ts
+++ b/uno.config.ts
@@ -8,18 +8,18 @@ import { parseColor } from 'unocss/preset-mini'
 
 function createColorSchemeConfig(hueOffset = 0) {
   return {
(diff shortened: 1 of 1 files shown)


nekomeowww (approved):

LemonNekoGH:
COOL

Issue #137, opened 2025-04-23
stage-tamagotchi: error message also send to completions API


LemonNekoGH:
### Describe the bug

I got error message after I fixed settings issue `Provider credentials for  not found` but didn't refresh the stage page.

Error response:

```json
{
  "error": {
    "message": "Invalid value: 'error'. Supported values are: 'system', 'assistant', 'user', 'function', 'tool', and 'developer'.",
    "type": "invalid_request_error",
    "param": "messages[0].role",
    "code": "invalid_value"
  }
}
```

Request body:

```json
{
  "model": "openai/gpt-4o",
  "messages": [
    {
      "role": "error",
      "content": "Provider credentials for  not found"
    },
    {
      "role": "user",
      "content": "你好，随便说点什么"
    }
  ],
  "stream": true
}
```

Except request body:

```json
{
  "model": "openai/gpt-4o",
  "messages": [
    {
      "role": "user",
      "content": "你好，随便说点什么"
    }
  ],
  "stream": true
}
```

### System Info

```Shell
System:
    OS: macOS 15.4.1
    CPU: (14) arm64 Apple M4 Max
    Memory: 259.14 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.10.0 - ~/.asdf/installs/nodejs/23.10.0/bin/node
    npm: 10.9.2 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 9.12.3 - ~/.asdf/installs/nodejs/23.10.0/bin/pnpm
    bun: 1.1.34 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 135.0.7049.96
    Edge: 135.0.3179.85
    Safari: 18.4
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

PR #207, opened 2025-06-13
fix(stage) Player2 api typo, also replace error role with user role when calling api

lorisj:
## Description

1. fixed typo `config.baseURL -> config.baseUrl` for player2 provider

2. Next according to openai's specification you should not have an error role, (this makes the player2 api break). I can think of two ways to fix:
  - Drop all messages with error role before sending to api
  - Change the role of these messages to a user message, then add some more to the start to let the LLM know that this was an error message. (which is what this PR does)

### DIFF
--- a/packages/stage-ui/src/stores/llm.ts
+++ b/packages/stage-ui/src/stores/llm.ts
@@ -21,6 +21,7 @@ export const useLLM = defineStore('llm', () => {
     return await streamText({
       ...chatProvider.chat(model),
       maxSteps: 10,
+      // TODO: proper format for other error messages.
       messages: messages.map(msg => ({ ...msg, content: (msg.role as string === 'error' ? `User encountered error: ${msg.content}` : msg.content), role: (msg.role as string === 'error' ? 'user' : msg.role) } as Message)),
       headers,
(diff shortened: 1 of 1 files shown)


packages/stage-ui/src/stores/llm.ts
```diff
@@ -21,7 +21,7 @@ export const useLLM = defineStore('llm', () => {
     return await streamText({
       ...chatProvider.chat(model),
       maxSteps: 10,
-      messages,
+      messages: messages.map(msg => ({ ...msg, content: (msg.role as string === 'error' ? `User encountered error: ${msg.content}` : msg.content), role: (msg.role as string === 'error' ? 'user' : msg.role) } as Message)),
```
nekomeowww, line 25:
```suggestion
      // TODO: proper format for other error messages.
      messages: messages.map(msg => ({ ...msg, content: (msg.role as string === 'error' ? `User encountered error: ${msg.content}` : msg.content), role: (msg.role as string === 'error' ? 'user' : msg.role) } as Message)),
```

LemonNekoGH:
close #137

PR #214, opened 2025-06-19
perf(apps): use `rolldown-vite` instead of `vite`

kwaa:


nekomeowww:
It's ok, let me check.

PR #220, opened 2025-06-19
feat(stage) if no active provider and player2 is running locally, set provider to player2

lorisj:
## Description

- If there is no active provider, and player2 is configured, check if player2 is running locally. If it is, this sets it as the active provider.

- Also renamed `player2-api -> player2`

### DIFF
--- a/packages/stage-ui/src/stores/modules/consciousness.ts
+++ b/packages/stage-ui/src/stores/modules/consciousness.ts
@@ -65,7 +65,7 @@ export const useConsciousnessStore = defineStore('consciousness', () => {
     if (newProvider === 'player2') {
       player2Interval = window.setInterval(async () => {
         fetch('http://localhost:4315/v1/health').catch(() => {})
-      }, 6000)
+      }, 60000)
     }
     else {
       if (player2Interval)


packages/stage-ui/src/stores/modules/consciousness.ts
```diff
@@ -58,27 +58,70 @@ export const useConsciousnessStore = defineStore('consciousness', () => {
...
+              const res = await fetch('http://localhost:4315/v1/health')
+              if (res.ok) {
+                // eslint-disable-next-line no-console
+                console.log('No provider selected & player2 is running, setting player2 as provider')
+                if (player2Interval !== null) {
+                  clearInterval(player2Interval)
+                  player2Interval = null
+                }
+                activeProvider.value = 'player2'
+              }
```
nekomeowww, line 104: While auto-switching sounds ok, but `http://localhost:4315/v1/health` can be used for any programs (bidirectional verifications requires, for example, we could do small API calls to verify that `localhost:4315` is indeed the application called Player 2), including MDM applications that many companies are using, this introduces unexpected data leaks, may even trigger malicious code execution (tool use). 

I don't think that we should **implicitly** and **directly** switch to `player-2-api` as **default** provider. At least, please wait until we finished the new chat UI, where user can **explicitly** see a **active provider** selector for at least, noticing...

nekomeowww:
Rebase main.

nekomeowww:
I will close it for now since it's staled for quite a while, feel free to re-open this Pull Reuqest once finished.

PR #223, opened 2025-06-21
feat(stage-tamagotchi): allow visible on all workspace

LemonNekoGH:
https://github.com/user-attachments/assets/85322a28-d0a0-458f-ba4c-28fe9da7d8b7

### DIFF
--- a/packages/stage-ui/src/stores/settings.ts
+++ b/packages/stage-ui/src/stores/settings.ts
@@ -34,7 +34,7 @@ export const useSettings = defineStore('settings', () => {
   const themeColorsHue = useLocalStorage('settings/theme/colors/hue', DEFAULT_THEME_COLORS_HUE)
   const themeColorsHueDynamic = useLocalStorage('settings/theme/colors/hue-dynamic', false)
 
-  const allowVisibleOnAllWorkspaces = useLocalStorage('settings/allow-visible-on-all-workspaces', false)
+  const allowVisibleOnAllWorkspaces = useLocalStorage('settings/allow-visible-on-all-workspaces', true)
 
   function setThemeColorsHue(hue = DEFAULT_THEME_COLORS_HUE) {
     themeColorsHue.value = hue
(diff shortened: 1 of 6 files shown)


packages/stage-ui/src/stores/settings.ts
```diff
@@ -34,6 +34,8 @@ export const useSettings = defineStore('settings', () => {
   const themeColorsHue = useLocalStorage('settings/theme/colors/hue', DEFAULT_THEME_COLORS_HUE)
   const themeColorsHueDynamic = useLocalStorage('settings/theme/colors/hue-dynamic', false)
 
+  const allowVisibleOnAllWorkspaces = useLocalStorage('settings/allow-visible-on-all-workspaces', false)
```
nekomeowww: So can we make this `true` by default? If `false` is used by default, then what's the difference of manually setting AIRI to all workspaces from option menu?

nekomeowww:
Rest is LGTM.

packages/stage-ui/src/stores/settings.ts
nekomeowww:
```suggestion
  const allowVisibleOnAllWorkspaces = useLocalStorage('settings/allow-visible-on-all-workspaces', true)
```

LemonNekoGH: Nice catch! I forgot it.

LemonNekoGH: > So can we make this `true` by default? If `false` is used by default, then what's the difference of manually setting AIRI to all workspaces from option menu?

We don't have a docker icon, so you need to enable `Recent Apps` options in `System Preferences`, but I don't know should we provide an option to enable docker icon?

nekomeowww: You can. And for what we should do next, is to add more tutorials for users. But it's up to you, research the advanced part or making new tutorial.

PR #229, opened 2025-06-25
fix(settings): unify dark theme background color for content area and header

SliverKeigo:
## What’s Changed

- Unified the background color of the settings content area and PageHeader in dark mode to match the card  style for a more consistent visual experience.
- Fixed the issue where the PageHeader background color did not match the content/card area in dark theme.

## Visual Comparison

- Before
c0b05fe92e56b6017841e8d0ee70bc82

- After
ae1222df69ab9c9313f1027fd1504479

## Impact

- This change only affects the settings page in the desktop app (stage-tamagotchi) and does not impact other pages or features.

### DIFF
--- a/apps/stage-tamagotchi/src/layouts/settings.vue
+++ b/apps/stage-tamagotchi/src/layouts/settings.vue
@@ -138,6 +138,7 @@ const routeHeaderMetadata = computed(() => routeHeaderMetadataMap.value[route.pa
         <!-- Content -->
         <div flex="~ col" mx-auto max-w-screen-xl>
           <PageHeader
+            bg="bg-neutral-50 dark:bg-neutral-800"
             :title="routeHeaderMetadata?.title"
             :subtitle="routeHeaderMetadata?.subtitle"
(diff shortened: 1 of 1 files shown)


apps/stage-tamagotchi/src/layouts/settings.vue
```diff
@@ -138,6 +138,7 @@ const routeHeaderMetadata = computed(() => routeHeaderMetadataMap.value[route.pa
         <!-- Content -->
         <div flex="~ col" mx-auto max-w-screen-xl>
           <PageHeader
+            bg="bg-neutral-50 dark:bg-neutral-800"
```
LemonNekoGH: Do we still need "bg" prefix when using attributify mode?

```suggestion
            bg="neutral-50 dark:neutral-800"
```

SliverKeigo: Ah, got it. I wasn't aware of that, my knowledge on these modern CSS styling conventions is a bit limited. Sorry about that, and thanks for the clarification !

PR #282, opened 2025-07-16
debug(workspace): unocss fonts split shared testing

nekomeowww:


Weathercold:
oh you just merged it. I also just tested this PR and it does fix the build

PR #325, opened 2025-07-26
fix(docs): window not defined

nekomeowww:


sumimakito (approved):
Nice catch! Thank you

PR #330, opened 2025-07-28
fix(stage-ui): correctly chunk tokens streamed to tts pipeline

sumimakito:
## Description

Correctly pipe tokens received with a ReadableStream to the TTS chunking processor, and introduce a `TTS_FLUSH_INSTRUCTION` (currently the ZWSP character) to instruct the pipeline to flush, as the queue is long-lived (there is no obvious end). This is not the perfect solution, and we can come up with a better one later.


packages/stage-ui/src/composables/queues.ts
```diff
@@ -103,23 +101,24 @@ export function useDelayMessageQueue() {
...
-
-          processed.value = ''
-          return
-        }
-
-        for await (const chunk of chunkTTSInput(ctx.data)) {
-          await ttsQueue.add(chunk.text)
-        }
+        enqueue(encoder.encode(ctx.data))
       },
```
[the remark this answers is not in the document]
sumimakito: It's a good suggestion, but it seems `useQueue` only accepts async functions as handlers

packages/stage-ui/src/composables/queues.ts
```diff
@@ -103,23 +101,24 @@ export function useDelayMessageQueue() {
...
+    start(controller) {
+      enqueue = data => controller.enqueue(data)
+    },
+  });
+
+  (async () => {
+    for await (const chunk of chunkTTSInput(stream.getReader())) {
+      await ttsQueue.add(chunk.text)
+    }
+  })()
```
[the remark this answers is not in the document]
sumimakito: Nice ketchup. Thanks

packages/stage-ui/src/stores/chat.ts
```diff
@@ -10,6 +10,7 @@ import { ref, toRaw } from 'vue'
 import { useQueue } from '../composables'
 import { useLlmmarkerParser } from '../composables/llmmarkerParser'
 import { useLLM } from '../stores/llm'
+import { TTS_FLUSH_INSTRUCTION } from '../utils/tts'
```
nekomeowww, line 13: Much better!

packages/stage-ui/src/composables/queues.ts
```diff
@@ -103,23 +101,29 @@ export function useDelayMessageQueue() {
...
+  (async () => {
+    try {
+      for await (const chunk of chunkTTSInput(stream.getReader())) {
+        await ttsQueue.add(chunk.text)
+      }
+    }
+    catch (e) {
+      console.error('Error chunking input stream for TTS:', e)
+    }
+  })()
```
nekomeowww, line 121: Must we use IIFE here? (I mean... I don't know any existing solutions for this. Wondering if we could make it better without IIFE.)

sumimakito: The IIFE here would not be mandatory, and we can also construct something like:

```ts
async function pipeToQueue(stream: ReadableStream, queue: UseQueueReturn<string>): Promise<void>
```

… and consume it with `pipeToQueue(stream, ttsQueue)`.

As for the stream manipulation, we can also build some helper functions to make the exposure of the controller reusable in other places.

Actually, you mentioned a good point. Let me improve this in the follow-up PRs.

PR #332, opened 2025-07-29
fix(docs): call writeHead before write and remove next()

sumimakito:
Moving `writeHead` before `write` and removing the `next()` call to fix the following error while starting Vitepress:

```
node:_http_server:351
    throw new ERR_HTTP_HEADERS_SENT('write');
          ^

Error [ERR_HTTP_HEADERS_SENT]: Cannot write headers after they are sent to the client
    at ServerResponse.writeHead (node:_http_server:351:11)
    at file:///.../docs/node_modules/.vite-temp/config.ts.timestamp-1753799242233-dd15bbae2280f.mjs:193:13 {
  code: 'ERR_HTTP_HEADERS_SENT'
}
```


nekomeowww:
Thanks. Learned!

nekomeowww:
If so... I think many vite plugins was implementing this feature incorrectly, even the plugins inside vite itself...

PR #335, opened 2025-07-29
feat(stage-ui): introduce helpers for stream and tts chunking

sumimakito:
## Description

As a follow-up to our [discussion](https://github.com/moeru-ai/airi/pull/330/files/7d973d5f9d38cbaa1c2bdd8c7d2c8745d7dea7b3#r2240268721), introducing a helper function `chunkToTTSQueue` to provide the reader to the TTS chunking processor and then pipe the chunks to the TTS queue, and a helper function `createControllableStream` to create externally controllable `ReadableStream`s at ease.

### DIFF
--- a/packages/stage-ui/src/utils/stream.ts
+++ b/packages/stage-ui/src/utils/stream.ts
@@ -4,11 +4,13 @@ export interface ControllableStream<R = any> {
 }
 
 export function createControllableStream<R = any>(): ControllableStream<R> {
-  let controller: ReadableStreamDefaultController<R>
+  // WHY!: ReadableStream.start is called synchronously and immediately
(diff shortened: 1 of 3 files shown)


packages/stage-ui/src/utils/stream.ts
```diff
@@ -0,0 +1,16 @@
+export interface ControllableStream<R = any> {
+  stream: ReadableStream<R>
+  controller: ReadableStreamDefaultController<R>
+}
+
+export function createControllableStream<R = any>(): ControllableStream<R> {
+  // WHY!: ReadableStream.start is called synchronously and immediately
```
nekomeowww, line 7: I suppose this jammed you for quite a while haha.

nekomeowww (approved):

PR #352, opened 2025-08-06
fix(stage-ui): auto eye blink wasn't functioning

nekomeowww:


packages/stage-ui/src/components/Scenes/Live2D/Model.vue
```diff
@@ -179,15 +188,51 @@ async function loadModel() {
   }
 
   // This is hacky too
```
sumimakito, line 190: I started to agree with my past self 🥲

nekomeowww: No. I love this implementation, it's just we were missing pieces for customized models. My ReLU model doesn't have any motion groups integrated...

PR #356, opened 2025-08-07
chore: remove tool version

luoling8192:
For better DX

```
No version is set for command pnpm
Consider adding one of the following versions in your config file at /Users/luoling8192/Projects/airi/.tool-versions
nodejs 23.9.0
```


LemonNekoGH:
Don't remove it, you need to use `corepack enable` after `asdf install`

Issue #353, opened 2025-08-06
Crowdin integration


typed-sigterm:
> https://airi.moeru.ai/docs/en/blog/DevLog-2025.08.05/#:~:text=Crowdin

I'd be happy to help with this. Crowdin offers [free license](https://crowdin.com/page/open-source-project-setup-request) for open source projects, which requires the authorization from the project lead. Once it's ready, I can help set up the integration and coordinate the translation workflow.

- [x] Skip some CI & gemini review for Crowdin PRs
- [x] Bot account for Crowdin PR?

nekomeowww:
Authorized but configuration file required: https://support.crowdin.com/configuring-vcs-integrations-online/

nekomeowww:
<img width="1038" height="865" alt="Image" src="https://github.com/user-attachments/assets/2558dcc8-4f30-4ace-be58-ffbc83411986" />

Ok I finally achieved it...

nekomeowww:
Is this correct?

https://github.com/moeru-ai/airi/blob/main/crowdin.yml

typed-sigterm:
It's enough if using GitHub integration. If using Actions, `project_id` `api_token_env` is required.

nekomeowww:
<img width="1426" height="409" alt="Image" src="https://github.com/user-attachments/assets/9e375ee7-8f33-479f-bcd9-b3cae0084590" />

It seems, no strings detected for source files. Am I missing some configurations? I could invite you as manager.

nekomeowww:
Submitted: https://crowdin.com/page/open-source-project-setup-request.

typed-sigterm:
You can invite me via the email on my profile page.

nekomeowww:
<img width="1413" height="551" alt="Image" src="https://github.com/user-attachments/assets/ab2cde78-93bc-4b49-a51f-8aa9ee76fc46" />

It works!

typed-sigterm:
GitHub integration still has some problems, tracking above

nekomeowww:
![Image](https://github.com/user-attachments/assets/e052532a-6d02-45e5-94d6-90507ac2e8a0)

It works now I guess.

nekomeowww:
Haha, I think I break it again. Feel free to commit to fix it.

typed-sigterm:
Should be fixed now

nekomeowww:
![Uploading 474992042...]

nekomeowww:
[...]

PR #376, opened 2025-08-21
ci: skip crowdin

typed-sigterm:
link #353
close #354

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,7 @@ jobs:
 
   build-test:
     name: Build Test (${{ matrix.app_name }})
+    if: ${{ !(github.event_name == 'pull_request' && startsWith(github.event.pull_request.head.label, 'i18n/')) }}
     strategy:
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -48,6 +48,7 @@ jobs:
 
   build-test:
     name: Build Test (${{ matrix.app_name }})
+    if: ${{ !(github.event_name == 'pull_request' && startsWith(github.event.pull_request.head.label, 'i18n/')) }}
```
nekomeowww, line 51: Will this cause issues? For example we may not be able to be aware of bugs if i18n breaks vue-i18n or something.

typed-sigterm: I think... it won't happen at most time. At least we still test on `main`

nekomeowww: Sure. You are right. I missed the event filter for `pull_request` only. Then it is LGTM for me.

nekomeowww (approved):
Thanks. Long waited!

Issue #381, opened 2025-08-21
Creating new card on latest commit web gives error


0xGingi:
### Describe the bug

hello! When attempting to create a new card running from latest commit web, I'm seeing this in the console:
TypeError: crypto.randomUUID is not a function. (In 'crypto.randomUUID()', 'crypto.randomUUID' is undefined)

### System Info

```Shell
System:
    OS: macOS 26.0
    CPU: (10) arm64 Apple M4
    Memory: 210.75 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.6.0 - /opt/homebrew/bin/node
    npm: 11.5.1 - /opt/homebrew/bin/npm
    pnpm: 10.14.0 - /opt/homebrew/bin/pnpm
    bun: 1.2.20 - ~/.bun/bin/bun
  Browsers:
    Safari: 26.0
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [x] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

PR #382, opened 2025-08-21
Fix Card Creation

0xGingi:
## Description

Adds nanoid to replace crypto.randomuuid to fix card creation on web

## Linked Issues

https://github.com/moeru-ai/airi/issues/381


kwaa:
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

~~All modern browsers support crypto.randomUUID(), so introducing new dependencies to replace it is unnecessary~~.

Considering that the Web Crypto API is restricted to secure contexts, switching to `nanoid` makes sense.

PR #412, opened 2025-08-23
Adding voice interaction for both input and output text.

Ksirailway-base:
Added buttons and indication of TTS operation and back, speach-to-text.


Ksirailway-base:
Tomorrow I'll update some functions and clean up the code from logging.

nekomeowww (commented):
Why deleting so many files. Please check before submitting PR.

Issue #401, opened 2025-08-23
Add a checkbox "Use custom name without checking" and use third-party API platforms compatible with OpenAI or other platforms


Kevin589981:
### Clear and concise description of the problem

On some lesser-known third-party API platforms, their API interfaces are compatible with OpenAI (of course, the Base URL has been adjusted), but the model names used are not exactly the same. For example, the name of the models will be renamed, so when selecting a model, it prompts "No model matches' keyword", just add a checkbox "Use custom name without checking", which can solve this problem, and you won't have to adapt specifically anymore.
And I think it would be helpful to https://github.com/moeru-ai/airi/issues/388#issue-3344495663

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

perlicacute:
I think you can use every OpenAI Compatible with OpenAI as Provider, just put the Base URL from your provider

<img width="537" height="360" alt="Image" src="https://github.com/user-attachments/assets/73af8f39-0db4-4ceb-82fa-46cafef0014c" />

Its work great for me,

It will fetch model list from /models

Issue #388, opened 2025-08-22
ModelScope API support?


yrk111222:
### Clear and concise description of the problem

Excuse me, I am planning to submit a PR for ModelScope API support to airi Repo. However, it seems that things are not going smoothly. Do I need to submit a related PR to [xsai Repo](https://github.com/moeru-ai/xsai) before submitting a PR to this application repository? If it is convenient, could you tell me some steps and things to note to contribute?

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

nekomeowww:
Is this the documentation site, https://modelscope.cn/docs/model-service/API-Inference/intro?

It's fine to not to contribute to xsAI, it's not necessary.

I assume it's a OpenAI compatible API, `https://api-inference.modelscope.cn/v1/`, so I believe you can use `createOpenAI` and assign default baseUrl to `https://api-inference.modelscope.cn/v1/` then it should be finished. All you need to do is to create a new provider entry in https://github.com/moeru-ai/airi/blob/main/packages/stage-ui/src/stores/providers.ts

yrk111222:
> Is this the documentation site, https://modelscope.cn/docs/model-service/API-Inference/intro?
> 
> It's fine to not to contribute to xsAI, it's not necessary.
> 
> I assume it's a OpenAI compatible API, `https://api-inference.modelscope.cn/v1/`, so I believe you can use `createOpenAI` and assign default baseUrl to `https://api-inference.modelscope.cn/v1/` then it should be finished. All you need to do is to create a new provider entry in https://github.com/moeru-ai/airi/blob/main/packages/stage-ui/src/stores/providers.ts

<think>


</think>


Thank you for your detailed explanation and answers. I will submit a PR as soon as I finish testing the performance of some models. Appreciate your help again! 🙏🙏

PR #415, opened 2025-08-24
feat: Add generic OpenAI-compatible provider for chat and audio

skirkru:
### Description

- Directly address #401 #388

### Summary

This change adds a flexible, generic provider that can be configured with a custom baseUrl and apiKey. While this might seem redundant next to the existing OpenAI provider, its purpose is to make future requests like "Can you add Provider A, B, or C?" obsolete. Since many current and emerging AI services follow the OpenAI API specification, this single provider empowers users to connect to almost any of them without waiting for a dedicated integration. This approach removes user doubt about whether their preferred service is compatible and provides a unified, future-proof way to integrate with a growing ecosystem of tools. The provider is designed for chat, TTS, and STT tasks and has been successfully tested with external services like Groq's playtts-ai.

### Problem
Previously, while it was technically possible to use other OpenAI-compatible services (like Groq) by changing the base URL in the official OpenAI provider, this created significant ambiguity. The UI was labeled "OpenAI," which led to user confusion about whether other services were officially supported or would work correctly. This approach was not intuitive and lacked a clear, dedicated path for third-party integrations, forcing users to guess at the intended functionality.

### Solution
- Generic Provider: A single "OpenAI Compatible" provider was created to handle chat, TTS, and STT tasks, configurable via a baseUrl and apiKey.
- Separation of Concerns: This new provider is distinct from the official OpenAI provider, which clarifies the user experience and resolves the ambiguity.
- New UI Components: Dedicated settings pages and a testing playground for audio services were implemented to ensure easy configuration and verification.
- Performance: The speech settings page was optimized with watchDebounced to prevent redundant API calls when switching between providers.

### Comment
I just saw "wip: Transcription / Hearing module" when finishing this, so this probably conflicting with existing code

### DIFF
--- a/apps/stage-tamagotchi/src/pages/settings/modules/speech.vue
+++ b/apps/stage-tamagotchi/src/pages/settings/modules/speech.vue
@@ -17,6 +17,7 @@ import {
   FieldRange,
   Textarea,
 } from '@proj-airi/ui'
+import { watchDebounced } from '@vueuse/core'
 import { generateSpeech } from '@xsai/generate-speech'
 import { storeToRefs } from 'pinia'
 import { onMounted, onUnmounted, ref, watch } from 'vue'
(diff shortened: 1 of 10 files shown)


apps/stage-tamagotchi/src/pages/settings/modules/speech.vue
```diff
@@ -58,9 +59,20 @@ onMounted(async () => {
   await speechStore.loadVoicesForProvider(activeSpeechProvider.value)
 })
 
-watch(activeSpeechProvider, async () => {
+// stop spam 8 times
```
nekomeowww:
```suggestion
```

nekomeowww (approved):

PR #451, opened 2025-08-25
fix(stage-ui,stage-web): emit literals asap and show streaming messages

sumimakito:
* Move back to the FSM-style opening/closing tags scanner in `useLlmmarkerParser` to emit literals ASAP, otherwise they are only emitted when a complete tag is parsed
  * Time complexity should be linear
  * Pass all tests in `llmmarkerParser.test.ts`
  * Fix: Emit special tokens with the surrounding `<|` and `|>` to keep the behavior untouched
  * Fix: Do not emit a literal if it's part of an incomplete special tag
* Show the streaming message in the web chat history

### DIFF
--- a/packages/stage-ui/src/composables/llmmarkerParser.ts
+++ b/packages/stage-ui/src/composables/llmmarkerParser.ts
@@ -1,3 +1,6 @@
+const TAG_OPEN = '<|'
+const TAG_CLOSE = '|>'
+
 /**
  * A streaming parser for LLM responses that contain special markers (e.g., for tool calls).
  * This composable is designed to be efficient and robust, using a regular expression
@@ -15,9 +18,15 @@
 export function useLlmmarkerParser(options: {
   onLiteral?: (literal: string) => void | Promise<void>
   onSpecial?: (special: string) => void | Promise<void>
+  /**
+   * The minimum length of text required to emit a literal part.
+   * Useful for avoiding emitting literal parts too fast.
+   */
+  minLiteralEmitLength?: number
 }) {
+  const minLiteralEmitLength = Math.max(1, options.minLiteralEmitLength ?? 1)
   let buffer = ''
-  const tagRegex = /(<\|.*?\|>)/
+  let inTag = false
 
   return {
     /**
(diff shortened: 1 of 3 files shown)


perlicacute:
LGTM

packages/stage-ui/src/stores/chat.ts
```diff
@@ -65,26 +64,57 @@ export const useChatStore = defineStore('chat', () => {
     onAssistantResponseEndHooks.value.push(cb)
   }
 
+  // I know this nu uh, better than loading all language on rehypeShiki
+  const codeBlockSystemPrompt = '- For any programming code block, always specify the programming language that supported on @shikijs/rehype on the rendered markdown, eg. ```python ... ```\n'
+  const mathSyntaxSystemPrompt = '- For any math equation, use LaTeX format, eg: $ x^3 $, always escape dollar sign outside math equation\n'
```
nekomeowww: Interesting that we do prompt injecting.

packages/stage-ui/src/composables/llmmarkerParser.ts
```diff
@@ -1,85 +1,86 @@
...
+        if (!inTag) {
+          const openTagIndex = buffer.indexOf(TAG_OPEN)
+          if (openTagIndex < 0) {
+            if (buffer.length - 1 >= minLiteralEmitLength) {
+              const emit = buffer.slice(0, -1)
+              buffer = buffer[buffer.length - 1]
+              await options.onLiteral?.(emit)
+            }
+            break
+          }
```
nekomeowww, line 51: Oh nice way, thanks.

apps/stage-tamagotchi/src/components/InteractiveArea.vue
```diff
@@ -46,6 +49,34 @@ async function handleSend() {
...
+            data: base64Data,
+            mimeType: file.type,
+            url: URL.createObjectURL(file),
+          })
+        }
+      }
+      reader.readAsDataURL(file)
+    }
+  }
+}
```
nekomeowww: Image support?

nekomeowww:
We can merge if ready.

packages/stage-ui/src/stores/chat.ts
perlicacute: @nekomeowww Yes, its can be omitted, its just enforcing the llm to use specific format to response with language and latex format, without that, it can be intermittence

nekomeowww:
I merged it.

PR #465, opened 2025-08-26
fix(stage-tamagotchi): fix first time initialization blank and an overlook

skirkru:
#### Summary

This commit fixes an issue where the application window was blank on its first launch and corrects the data structure for the `tauri://move` event to handle window repositioning correctly.

#### Problem

1.  **Blank Window on First Load:** On initial startup, the main window was rendered with zero opacity, making it appear blank until the first mouse interaction.
2.  **Incorrect Event Payload:** The `tauri://move` event listener was attempting to access window coordinates from a nested `payload` object (`event.payload.payload.x`) that did not exist, causing errors when moving the window.

#### Solution

1.  A new `isFirstTime` reactive flag was added. The UI's opacity is now prevented from being set to zero on the initial render, ensuring content is visible immediately. The transparency logic is now only applied after the first mouse event.
2.  The type definition for the `tauri://move` event in [`apps/stage-tamagotchi/src/composables/tauri.ts`](apps/stage-tamagotchi/src/composables/tauri.ts:9) was corrected. The event listener in [`apps/stage-tamagotchi/src/pages/index.vue`](apps/stage-tamagotchi/src/pages/index.vue:198) was updated to access the coordinates directly from `event.payload`, aligning it with the actual data structure.

#### My comment

I remembered this is on yesterday commit, i guess i somehow undoing it

### DIFF
--- a/apps/stage-tamagotchi/src/composables/tauri.ts
+++ b/apps/stage-tamagotchi/src/composables/tauri.ts
@@ -36,7 +36,7 @@ export interface WindowFrame {
 
 interface Events {
   'tauri://resize': unknown
-  'tauri://move': { payload: { x: number, y: number } }
+  'tauri://move': { x: number, y: number }
   'tauri://close-requested': unknown
   'tauri://destroyed': unknown
   'tauri://focus': unknown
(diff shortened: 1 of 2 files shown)


perlicacute:
ok messed up again to not using main branch lol

PR #471, opened 2025-08-26
Fix: prevent module peer overwrite on re-announce

Iro96:
*Fixed an issue* where `p.index` was always undefined because `event.data.index` was never assigned, causing multiple modules with the same name but different indexes to overwrite each other in `peersByModule.`

**Changes:**
- Correctly assign event.data.index to p.index.
- Unregister the existing module peer before re-announcing.
- Ensure peersByModule remains consistent on disconnect.

**Description:**

This update prevents module peers from being overwritten when they share the same name but use different indexes. It improves peer lifecycle management by properly assigning indexes, unregistering before re-announcing, and ensuring cleanup on disconnect.

### DIFF
--- a/packages/server-runtime/src/index.ts
+++ b/packages/server-runtime/src/index.ts
@@ -74,8 +74,9 @@ function main() {
         let event: WebSocketEvent;
         try {
           event = message.json() as WebSocketEvent;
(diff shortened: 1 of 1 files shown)


Iro96:
fix all

Issue #297, opened 2025-07-18
bug: use https://v2.tauri.app/plugin/http-client/ for fetch on Tauri


nekomeowww:
### Describe the bug

Currently we are using fetch / ofetch directly, we need to replace it to https://v2.tauri.app/plugin/http-client/ for LLM's call at least, or users will have to deal with CORS erros.

### System Info

```Shell
None.
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

nekomeowww:
We moved to Electron so this is no longer required.

PR #474, opened 2025-08-26
feat(stage-ui): Refactor providers and add credential validation

skirkru:
#### Summary
This pull request introduces a major refactoring of the provider configuration in `stage-ui`. It abstracts the creation of OpenAI-compatible providers into a reusable builder function, significantly reducing code duplication. It also adds a user-facing feature for real-time credential validation, which has been abstracted into a `useProviderValidation` composable for maintainability. Additionally, it includes a fix for a CORS issue with the Anthropic provider.

#### Problem
1.  The setup for multiple OpenAI-compatible providers (like Mistral, Moonshot, etc.) involved a lot of boilerplate and repeated logic, making the `providers.ts` file difficult to maintain and prone to errors.
2.  Users had no immediate feedback on whether the credentials they entered for a provider were correct, leading to a poor user experience.
3.  Direct browser requests to the Anthropic API were failing due to missing CORS headers. This addresses issue #297.
4.  The initial implementation of credential validation introduced significant code duplication across more than 20 Vue components.

#### Solution
1.  **Provider Refactoring:** A `buildOpenAICompatibleProvider` function was created to abstract the common logic for defining these providers. This makes adding new providers a one-line change and centralizes the logic for listing models and validating configurations.
2.  **Anthropic CORS Fix:** A custom fetch wrapper is now used for the Anthropic provider to inject the required `anthropic-dangerous-direct-browser-access: true` header, resolving the CORS issue.
3.  **Credential Validation UI:** A validation mechanism was added to the provider settings pages. It uses a debounced function to check the credentials against the provider's API and displays a clear success or error message to the user.
4.  **Validation Logic Abstraction:** The duplicated UI logic for validation was refactored into a `useProviderValidation` Vue composable. This hook encapsulates all the state management, API calls, and debouncing, dramatically simplifying the provider setting components.
5.  **Internationalization:** All new user-facing strings for the validation feature have been added to the i18n locale files.

#### My Comment
Just a small PR, but the code is much better than before. While the provider system is still a bit monolithic, this refactoring is a good step forward. It also fixes the annoying Anthropic CORS issue.


packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
```diff
@@ -0,0 +1,134 @@
...
+      let responseModelList = null
+
+      if (validationChecks.includes('health')) {
+        try {
+          const responseChat = await fetch(`${config.baseUrl as string}chat/completions`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders }, method: 'POST' })
+          responseModelList = await fetch(`${config.baseUrl as string}models`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders } })
+          // Little Hack because every provider with OpenAI compatible support this
+          if (!([200, 400, 401].includes(responseChat.status) || [200, 400, 401].includes(responseModelList.status))) {
+            errors.push(new Error(`Invalid Base URL, ${config.baseUrl} is not supported`))
+          }
```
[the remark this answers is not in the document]
perlicacute: 404 will be false positive, so its either 200, 400, 401

apps/stage-tamagotchi/src/pages/settings/providers/openai-compatible-audio-transcription.vue
```diff
@@ -102,33 +67,38 @@ function handleResetSettings() {
...
+          </div>
+        </template>
+      </Alert>
+      <Alert v-if="isValid && isValidating === 0" type="success">
+        <template #title>
+          {{ t('settings.dialogs.onboarding.validationSuccess') }}
+        </template>
+      </Alert>
+    </ProviderSettingsContainer>
   </ProviderSettingsLayout>
```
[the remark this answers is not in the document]
perlicacute: cant get my mic to work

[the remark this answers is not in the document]
perlicacute: ok fine now

packages/stage-ui/src/components/Scenarios/Dialogs/Onboarding/Onboarding.vue
```diff
@@ -409,6 +409,11 @@ onMounted(() => {
                 </div>
               </template>
             </Alert>
+            <Alert v-if="isValid && isValidating === 0" type="success">
```
typed-sigterm, line 413: Successful validation doesn't need to be highlighted since it's only a pre-check

perlicacute: @typed-sigterm right, but I also added actual API key validation, also to give a feedback for new user, especially non tech people

Since we already notify the user about an invalid key, I think it's consistent to also confirm when the key is valid. the additional alert does this, but another option could be to simply enable the 'Next' button

In my opinion, it would be counter-intuitive for a user to click 'Next', only to find out later that the key was wrong and they have to go back. The success alert prevents that confusion.

typed-sigterm: After #416, failling validation blocks the onboarding process. The spin icon indicates validation is in progress, `Next` button is now only enabled if the validation succeeds. If the hint still seems to weak, perhaps this is better:
image

perlicacute: Yeah that is better approach for the UX, do we already have the library for that?

But In my opinion, double Alert is good choice for now, since if we want to change the UI to show the error the message on the input, we should standarize it on everyplace that have input to show error there

Its probably best to have it on separate PR as this PR specifically trying to refactoring provider store from DRY and add functional validation, not just check if the string is there

typed-sigterm, line 413: Needs a little adjustment

Image

perlicacute: can you elaborate more?

packages/stage-ui/src/stores/providers.ts
```diff
@@ -184,112 +210,62 @@ export const useProvidersStore = defineStore('providers', () => {
...
-      if (!response.ok) {
-        throw new Error(`Failed to fetch OpenRouter models: ${response.statusText}`)
+  async function isTamagotchi() {
+    if ('window' in globalThis && globalThis.window != null) {
+      if (('__TAURI_INTERNALS__' in globalThis.window && globalThis.window.__TAURI_INTERNALS__ != null) || location.host === 'tauri.localhost') {
+        return true
       }
+    }
+    return false
+  }
```
nekomeowww, line 220: Haha, this is hard, isn't it, thanks for doing this.

packages/stage-ui/src/stores/providers.ts
```diff
@@ -184,112 +210,62 @@ export const useProvidersStore = defineStore('providers', () => {
...
-      }))
+  async function isBrowserAndMemoryEnough() {
+    const isInApp = await isTamagotchi()
+
+    if (isInApp)
+      return false
+
+    const webGPUAvailable = await isWebGPUSupported()
+    if (webGPUAvailable) {
+      return true
```
nekomeowww, line 230: Thank you for fixing this.

packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
```diff
@@ -0,0 +1,142 @@
...
+        return { errors, reason: errors.map(e => e.message).join(', '), valid: false }
+      }
+
+      const validationChecks = validation || []
+      let responseModelList = null
+
+      if (validationChecks.includes('health')) {
+        try {
+          const responseChat = await fetch(`${config.baseUrl as string}chat/completions`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders }, method: 'POST' })
+          responseModelList = await fetch(`${config.baseUrl as string}models`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders } })
```
nekomeowww: Should we use `@xsai/models` and `@xsai/generate-text` for this?

packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
```diff
@@ -0,0 +1,142 @@
...
+          // Little Hack because every provider with OpenAI compatible support this
+          if (!([200, 400, 401].includes(responseChat.status) || [200, 400, 401].includes(responseModelList.status))) {
+            errors.push(new Error(`Invalid Base URL, ${config.baseUrl} is not supported`))
```
nekomeowww, line 90: Already handled by xsai as well I guess?

nekomeowww:
Need to resolve conflict i think.

packages/stage-ui/src/stores/providers.ts
perlicacute: nope, actually simpler than you thought, I just dont know what you want to do with that ar first

perlicacute:
@nekomeowww I havent add the local version, I will do the conflicts first, have bad eyesight from yesterday haha, i keep the CI a problem so you know its still not finished

packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
perlicacute: @nekomeowww its check its behavior, there is no need to use there are built in functions that support it

nekomeowww: But xsai handles endpoint, apikey, and everything well enough, xsai is part of Moeru AI, which is also part of AIRI, feel free to use it, small and compact.

perlicacute: I just reviewed how the implementation there

https://github.com/moeru-ai/xsai/blob/main/packages/model/src/utils/list-models.ts
https://github.com/moeru-ai/xsai/blob/main/packages/shared/src/utils/response-catch.ts
and .https://github.com/moeru-ai/xsai/blob/main/packages/generate-text/src/index.ts

Well we can use that, but the implementation will be quite longer, but sure

I just preventing using chainsaw when saw work fine

perlicacute:
@nekomeowww ok this should be it

packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
perlicacute: I can't get the same thing with those dependency

nekomeowww: if we choose to manage the impl of `baseURL` and endpoint concat from `chat/completions` / `models`, there might be a chance where will not be able to  maintain the same behavior as `generateText(...)`.

those dependencies were already installed, they are not something huge and big as Vercel SDK.

image

image

perlicacute: you missed the point why its there, its not checking whats the output, its only check whats is the respond code

Why you might ask? even if some provider use OpenAI Compatible format, the API key validation is different, they just share the same endpoint, the content who knows

The point is, that dependency is to generate text, while what validation need is checking what the response code.

Its simple, consistent, as we dont care about the content

perlicacute: unless the dependency have a function to validate API key, we can move the current code to that

nekomeowww (approved):

Issue #456, opened 2025-08-26
TTS lags behind conversation


lorisj-elefant:
### Describe the bug

Occasionally the TTS will lag behind the conversation, and will stop. Essentially the bug is something like this:

- user : "hello"
- assistant(text) : "hello! how are you"
- tts : "hello" 
- user "what?"
-assistant(text) :  "I just told you!"
- tts : "how are you"

at this point it stalls, it doesnt send the tts request with "I just told you" to provider, and only does so after the user message. It seems like the tts queue is out of sync with the conversation. This doesnt happen all of the time, but it seems like there is some issue that causes this to happen, and the only way to fix is to reload.

Unfortunately I don't have a solid way to reproduce yet. 

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (12) x64 12th Gen Intel(R) Core(TM) i5-1235U
    Memory: 675.45 MB / 7.69 GB
  Binaries:
    Node: 24.6.0 - ~\scoop\apps\nodejs\current\node.EXE
    npm: 11.5.1 - ~\scoop\apps\nodejs\current\npm.CMD
    pnpm: 10.14.0 - ~\scoop\apps\nodejs\current\pnpm.CMD
  Browsers:
    Edge: Chromium (139.0.3405.111)
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

lorisj-elefant:
@nekomeowww

nekomeowww:
Fixed already.

PR #476, opened 2025-08-26
fix: TTS lags behind conversation

Iro96:
## Description
Um.. I did some search over all related files again and found currently, there is no logic in `Stage.vue` to stop or clear the `audioQueue` or `interrupt audio playback` when a new message is sent. This means if a user sends a new message while `TTS/audio` is still playing, the previous audio will continue, causing the lag and out-of-sync like issues https://github.com/moeru-ai/airi/issues/456 described.

## Linked Issues

Issue 456: 
- https://github.com/moeru-ai/airi/issues/456

## Additional Context

- When a new message is composed `onBeforeMessageComposed`, stop any currently playing audio and clear the audioQueue to ensure only the new message is spoken.


Iro96 (commented):
There are some error on self-installer? idk bru

`Error: Something went wrong, self-installer exits with code 1`

image

packages/stage-ui/src/composables/queues.ts
```diff
@@ -1,12 +1,11 @@
...
+import type { UseQueueReturn, useQueue } from './queue'
 
 import { sleep } from '@moeru/std'
 
-import { EMOTION_VALUES } from '../constants/emotions'
 import { createControllableStream } from '../utils/stream'
-import { chunkToTTSQueue } from '../utils/tts'
-import { useQueue } from './queue'
+import { EMOTION_VALUES } from '../constants/emotions'
+import { TTS_FLUSH_INSTRUCTION, chunkToTTSQueue } from '../utils/tts'
```
sumimakito: Let's adjust the imports a bit:

```suggestion
import type { Emotion } from '../constants/emotions'
import type { UseQueueReturn } from './queue'

import { sleep } from '@moeru/std'
import { llmInferenceEndToken } from '@proj-airi/stage-ui/constants'

import { EMOTION_VALUES } from '../constants/emotions'
import { createControllableStream } from '../utils/stream'
import { chunkToTTSQueue, TTS_FLUSH_INSTRUCTION } from '../utils/tts'
import { useQueue } from './queue'
```

packages/stage-ui/src/composables/queues.ts
```diff
@@ -104,13 +103,18 @@ export function useDelayMessageQueue() {
...
   const { stream, controller } = createControllableStream<Uint8Array>()
-
   chunkToTTSQueue(stream.getReader(), ttsQueue)
 
   return useQueue<string>({
     handlers: [
       async (ctx) => {
-        controller.enqueue(encoder.encode(ctx.data))
+        // If the message is the llmInferenceEndToken, enqueue a flush instruction to TTS queue
+        if (ctx.data === llmInferenceEndToken) {
```
sumimakito: I searched through the code base, and it seems we don't enqueue `llmInferenceEndToken` (i.e., `<|llm_inference_end|>`) to the `messageContentQueue`, except for `apps/stage-web/src/pages/test/queues/messages.vue`, which is for debugging purposes.

In this case, I feel `llmInferenceEndToken` should never appear in the queue handler here. Furthermore, before calling the `onStreamEnd` hooks, we do enqueue two `TTS_FLUSH_INSTRUCTION` tokens to instruct the TTS chunker and the grapheme cluster reader to flush:

https://github.com/moeru-ai/airi/blob/b169cd999d3e1e058f5cf7bd262a1adfc5319719/packages/stage-ui/src/stores/chat.ts#L214-L221

We may need to inspect the original issue further, instead of enqueuing another `TTS_FLUSH_INSTRUCTION` here. It could be due to the TTS queue being interrupted by a newly enqueued message, but I cannot reproduce either.

cc @nekomeowww for double-checking the facts with `llmInferenceEndToken`

Iro96:
@sumimakito @nekomeowww

packages/stage-ui/src/components/Scenes/Stage.vue
```diff
@@ -74,11 +74,21 @@ vrmStore.onShouldUpdateView(async () => {
 const audioAnalyser = ref<AnalyserNode>()
 const nowSpeaking = ref(false)
 const lipSyncStarted = ref(false)
+let currentAudioSource: AudioBufferSourceNode | null = null
 
 const audioQueue = useQueue<{ audioBuffer: AudioBuffer, text: string }>({
   handlers: [
     (ctx) => {
       return new Promise((resolve) => {
+        // Stop any currently playing audio
```
sumimakito, line 83: Thank you for the rework. 

As internally in `useQueue`, we `await` for each handler called with a picked payload, and later in `source.onended` here we resolve this Promise, the handler here should be executed synchronously:

```
prev_source::start -> prev_source::end -> (dequeue) -> curr_source::start -> curr_source::end -> …
```

With this prerequisite, the audio playback itself is not likely to overlap.

Maybe we can try this: When interrupted by a new message, we should stop enqueuing items to the TTS and the audio queue, so that both the speech generation request and the audio playback won't "interweave" potentially.

Iro96:
i did something and this branch is broken... so let's start on a new pull request xDD. sorry for this everyone

Iro96:
lets move works to new pull request @typed-sigterm @nekomeowww @sumimakito 
- https://github.com/moeru-ai/airi/pull/494

PR #477, opened 2025-08-26
fix(stage-ui): add rehype-sanitize to markdown processors

sumimakito:
Add [rehype-sanitize](https://github.com/rehypejs/rehype-sanitize) to Markdown processors as we are rendering the output in `v-html`, although processors like `@shikijs/rehype` may provide sanitisation to some extent.


sumimakito:
Closing as we may need to preserve useful outputs from other processors. Maybe revisit later

PR #490, opened 2025-08-26
feat(nix): auto-update pnpmDeps hash

Weathercold:
## Description

I didn't test the action
The assets hash still needs to be manually updated

### DIFF
--- a/.github/workflows/update-nix-pnpm-deps-hash.yaml
+++ b/.github/workflows/update-nix-pnpm-deps-hash.yaml
@@ -1,0 +1,30 @@
+name: Update Nix pnpmDeps Hash
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'pnpm-lock.yaml'
+
+permissions:
+  contents: write
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: cachix/install-nix-action@v31
+        with:
+          extra_nix_config: experimental-features = nix-command flakes
+
+      - name: Update Hash
+        run: nix/update-pnpm-deps-hash.sh
+
+      - uses: stefanzweifel/git-auto-commit-action@v6
+        with:
+          commit_message: 'chore(nix): update pnpmDeps hash'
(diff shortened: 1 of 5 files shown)


.github/workflows/update-nix-pnpm-deps-hash.yaml
```diff
@@ -0,0 +1,30 @@
+name: Update Nix pnpmDeps Hash
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'pnpm-lock.yaml'
```
nekomeowww, line 9: Should we update this in cron hours? Otherwise each commit about pnpm-lock.yaml will produce one extra automated commit.

@sumimakito do you have any experience on similar thing to this? Perhaps cascading updating something based on changes, would love to hear from best practice.

nekomeowww (commented):
Rest is LGTM.

nekomeowww:
Or perhaps, we can achieve this through a postinstall hook? In such case, we could write script to do post release to update the versions for Nix as well when using `bumpp` to release new versions. 

I wrote on for updating `Cargo.toml` when releasing new things here https://github.com/moeru-ai/airi/blob/252ca8ef7b1e08cd75b77bd6c10cd077b5437153/bump.config.ts#L15-L44

Weathercold:
> Or perhaps, we can achieve this through a postinstall hook? In such case, we could write script to do post release to update the versions for Nix as well when using `bumpp` to release new versions.
> 
> I wrote on for updating `Cargo.toml` when releasing new things here
> 
> https://github.com/moeru-ai/airi/blob/252ca8ef7b1e08cd75b77bd6c10cd077b5437153/bump.config.ts#L15-L44

Unfortunately, since the nix flake uses the latest git commit, it'll be outdated as soon as a commit modifies pnpm-lock.yaml

What about a precommit hook along with an updated workflow that doesn't run if the hash is changed? Or throttling the workflow?

nekomeowww:
Ok I understand it now.

nekomeowww (approved):

nekomeowww:
It works https://github.com/moeru-ai/airi/commit/a81711137bedd54000b1a44bd7ec9b973d078b36

Issue #508, opened 2025-08-28
feat(stage:tamagotchi, AppImage) Bumping up a Tauri-Build version for stage-tamagotchi/src-tauri


gg582:
### Clear and concise description of the problem

As a developer, I found that AppImage build is broken when using Tauri-
At first, I thought that it is a GStreamer bundling error.
However, in most of the cases, GStreamer Plugins can be easily installed via distro's package manager.
And as a result, it was not related to building itself. GStreamer is related to Speech functionality(however I don't have paid Speech backend service, so I cannot test)

Surprisingly, build process is totally fine along with these plugin versions:
```
tauri-build = { version = "2.4.0", features = [] }
tauri-plugin-log = "2.6.0"
tauri-plugin-os = "2.3.1"
```

So, I want to propose a version up for an application.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #509, opened 2025-08-28
chore(deps): bump

gg582:
## Description

Current Tauri-plugin's version is outdated. So it fails to build an AppImage.
Also, it has log package that is release candidate.
Also, using Tauri-Build 2.4.0 to fix a bug is a simple solution inside a Cargo.toml(rust part)

## Linked Issues

#508


nekomeowww (commented):
`package.json`?

Oh and you don't need to bump everything by yourself, try using

https://github.com/moeru-ai/airi/blob/6142d16d27b16b15a6fcd465d9c75cfac1c603ba/package.json#L36

gg582:
okay, I will try using pnpm up. I think it is okay to change only tauri-build version

gg582:
Actually, it was about `rehype-katex` and `rehype-math`. They are missing at a main branch

gg582:
And that missed packages are the parts of `./package/stage-ui`. That is why only pnpm-lock.yaml changed

gg582:
Removed those two packages; although CI check fails

gg582:
Running pnpm up now.

gg582:
Also, we **must** bump up `tauri-build` version in `apps/stage-tamagotchi/src-tauri/Cargo.toml` if we would support linux. I hope that this project will support linux. Many projects are not handling linux client, but there are linux desktop users, although they are quite few. And I am an end user who uses linux too.

nekomeowww:
Feel free to bump, i mean to use pnpm up to bump npm modules.

nekomeowww (approved):

gg582:
Yes, I used pnpm bump from now on, but the core problem was on rust part. So necessary patches are solely Cargo.*. But it is better ti bump. Thnx

Issue #507, opened 2025-08-28
feature request: sending queue


lyratu:
### Describe the bug

<img width="600" height="559" alt="Image" src="https://github.com/user-attachments/assets/2589bdde-152f-42c6-aae6-9e2c784c4ee4" />

When I use Chinese input, pressing Enter types the text from the input method into the input box. Then, pressing Enter sends the message. If I type content again and press Enter, the text content that was typed into the input box will be sent to the chat.

Application version: AIRI_0.7.2-beta.2_windows_amd64-setup
This application was downloaded and installed from: https://airi.moeru.ai/docs/zh-Hans/

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i5-1240P
    Memory: 2.02 GB / 15.73 GB
  Binaries:
    Node: 14.19.0 - C:\nvm4w\nodejs\node.EXE
    Yarn: 1.22.22 - C:\nvm4w\nodejs\yarn.CMD
    npm: 6.14.16 - C:\nvm4w\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (139.0.3405.119)
    Internet Explorer: 11.0.26100.1882
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [x] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

Iro96:
This happens when you send new messages while the llm model is still processing previous messages. 
I suggest preventing users from sending new messages while the model is processing, additionally we can adding a typing effect to each message created for instead of sending the message directly. @nekomeowww

typed-sigterm:
Can you reproduce that in the web version?

perlicacute:
Havent tried zh input, I will try to take a look how its behavior when using it

lyratu:
> Can you reproduce that in the web version?

<div style="display: flex;">
<div>
 <p>The web version</p>
[...]

PR #512, opened 2025-08-28
fix(chat): Handle IME composition in chat input

skirkru:
#### Description

Fix: #507

### Summary:
This pull request addresses an issue with the chat input field where using an Input Method Editor (IME) would cause messages to be sent prematurely when the user confirms a composition.

### Problem:
When typing in languages that require an IME (e.g., Chinese, Japanese, Korean), pressing the Enter key to select a candidate from the composition list would also trigger the message submission event. This resulted in sending incomplete or unintended messages, providing a poor user experience for users of these languages.

### Solution:
- Introduced a `isComposing` reactive state to track whether an IME composition is currently active.
- Utilized the `@compositionstart` and `@compositionend` events on the textarea to toggle the `isComposing` state accordingly.
- Modified the `handleSend` function to prevent message submission if `isComposing` is `true`, ensuring the message is only sent when the user explicitly intends to.
- Changed the event listener from a generic `@submit` to a more specific `@keydown.enter.exact.prevent="handleSend"` to ensure only a direct press of the Enter key (without IME composition) triggers the send action.
- Added logic to clear the input field after a message is successfully sent, improving the chat flow.

### My comment:
I just tried IME Input

### DIFF
--- a/apps/stage-tamagotchi/src/components/InteractiveArea.vue
+++ b/apps/stage-tamagotchi/src/components/InteractiveArea.vue
@@ -24,8 +24,14 @@ const { messages } = storeToRefs(useChatStore())
 const { t } = useI18n()
 const providersStore = useProvidersStore()
 const { activeModel, activeProvider } = storeToRefs(useConsciousnessStore())
+const isComposing = ref(false)
 
 async function handleSend() {
+  // IME problemma 你好airi
+  if (isComposing.value) {
+    return
+  }
+
   if (!messageInput.value.trim() && !attachments.value.length) {
     return
   }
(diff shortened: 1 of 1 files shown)


apps/stage-tamagotchi/src/components/InteractiveArea.vue
```diff
@@ -24,8 +24,14 @@ const { messages } = storeToRefs(useChatStore())
 const { t } = useI18n()
 const providersStore = useProvidersStore()
 const { activeModel, activeProvider } = storeToRefs(useConsciousnessStore())
+const isComposing = ref(false)
 
 async function handleSend() {
+  // IME problemma 你好airi
```
nekomeowww:
```suggestion
```

Issue #520, opened 2025-08-28
proposal: better structure, stateless Client, and Plugin system


nekomeowww:
We are facing difficulties to support various of inputs, processors, middlewares, and message bridge of different need, and customization, we need better structure for abstraction with scopes and division of files in this project. And also, the verification capabilities.

## Introduction

Currently, we mixed all code of entire app within three directories:

- `apps/stage-web`: for Web specific impl
- `apps/stage-tamagotchi`: for Desktop specific impl
- `packages/stage-ui`: for shared components, stores, utils, types of both Web and Desktop.

and for Desktop (which is the so called tamagotchi) version, we have `apps/stage-tamagotchi/src-tauri` written in Rust with Tauri framework for cross platform support.

Renderers of models here:

- [`Stage.vue`](https://github.com/moeru-ai/airi/blob/main/packages/stage-ui/src/components/Scenes/Stage.vue): the core is implemented here for stage of models
- [VRM](https://github.com/moeru-ai/airi/tree/main/packages/stage-ui/src/components/Scenes/VRM)
- [Live2D](https://github.com/moeru-ai/airi/tree/main/packages/stage-ui/src/components/Scenes/Live2D)

But essentially, thanks for @sumimakito @luoling8192 and @kwaa, by referencing other impl of [AI VTuber, and AI companion apps](https://github.com/uezo/ChatdollKit), this kind of structure may survive in long term...

## Proposal

> All thoughts behind the scene: https://www.tldraw.com/p/CbJDTPT0Jh17cbwas0yUJ?d=v-8525.-3980.24463.12690.brxzFGJvDgFYbKF3sP2Fw

<img width="1440" height="720" alt="Image" src="https://github.com/user-attachments/assets/ca51d45d-4d6e-4a50-b343-f23489d0e9f2" />

I would love to propose a better structure similar to this with the following abstractions by introducing Primitives and data flows:

> [!WARNING]  
> We should not rely on any Browser specific capabilities, except `Scene`, and `UI`. Or otherwise it's not possible to run AIRI in headless mode in long running servers.

> [!WARNING]  
> When implementing the building blocks of the following Primitives, especially `Input`, `Output`, `ModelProvider`... except `Scene`, and `UI`, we should try to avoid to use any Vue specific library, or rely on Pinia to manage global states.

- Client
  - Scene
    - Renderer
      - Live2D
      - VRM
      - MMD
    - SceneControl
      - for Pixi
      - for three.js
    - BodyControl
[...]

PR #523, opened 2025-08-28
feat: ollama setting pages now has validating animation

Misaka-0x447f:
## Description

- ollama config page now has validator loading animation
image

### DIFF
--- a/packages/stage-ui/src/components/Misc/Alert.vue
+++ b/packages/stage-ui/src/components/Misc/Alert.vue
@@ -1,9 +1,11 @@
 <script setup lang="ts">
-import { computed } from 'vue'
+import { computed, useSlots } from 'vue'
 
 const props = defineProps<{
   type?: 'error' | 'warning' | 'success' | 'info' | 'loading'
 }>()
+
+const slots = useSlots()
 
 const containerClass = computed(() => {
   switch (props.type) {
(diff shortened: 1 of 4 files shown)


packages/stage-ui/src/components/Misc/Alert.vue
```diff
@@ -31,20 +43,26 @@ const titleClass = computed(() => {
...
-  <div class="flex flex-col gap-3 rounded-xl px-2 pb-3 pt-2" :class="containerClass">
+  <div class="flex flex-col gap-3 rounded-xl p-2" :class="containerClass">
     <div class="flex items-center gap-1.5 font-medium">
       <div class="text-2xl" :class="iconClass" />
       <div :class="titleClass">
         <slot name="title" />
       </div>
     </div>
-    <div class="px-1 text-sm">
+    <div v-if="$slots.content" class="px-1 text-sm">
```
nekomeowww: Can we avoid to use `$slots` / `$attrs` and `$props`? Try with useSlots or (i forgot the name)

```suggestion
    <div v-if="$slots.content" class="px-1 text-sm">
```

nekomeowww:
Btw check #520, not related, but you would love to know.

packages/stage-ui/src/components/Misc/Alert.vue
Misaka-0x447f: Wait what why `$slot` still works in vue3

Misaka-0x447f: fixed

PR #524, opened 2025-08-29
feat(linux): Add deb, rpm packages as installation formats

gg582:
## Description

Although this project has nix build for Linux, it is thought that Debian GNU/Linux based and Red Hat Enterprise Linux based distributions are prominent among Linux desktop users.

It is considered that Ubuntu, Debian, Linux Mint, Fedora, are commonly favored by contemporary desktop users, and most of the people prefer distro's own package formats over than AppImage.

Appimage owes a problem, that it is difficult to update an application while keeping their own local data.
Since this project's agenda is to make an AI VTuber, some people might feel intimacy to his own-prompted model. This problem would be bigger when AIRI supports Long-Term Memory.

It is suggested that we may support some user-friendly formats to further enhancement.

I am not used to packaging, so if somebody urges to add supports for Flatpak repository, I would like to engage.

## Additional Context

Would you focus on the package dependencies to automatically install required libraries? Some users who use Linux is not familiar with Software development.
So I'd like to mention that Linux Operating System is not only for developers, but also a personal choice which can be chosen as convenience. 

Thank you.


nekomeowww (approved):

Weathercold:
Wait don't merge yet
(I wonder if I can push to the pr directly?)

nekomeowww:
> (I wonder if I can push to the pr directly?)

Why?

Weathercold:
> > (I wonder if I can push to the pr directly?)
> 
> Why?

The binary names in nix/package.nix need to be changed

Weathercold (approved):

nekomeowww:
> > > (I wonder if I can push to the pr directly?)
> 
> > 
> 
> > Why?
> 
> 
> 
> The binary names in nix/package.nix need to be changed

Perhaps open another for that.

PR #543, opened 2025-09-02
fix(audio): wrap buffer in Float32Array to allow safe indexing

Iro96:
- Wrapped input buffer in a Float32Array so samples can be indexed safely. 
- Keeps `numSamples = buffer.byteLength` to preserve existing output behavior, so the generated WAV files remain consistent with the previous implementation.

## Description

This PR fixes a runtime error in WAV conversion where `ArrayBufferLike` could not be indexed. By wrapping it in a `Float32Array`, samples can now be accessed safely without changing the output format or size.

## Additional Context

The logic for numSamples remains unchanged (buffer.byteLength) to maintain backwards compatibility, even though samples.length would be more accurate for strict WAV sizing. 
### Here is that version

```ts
import { encodeBase64 } from '@moeru/std/base64'

function writeString(dataView: DataView, offset: number, str: string) {
  for (let i = 0; i < str.length; i++) {
    dataView.setUint8(offset + i, str.charCodeAt(i))
  }
}

export function toWav(samples: Float32Array, sampleRate: number, numChannels = 1) {
  const numSamples = samples.length
  const bytesPerSample = 2 // 16-bit PCM
  const blockAlign = numChannels * bytesPerSample
  const byteRate = sampleRate * blockAlign
  const dataSize = numSamples * bytesPerSample
  const buffer = new ArrayBuffer(44 + dataSize)
  const view = new DataView(buffer)

  // RIFF header
  writeString(view, 0, 'RIFF')
  view.setUint32(4, 36 + dataSize, true)
  writeString(view, 8, 'WAVE')

  // fmt chunk
  writeString(view, 12, 'fmt ')
  view.setUint32(16, 16, true) // PCM chunk size
  view.setUint16(20, 1, true)  // Audio format = PCM
  view.setUint16(22, numChannels, true)
  view.setUint32(24, sampleRate, true)
  view.setUint32(28, byteRate, true)
  view.setUint16(32, blockAlign, true)
  view.setUint16(34, 16, true) // bits per sample

  // data chunk
  writeString(view, 36, 'data')
  view.setUint32(40, dataSize, true)

  // PCM samples
  let offset = 44
  for (let i = 0; i < numSamples; i++) {
    const s = Math.max(-1, Math.min(1, samples[i]))
    const val = s < 0 ? s * 0x8000 : s * 0x7FFF
    view.setInt16(offset, val, true)
    offset += 2
  }

  return buffer
}

export function toWavBase64(samples: Float32Array, sampleRate: number) {
  return encodeBase64(toWav(samples, sampleRate))
}
```

### DIFF
--- a/packages/audio/src/encoding/wav.ts
+++ b/packages/audio/src/encoding/wav.ts
@@ -7,10 +7,10 @@ function writeString(dataView: DataView, offset: number, string: string) {
 }
 
 export function toWav(buffer: ArrayBufferLike, sampleRate: number, channel = 1) {
+  const samples = new Float32Array(buffer) // ✅ allows indexing
   const numChannels = channel
-  const numSamples = buffer.byteLength
+  const numSamples = buffer.byteLength // ✅ unchanged (same as your original)
 
-  // Create the WAV file container
   const arrayBuffer = new ArrayBuffer(44 + numSamples * 2)
   const dataView = new DataView(arrayBuffer)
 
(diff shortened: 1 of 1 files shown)


sumimakito:
Could we preserve the original code comments, as they might be helpful for readers in the future?

Iro96:
> Could we preserve the original code comments, as they might be helpful for readers in the future?

got it

packages/audio/src/encoding/wav.ts
```diff
@@ -7,8 +7,9 @@ function writeString(dataView: DataView, offset: number, string: string) {
 }
 
 export function toWav(buffer: ArrayBufferLike, sampleRate: number, channel = 1) {
+  const samples = new Float32Array(buffer) // allows indexing
```
nekomeowww, line 10: Will this cause the copy of array?

sumimakito: No. Passing an existing `ArrayBufferLike` into `Float32Array`'s constructor creates another view of the underlying data. Since ricky0123/vad outputs audio in `Float32Array` in `onSpeechEnd`, I think this will be a better match.

nekomeowww (approved):

PR #549, opened 2025-09-02
fix(stage-web): support on pc submit with enter

ricardomiss:
## Description
This PR is to Enable Enter key submission for desktop users in MobileInteractiveArea component.
Desktop users in `MobileInteractiveArea.vue` component couldn't submit with Enter key. Mobile users continue to get new lines with Enter (no change)
## Linked Issues
none

## Additional Context
none

### DIFF
--- a/apps/stage-web/src/components/Layouts/MobileInteractiveArea.vue
+++ b/apps/stage-web/src/components/Layouts/MobileInteractiveArea.vue
@@ -40,6 +40,16 @@ const { send, onAfterMessageComposed, discoverToolsCompatibility } = useChatStor
 const { messages } = storeToRefs(useChatStore())
 const { t } = useI18n()
 
+function isMobileDevice() {
+  return /Mobi|Android|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)
+}
+
+async function handleSubmit() {
+  if (!isMobileDevice()) {
+    await handleSend()
+  }
+}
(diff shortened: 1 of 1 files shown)


apps/stage-web/src/components/Layouts/MobileInteractiveArea.vue
```diff
@@ -40,6 +40,16 @@ const { send, onAfterMessageComposed, discoverToolsCompatibility } = useChatStor
 const { messages } = storeToRefs(useChatStore())
 const { t } = useI18n()
 
+function isMobileDevice() {
+  return /Mobi|Android|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)
+}
```
nekomeowww, line 45:
```suggestion
function isMobileDevice() {
  return "virtualKeyboard" in navigator
}
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent
https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API

Let's use feature detection instead of UA parse, shall we?

perlicacute: Limited avaibility, not available on firefox and apple safari

nekomeowww: Sure.

Issue #387, opened 2025-08-22
Memory System Architecture Implementation Alternatives


lucas-oma:
### Clear and concise description of the problem

> Please correct me if I'm wrong in any point

Looks like AIRI currently lacks a memory system for RAG. The existing DuckDB setup seems to only only creates a temporary table with no persistence or advanced memory functionality for web/app.

### Option 1: Standalone Memory-Core + Database Implementations

A shared TypeScript/JavaScript library (@proj-airi/memory-core) that provides database-agnostic memory operations, with platform-specific implementations for DuckDB (web) and PostgreSQL (desktop).

#### Pros:
- **Platform Flexibility**: Each platform uses its optimal database (DuckDB for web, Postgres for rust/tauri app)
- **No Infrastructure**: No servers needed for basic usage

#### Cons:
- **Data Isolation**: Memories don't sync between platforms
- **Setup Complexity**: Each platform needs database setup (code: classes, methods, etc)
- **Limited Sharing**: Can't share memories across devices

### Technical Requirements

***Memory-Core Package (@proj-airi/memory-core):***

- Database-agnostic interfaces for memory operations
- Shared consolidation and categorization logic
- Common types and utilities
- Platform-agnostic embedding service interface

**DuckDB WASM Persistence (for web UI):**

- Configure DuckDB for IndexedDB storage
- Implement database export/import for backup
- Handle browser storage limits

**PostgreSQL Integration (for rust/tauri base app):**

> Similar to what has been done for telegram-bot

- Use the pgvector extension
- Implement HNSW vector indexing
- Handle local PostgreSQL setup

### Option 2: Docker Memory Service

A centralized memory service running in Docker that provides a shared memory backend for all AIRI applications (web, desktop, Discord bot, Telegram bot) through a REST API.

#### Pros:
- **Data Synchronization**: Shared memories across all platforms (Web, desktop, Discord, Telegram, all share same memory if configured)
- **Centralized Management**: Single source of truth/memory
- **Advanced Features**: Full PostgreSQL + pgvector capabilities even for web app
- **Cloud Hosting**: Can be hosted on personal server ( or Moeru's cloud in the future!) for shared memory across devices. Could also be useful for sharing across computers, and maybe smartphones in the future.
- **Consistency:** All platforms use identical memory logic and features
- **Future-Proof:** Easy to extend with new features for all platforms simultaneously

#### Cons:
- **Infrastructure Required**: Needs server setup (user will have to run docker alongside the usual app)
[...]

PR #564, opened 2025-09-08
Memory system feature

lucas-oma:
## Description

This  PR introduces a **Memory System** for AIRI, enabling it to access and utilize long-term context to generate more relevant and consistent responses. This is still in early stage, hence lot of optimization opportunity available, and that is one of the reasons for the PR (so admins could provide feedback and optimization ideas)

> Note: it has heavily based/inspired by the telegram-bot memory

The core functionality is a new context-building pipeline that runs before each LLM call. As detailed in the sequence diagram:

```mermaid
sequenceDiagram
  actor Alice as User
  participant P1 as Backend Service: <br>API Server
  participant P2 as LLM Providers<br>(multiple)
  participant P3 as Backend Service: <br>Context Building
  participant P4 as Postgres DB

  Alice ->> P1: 1. new msg
  P1 ->> P2: 2. generate_embedding(msg)
  P2 ->> P1: 3. return msg_embedding
  P1 ->> P3: 4. msg_embedding
  par 🧠 Context building
    P3 ->> P4: 5a. get last 10 messgaes + responses
    P3 ->> P4: 5b. get top 5 most relevant
    P3 ->> P4: 5c. get top 10 most relevant memory_fragments
    P3 ->> P4: 5d. get top 10 most memory_associated
    P3 ->> P4: 5e. get top 10 most memory_consolidated
    P3 ->> P4: 5f. get all memory_entities
    P3 ->> P4: 5g. get all memory_long_term_goals
    P3 ->> P4: 5h. get all memory_short_term_ideas
  end
  P4 ->> P3: 6. return context data
  P3 ->> P3: 7. assemble context
  P3 ->> P1: 8a.return assembled context
  P1 ->> Alice: 8b. return assembled context
  Alice ->> P2: 9. send msg + context
  P2 ->> Alice: 10. return LLM response
```

- When a new message arrives, it is first converted into a numerical embedding.
- This embedding is used to query the PostgreSQL database to retrieve various types of context, including recent messages, relevant memory fragments, consolidated memories, and key entities and goals.
- This assembled context is then passed to the LLM (message + context) to inform its response, allowing it to "remember" past conversations and user preferences. **Here we can optimize by doing this call server side, hence not returning back to user for it to make another call)**

This is the main functional change in this PR.

### Additional Changes

This PR also includes a few nice updates:

- The chat history loads (some) past messages on init.
-  A "Load More History" button when scrolled to the top to load more (past) messages.

## Linked Issues

https://github.com/moeru-ai/airi/issues/387

## Additional Context

- This is **not** the final version, but a working architecture which should be easy to upgrade (context building logic is concentrated in a single file).
- Some memory tables are **not yet** used for context building (or not even populated)
- I have some more ideas for upgrades and optimizations, these are mentioned in the [memory-service-overview.md](https://github.com/lucas-oma/airi/blob/memory-system-feature/services/memory-service/docs/memory-service-overview.md) file
- [Memory service Readme file](https://github.com/lucas-oma/airi/blob/memory-system-feature/services/memory-service/README.md)


lucas-oma:
Feedback is welcome and expected! I'm particularly interested in any thoughts you might have on the current (very basic) context-building system (Airi's "brain"). I have some ideas for how to make it more efficient, and I'm looking forward to collaborating on them.

nekomeowww:
Rebase needed.

lucas-oma:
@nekomeowww Rebased and updated

Tho I see some workflows here in github are failing, should I look into that?

perlicacute:
@lucas-oma from diagram, steps 1-8, I can see you change how it interact with llm, which add "context builder" before its actual send message to the llm

So this mean user will be forced to use embedding and postgresql? 
For newbies these two probably will be a major roadblock

"What is an API key" is already a problem for newbies

and what about web based one? like https://airi.moeru.ai/
surely it doesnt have those access

Im not saying the implementation was bad, but Im trying to become a non tech weebs who trying to install AIRI later

lucas-oma:
Hi @skirkru , you are right that this type of setup might be a bit difficult for non-techs, although for creating a decent memory system I think embeddings and a vector database are needed.

> So this mean user will be forced to use embedding and postgresql?

My implementation is managed in the settings page, and must be activated via a checkbox from there, if not activated then the behavior is as usual.

Still you raise a valid point, maybe DuckDB-Wasm can be used instead (or as an alternative) of postgres, then having this extra postgres step will not be necessary. Not sure about the performance when talking about vectors in DuckDB tho, but I can give it a try (probably non-techs will prefer this kind of approach)

Jhon4561488228:
> Hi @skirkru , you are right that this type of setup might be a bit difficult for non-techs, although for creating a decent memory system I think embeddings and a vector database are needed.
> 
> > So this mean user will be forced to use embedding and postgresql?
> 
> My implementation is managed in the settings page, and must be activated via a checkbox from there, if not activated then the behavior is as usual.
> 
> Still you raise a valid point, maybe DuckDB-Wasm can be used instead (or as an alternative) of postgres, then having this extra postgres step will not be necessary. Not sure about the performance when talking about vectors in DuckDB tho, but I can give it a try (probably non-techs will prefer this kind of approach)

You are right, ChromaDB is needed for good long-term memory storage.

Together with another guy, we are currently trying to implement and locally embed Short-term and Long-term memory with various AGI features (determining the user's emotions, adjusting to their emotions, and improved semantic search and storage of memory by emotional parameters and others, up to adapting the AI ​​voice to the user's mood)

But we use Rust as a base for storing all memory locally, we tried to raise local servers through Py scripts, but they crashed from the received data volume, besides, there will be an obvious limitation on data storage, so I think Rust local storage is the best option.
Short-term memory - we use SQL (it will be slower than RAM, but the context will be saved much longer)
Long-term memory - CromaDB (full integration of mem0 and partially memvid for video storage of mp4 with the h265 codec (ultra compression, ffmpeg is required))

perlicacute:
Hi @lucas-oma, yes, embedding and vector database are essential for memory, but my main concern are that for non tech people, who want to test it on browser first 

While for embedding it can use the current provider list, but not all provider like open router provide embedding

I think the stack should be discussed with @nekomeowww, while the current model list are using indexeddb to store its files.

Also I think that we also should able to use both database format, and can be choosen in the setting, so there are options for non tech friendly and for someone who need performant database

gg582:
How about using PGLite?

gg582:
Wait, but if we use PGlite we need to rewrite all queries as it does not support drizzle ORM

gg582:
Would you mind checking PR #589 ?
This uses embedded postgres.
Thanks,

nekomeowww:
> Wait, but if we use PGlite we need to rewrite all queries as it does not support drizzle ORM

They support. https://orm.drizzle.team/docs/connect-pglite
Even if they do not support, you can use `new PgDialect()` and `new QueryBuilder()` from `drizzle-orm/pg-core` for this.

gg582:
Cool. Then is it better to rewrite #597 to use PGLite? Currently it is using embedded postgres and the last step is packaging(considering pkg, or tauri sidecar)

nekomeowww:
> Cool. Then is it better to rewrite #597 to use PGLite? Currently it is using embedded postgres and the last step is packaging(considering pkg, or tauri sidecar)

Yes.

nekomeowww:
Rebase needed.
Generally LGTM. I think this PR have no issues once merged with #597. I can improve it later.

gg582:
Thank you! And now we need a last step to implement PGLite support :)

gg582:
My idea is to support both variants, Embedded Postgres and PGLite(since they have different purpose)

gg582:
I am currently implementing PGLite support on #597

lucas-oma:
Hey, sorry I've been kinda busy lately, will return to this and related subject in the next days. 

Was the PGLite feature merged into main or is it still in development? @nekomeowww @gg582 

Also any opinions on using duckdb instead? (not sure about its vector capabilities but might be a way to integrate a local  memory system for people that prefer a plug-and-play approach)

gg582:
Yes, duckdb will be fine. I have my pglite merged branch on my forked repo. Can I make a PR to your fork? :) or you can use #597.

gg582:
I reopened #597 in advance.
You can check this and partially/fully merge.
Also, Memory separation per model cards hasn't implemented.
Can we use model's name as a identifier?
We can separate chat history and model by simply adding a delimeter.

gg582:
Also, I've got a request to move this into /packages...can we do it?

gg582:
but if we rewrite it in duckdb, we should rewrite backup functionality, etc...I've made an export function to manage history in #597, and I think it will be fine after adding memory separation per model cards...And it is a small request...can you write a db structure when it's changed? I'd like to reform my memory backup to changed database...

gg582:
So..you are still working on it, right? I hope that I can get how much have you done.

nekomeowww:
Not updated for too long. Closing for now. Comment to ask me to re-open once you have time to apply suggestions.

Issue #567, opened 2025-09-09
[BUG] Alibaba provider in speed module cannot work


yanghui1-arch:
### Describe the bug

I follow the guide and run `pnpm dev` and set the speech(tts) provider api key. Then I enter the website `http://localhost:5173/settings/modules/speech` to select `alibaba`, `Cozyvoice` model and a voice. Finally I click the `test` button on the right plane. Nothing happens. No any error information on the web. Then I open my browser dev tool error information as follows:
```Shell
Error generating speech: XSAIError: Remote sent 400 response: 
{
  "errors": [
    {
      "id": "BAD_REQUEST",
      "links": null,
      "status": 400,
      "code": "BAD_REQUEST",
      "title": "Bad Request",
      "detail": "failed to run task, task-failed event received, error_code: ModelNotFound, error_message: Model not found (cozyvoice-v2)!",
      "source": null,
      "meta": null
    }
  ]
}
```

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 AMD Ryzen 7 8845H w/ Radeon 780M Graphics
    Memory: 1.34 GB / 13.76 GB
  Binaries:
    Node: 24.7.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.5.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.15.1 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (140.0.3485.54)
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

nekomeowww:
Should be fixed.

PR #568, opened 2025-09-09
fix(stage-ui): reolve speech module cannot test voice using alibaba cosyvoice-v1

yanghui1-arch:
## Description

This pr fix the wrong alibaba speech model name in stage-ui/src/stores to ensure test voice function in speech module works well.
 - fix alibaba tts model name from `cozyvoice` to `cosyvoice`
## Linked Issues

#567


yanghui1-arch:
The Netlify build seems to fail during `postinstall`. I didn’t touch dependencies in this PR.

Issue #570, opened 2025-09-10
SSML text is not supported at the moment while using cosyvoice in Speech module


yanghui1-arch:
### Describe the bug

I select cosyvoice-v1 and then click test voice. The error information is shown in following image. 

<img width="1990" height="870" alt="Image" src="https://github.com/user-attachments/assets/9c61138f-0f43-48b3-b941-38857e4d17c0" />

According to alibaba official document, only part voice names are supported SSML. I hope the error information can be shown on the page. No error information will make AIRI user confused expecially for beginner.

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 AMD Ryzen 7 8845H w/ Radeon 780M Graphics
    Memory: 1.04 GB / 13.76 GB
  Binaries:
    Node: 24.7.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.5.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.15.1 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (140.0.3485.54)
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

yanghui1-arch:
The error raises because the code in apps/stage-web/src/pages/settings/modules/speech.vue
```typescript
const input = useSSML.value
      ? ssmlText.value
      : speechStore.supportsSSML ? speechStore.generateSSML(testText.value, activeSpeechVoice.value, { ...providerConfig, pitch: pitch.value }) : testText.value
```
`useSSML.value` is false  now and speechStore.suuportsSSML is implemented as follows in packages/stage-ui/src/stores/speech.ts.
```typescript
const supportsSSML = computed(() => {
    // Currently only ElevenLabs and some other providers support SSML
[...]

PR #579, opened 2025-09-12
fix(speech): unable to test voice with openai and cosyvoice-v2 in speech module

yanghui1-arch:
## Description

This PR fixes speech module is unable to test voice with openai and alibaba.
- add `compatibleModels` for `VoiceInfo`
- add more friendly information when no supported voices of selected speech model.
- add more friendly information when no preview voice samples.
- fix OpenAI speech models are not shown.
- fix unable to test voice

Add `compatibleModels` field because voices are not only decided by provider but also player's selected model. For example, 龙婉 is only supported by cosyvoice-v1. Previous renderer logic is using provider to select all voices and then render. It's fail to test voice when player select a voice which is not supported by his/her selected model. <br/>

Add `listModels` for `providerMetadata['openai-audio-speech']` reason is openai-audio-speech has `capabilities` field which makes it no list models during building OpenAI compatible provider. It is the main reason results into no models to be selected when player choose OpenAI provider in speech module. <br/>

Add an alert in `VoiceCardManySelect` component when the model voices are not supported now.

## Linked Issues

#567
#570

### DIFF
--- a/apps/stage-web/src/pages/settings/modules/speech.vue
+++ b/apps/stage-web/src/pages/settings/modules/speech.vue
@@ -320,7 +320,7 @@ function updateCustomModelName(value: string) {
               v-model:search-query="voiceSearchQuery"
               v-model:voice-id="activeSpeechVoiceId"
               :voices="availableVoices[activeSpeechProvider]?.filter(voice => {
-                return voice.compatible_models?.includes(activeSpeechModel)
+                return voice.compatibleModels?.includes(activeSpeechModel)
               }).map((voice) => {
                 return {
                   id: voice.id,
--- a/apps/stage-web/src/pages/settings/modules/speech.vue
+++ b/apps/stage-web/src/pages/settings/modules/speech.vue
@@ -321,15 +321,13 @@ function updateCustomModelName(value: string) {
               v-model:voice-id="activeSpeechVoiceId"
               :voices="availableVoices[activeSpeechProvider]?.filter(voice => {
                 return voice.compatibleModels?.includes(activeSpeechModel)
-              }).map((voice) => {
-                return {
-                  id: voice.id,
-                  name: voice.name,
-                  description: voice.description,
-                  previewURL: voice.previewURL,
-                  customizable: false,
-                }
-              })"
+              }).map(voice => ({
+                id: voice.id,
+                name: voice.name,
+                description: voice.description,
+                previewURL: voice.previewURL,
+                customizable: false,
+              }))"
               :searchable="true"
               :search-placeholder="t('settings.pages.modules.speech.sections.section.provider-voice-selection.search_voices_placeholder')"
               :search-no-results-title="t('settings.pages.modules.speech.sections.section.provider-voice-selection.no_voices')"
--- a/apps/stage-web/src/pages/settings/modules/speech.vue
+++ b/apps/stage-web/src/pages/settings/modules/speech.vue
@@ -319,18 +319,24 @@ function updateCustomModelName(value: string) {
             <VoiceCardManySelect
               v-model:search-query="voiceSearchQuery"
               v-model:voice-id="activeSpeechVoiceId"
-              :voices="availableVoices[activeSpeechProvider]?.map(voice => ({
-                id: voice.id,
-                name: voice.name,
-                description: voice.description,
-                previewURL: voice.previewURL,
-                customizable: false,
-              }))"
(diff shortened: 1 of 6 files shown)


packages/i18n/src/locales/zh-Hans/settings.yaml
```diff
@@ -214,6 +214,8 @@ pages:
             search_models_results: 找到 {count} / {total} 个模型
             search_voices_placeholder: 搜索声线...
             search_voices_results: 找到 {count} / {total} 个声线
+            unsupported_voice_warning_title: 没有支持的声线
+            unsupported_voice_warning_content: 我们正在尽快支持该模型的所有音色，如果你迫切希望支持该模型音色，请在github上告诉我们
```
nekomeowww:
```suggestion
            unsupported_voice_warning_content: 我们正在尽快支持该模型的所有音色，如果你迫切希望支持该模型音色，请在 GitHub 上联系我们 https://github.com/moeru-ai/airi/issues
```

packages/i18n/src/locales/en/settings.yaml
```diff
@@ -239,6 +239,8 @@ pages:
             search_models_results: Found {count} of {total} models
             search_voices_placeholder: Search voices...
             search_voices_results: Found {count} of {total} voices
+            unsupported_voice_warning_title: No supported voices
+            unsupported_voice_warning_content: Try a different model or provider. We are working on supporting all the voice for this model as quickly as possible. If you need it urgently, please let us know on GitHub.
```
nekomeowww:
```suggestion
            unsupported_voice_warning_content: Try a different model or provider. We are working on supporting all the voice for this model as quickly as possible. If you need it urgently, please let us know on GitHub at https://github.com/moeru-ai/airi/issues.
```

apps/stage-web/src/pages/settings/modules/speech.vue
```diff
@@ -319,18 +319,24 @@ function updateCustomModelName(value: string) {
...
+              :voices="availableVoices[activeSpeechProvider]?.filter(voice => {
+                return voice.compatibleModels?.includes(activeSpeechModel)
+              }).map((voice) => {
+                return {
+                  id: voice.id,
+                  name: voice.name,
+                  description: voice.description,
+                  previewURL: voice.previewURL,
+                  customizable: false,
+                }
```
nekomeowww: Why mapping again?

yanghui1-arch: > Why mapping again?

I notice previous return object includes these fields. So I map again to avoid potential bugs.

apps/stage-web/src/pages/settings/modules/speech.vue
```diff
@@ -319,18 +319,24 @@ function updateCustomModelName(value: string) {
...
               v-model:voice-id="activeSpeechVoiceId"
-              :voices="availableVoices[activeSpeechProvider]?.map(voice => ({
-                id: voice.id,
-                name: voice.name,
-                description: voice.description,
-                previewURL: voice.previewURL,
-                customizable: false,
-              }))"
+              :voices="availableVoices[activeSpeechProvider]?.filter(voice => {
+                return voice.compatibleModels?.includes(activeSpeechModel)
```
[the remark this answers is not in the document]
yanghui1-arch: > high
> 
> The current filter logic `voice.compatibleModels?.includes(activeSpeechModel)` will evaluate to `undefined` (a falsy value) if `voice.compatibleModels` is not defined. This will cause voices that are compatible with all models (and thus have no `compatibleModels` property, e.g., from ElevenLabs) to be incorrectly filtered out. To fix this, you should also check if `compatibleModels` is undefined, which signifies compatibility with all models.
> 
> ```
>                 return !voice.compatibleModels || voice.compatibleModels.includes(activeSpeechModel)
> ```

I have solved this potential bug now.

packages/i18n/src/locales/zh-Hans/settings.yaml
yanghui1-arch: > 我们正在尽快支持该模型的所有音色，如果你迫切希望支持该模型音色，请在 GitHub 上联系我们 https://github.com/moeru-ai/airi/issues

Thanks! I have fix it. Please recheck it.

packages/i18n/src/locales/en/settings.yaml
yanghui1-arch: > 

Thanks! I have fix it. Please recheck it.

Issue #566, opened 2025-09-08
I configure OpenAI Compatible with a self-deployed service which doesn't need API key, but the input is required


LoTwT:
## description

I configure OpenAI Compatible with a self-deployed service, which is authorised by other methods instead of API key. But the API key is required in the form.

https://github.com/user-attachments/assets/5175aba6-6f9c-4c75-b841-25e2b22aa8fa

### System Info

```Shell
System:
    OS: macOS 15.7
    CPU: (10) arm64 Apple M1 Pro
    Memory: 484.45 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.6.0 - ~/.local/state/fnm_multishells/62711_1757346978783/bin/node
    npm: 11.5.2 - ~/.local/state/fnm_multishells/62711_1757346978783/bin/npm
    pnpm: 10.15.1 - ~/.local/state/fnm_multishells/62711_1757346978783/bin/pnpm
    bun: 1.2.22 - ~/.bun/bin/bun
  Browsers:
    Chrome: 140.0.7339.81
    Safari: 18.6
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

MLTQ:
seconded, would like to see this fixed

nekomeowww:
Should be fixed already for Web.

LoTwT:
> Should be fixed already for Web.

I checked the git history and found that you have fixed the issue. However, it appears that your [fix code](https://github.com/moeru-ai/airi/blob/38f748bcc4f8794d8d311650e0db3d6214df3549/apps/stage-web/src/pages/settings/providers/openai-compatible.vue#L65) is missing.

nekomeowww:
Let me check this.

nekomeowww:
Should be fixed now.

PR #582, opened 2025-09-12
fix(providers): Improve OpenRouter validation and support key-less OpenAI-compatible services

skirkru:
#### Summary

This pull request introduces two key improvements for provider configuration: it adds a robust, custom validation mechanism for OpenRouter to prevent false positives with invalid API keys, and it makes the API key field optional for all OpenAI-compatible providers, enabling support for local models and other services that do not require an API key.

### DIFF
--- a/packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
+++ b/packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
@@ -54,9 +54,6 @@ export function buildOpenAICompatibleProvider(
     validateProviderConfig: async (config: Record<string, unknown>) => {
       const errors: Error[] = []
 
-      if (!config.apiKey) {
-        errors.push(new Error('API key is required'))
-      }
       if (!config.baseUrl) {
         errors.push(new Error('Base URL is required'))
       }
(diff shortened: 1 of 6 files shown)


perlicacute:
Also fix #566

packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
```diff
@@ -83,7 +80,7 @@ export function buildOpenAICompatibleProvider(
 
       if (validationChecks.includes('health')) {
         try {
-          responseChat = await fetch(`${config.baseUrl as string}chat/completions`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders }, method: 'POST' })
+          responseChat = await fetch(`${config.baseUrl as string}chat/completions`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders }, method: 'POST', body: '{"model": "test"}' })
```
nekomeowww, line 83: `messages` is required field, is it? AFAIK, for many LLM gateway implementation, this results in invalid argument error with 400 Bad Request. Should we choose to use `@xsai/generate-text` for this now?

perlicacute: @nekomeowww I just made quick fix for this error for openrouter checker, Im still in travel, but you are right,

We can go to use `@xsai/generate-text`  to check the validity of the API key through generating sample test like in lobe chat using known model

perlicacute: But that method incurring cost

nekomeowww: Single `Hi` call should not waste too much.

PR #583, opened 2025-09-12
lip-syncing for live2d models

303-SeeOther:
## Description
The function of lip-syncing for live2d models has been initially implemented, and a character card editing feature has been added.

## Additional Context

Currently, the default lip shape voice synchronization configuration only includes vowel lip shape recognition. In the future, we will add support for users to upload their own lip shape voice synchronization configuration files.


packages/stage-ui/src/components/Scenes/Stage.vue
```diff
@@ -73,197 +58,6 @@ vrmStore.onShouldUpdateView(async () => {
   }, 100)
 })
 
-const audioAnalyser = ref<AnalyserNode>()
-const nowSpeaking = ref(false)
-const lipSyncStarted = ref(false)
-let currentAudioSource: AudioBufferSourceNode | null = null
-
-const audioQueue = createQueue<{ audioBuffer: AudioBuffer, text: string }>({
```
nekomeowww: If those were all deleted, how can the character connect and interact with TTS service?

303-SeeOther: Thank you very much for your attention. I have migrated the code related to interaction with the TTS server to packages/stage-ui/src/components/Scenes/Live2D/Model.vue. After testing, it can interact with the TTS server normally.

nekomeowww:
Now natively supported.

PR #486, opened 2025-08-26
chore: add Russian translation

Neko-233:
This PR introduces a complete Russian translation (README.ru-RU.md) of the main README file. This will make the project more accessible to the Russian-speaking community.

This PR includes:

1. A new README.ru-RU.md file in the /docs directory.
2. An update to the main README.md to include a link to the new Russian translation.
3. A link to a community-written guide for new developers.

PR #589, opened 2025-09-14
Memory System Feature: Add Embedded Postgres implementation for PR #564

gg582:
## Description

Using Embedded Postgres for non-tech users can be a good solution.
Goal: By toggling boolean from Settings UI, user can turn on/off it.

## Linked Issues

#564

### DIFF
--- a/apps/stage-tamagotchi-electron/src/renderer/pages/index.vue
+++ b/apps/stage-tamagotchi-electron/src/renderer/pages/index.vue
@@ -55,6 +55,7 @@ const modeIndicatorClass = computed(() => {
     flex="~ col"
     relative z-2 h-full overflow-hidden rounded-xl
     transition="opacity duration-500 ease-in-out"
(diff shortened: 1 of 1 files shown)


gg582:
I merged PR #486 temporarily because I could not enter the chatroom, but you can revert it if you want to merge those two separately

gg582:
Reverted #486

gg582:
I guess that currently working memory is not Embedded Postgres based long term memory. I prepared some bare-bone codes.
If I saw it right, is it relying on different memories?
But all I could see from codes were just postgres.
And I haven't installed postgreSQL in my computer locally.

But it is kinda strange that the memory worked, I guess it is IndexedDB

Also, if a user turns on long-term memory, short-term memories(?) from current IndexedDB should be pushed into database too.

gg582:
I added PR on different branch named `experimental/rag-custom-build`

Issue #557, opened 2025-09-06
SST  OpenAI Compatible Provider


DYSDYW:
### Describe the bug

I have successfully set up and run a local proxy server for speech recognition, compatible with the OpenAI API. The server runs on http://127.0.0.1:5000 and uses the Flask and faster-whisper libraries (with the small model). I can send POST requests via curl and receive correct transcriptions.

The issue is with Airi configuration:

OpenAI Compatible:
I enter Base URL: http://127.0.0.1:5000 in the OpenAI Compatible provider settings.
Result: The provider does not activate, settings are not saved, and the "Hearing" module does not work.

<img width="1905" height="537" alt="Image" src="https://github.com/user-attachments/assets/e0bb7769-9148-4dd5-a007-2273f599ec67" />

OpenAI.com:
I enter the same Base URL: http://127.0.0.1:5000 in the OpenAI.com provider settings.

<img width="1896" height="708" alt="Image" src="https://github.com/user-attachments/assets/3967032b-a81b-4ab4-bed3-4f07f91661ef" />

Result: The provider visually does not activate (the "No Providers Configured" message remains), 

<img width="1277" height="205" alt="Image" src="https://github.com/user-attachments/assets/7b2f9e81-537e-437f-b9cb-c0eb6a12006b" />

HOWEVER:
The whisper-1 model appears in the "Hearing" module.
I can turn on voice recording.
Airi successfully sends audio to my server and receives a transcription (visible in server logs and the sandbox).
Despite this, the "Hearing" module remains inactive (no green "Speech" indicator), and core functionality is broken.

<img width="1909" height="869" alt="Image" src="https://github.com/user-attachments/assets/fe744102-cd43-4701-bfed-52d651941090" />

Questions:

Is this a bug in the OpenAI Compatible provider setup interface?
Why does OpenAI.com partially work, but the module still fails to activate?
How can I correctly configure my local server to get the "Hearing" module working?
Thank you for your work!

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 AMD Ryzen 5 7500F 6-Core Processor
    Memory: 18.10 GB / 31.61 GB
  Binaries:
    Node: 22.18.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.5.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.15.1 - ~\AppData\Roaming\npm\pnpm.CMD
  Browsers:
[...]

PR #606, opened 2025-09-26
fix: enhance OpenAI provider validation for speech recognition servers (issue #557)

Iro96:
## Description

The OpenAI Compatible provider was failing to activate when connected to speech recognition servers (like faster-whisper) that only support transcription endpoints `(/audio/transcriptions)`. The validation logic was too restrictive, requiring chat completion endpoints `(/chat/completions)` to pass health checks, which caused failures for transcription-only servers.

### Solution
Modified the validation logic in `openai-compatible-builder.ts` to accept any valid OpenAI-compatible endpoint (chat, models, or transcription) as sufficient for provider activation.

---

## Linked Issues

This PR try to resolves [issue 557](https://github.com/moeru-ai/airi/issues/557) where OpenAI Compatible provider was not activating with speech recognition servers.

See New Changes


packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
```diff
@@ -85,8 +85,19 @@ export function buildOpenAICompatibleProvider(
           responseChat = await fetch(`${config.baseUrl as string}chat/completions`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders }, method: 'POST', body: '{"model": "test"}' })
           responseModelList = await fetch(`${config.baseUrl as string}models`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders } })
 
-          if (!([200, 400, 401].includes(responseChat.status) || [200, 400, 401].includes(responseModelList.status))) {
-            errors.push(new Error(`Invalid Base URL, ${config.baseUrl} is not supported`))
+          // Also try transcription endpoints for speech recognition servers
+          let responseTranscription = null
+          try {
+            responseTranscription = await fetch(`${config.baseUrl as string}audio/transcriptions`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders }, method: 'POST', body: new FormData() })
```
nekomeowww, line 92:
```suggestion
            responseTranscription = await fetch(`${config.baseUrl as string}audio/transcriptions`, { headers: { Authorization: `Bearer ${config.apiKey}`, ...additionalHeaders }, method: 'POST', body: new FormData() })
```

If no audio file uploaded, upstream server may still respond with 400, which means L99 may fail to check, is this expected?

Iro96: lemme check again

Iro96: i think sending empty FormData is fine, 400 still counts as a valid endpoint

PR #597, opened 2025-09-18
enhancement(stage-tamagotchi): Use lucas-oma's RAG implementation on desktop app

gg582:

PR #622, opened 2025-09-30
feat(package:memory-service): Introduce PGlite+Drizzle Memory Backend to AIRI

gg582:
## Description

Hello, this is a first commit to implement integrated memory package for AIRI that started from Idea sketch from #597. This is using PGlite as recommended option, while providing Postgres and Embedded Postgres. This uses separate LLM provider factory to organize memory, but I think that it is better to move those parts as separate package. Current organizing plan is rather primitive, which is not suitable for ready-for-use AI VTuber.

- So, at first we should refactor huge amounts of code.
- Second, since this project need to support 3rd party memory, it should be connected with multi-- backend wrappers.
Third, Implement actual Long-Term and Short-Term memory.

## Linked Issues

#387

## Additional Context

This is a part of memory service.
Different parts(e.g. App Integration, Front-end) should be followed by other pull requests.


packages/memory-service/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "@xsai/embed": "catalog:",
+    "@xsai/generate-text": "catalog:",
+    "cors": "^2.8.5",
+    "dotenv": "^17.2.1",
+    "drizzle-orm": "^0.44.5",
+    "elysia": "^1.4.9",
+    "embedded-postgres": "17.6.0-beta.15",
+    "helmet": "^8.0.0",
+    "multer": "^2.0.2",
+    "openai": "^4.28.0",
```
nekomeowww: We should remove this.

packages/memory-service/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "pg": "^8.16.3",
+    "uuid": "^11.1.0",
+    "valibot": "1.0.0-beta.9",
+    "ws": "^8.18.0",
+    "zod": "^3.23.8"
```
nekomeowww: We should use valibot or zod only instead of both two.

packages/memory-service/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "drizzle-orm": "^0.44.5",
+    "elysia": "^1.4.9",
+    "embedded-postgres": "17.6.0-beta.15",
+    "helmet": "^8.0.0",
+    "multer": "^2.0.2",
+    "openai": "^4.28.0",
+    "pg": "^8.16.3",
+    "uuid": "^11.1.0",
+    "valibot": "1.0.0-beta.9",
+    "ws": "^8.18.0",
```
nekomeowww: Let's use crossws?

packages/memory-service/package.json
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "scripts": {
+    "dev": "pnpm run build",
+    "build": "tsdown"
+  },
+  "dependencies": {
+    "@electric-sql/pglite": "^0.3.9",
+    "@electric-sql/pglite-tools": "^0.2.15",
+    "@elysiajs/cors": "^1.4.0",
+    "@google/generative-ai": "^0.21.0",
```
nekomeowww: We don't need this I guess

packages/memory-service/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "@proj-airi/memory-service",
+  "type": "module",
+  "version": "0.7.2-beta.3",
+  "description": "Centralized memory service for AIRI with PostgreSQL + pgvector",
+  "author": {
+    "name": "lucas-oma"
+  },
```
nekomeowww: You are the author too, you can make this an array.

packages/memory-service/package.json
gg582: okay!

packages/memory-service/package.json
gg582: google generative ai is unnecessary.
```
yjlee@elegant:~/airi/packages/memory-service$ grep -rl electric-sql/pglite
package.json
dist/index.js
src/api/memory.ts
src/db/index.ts
yjlee@elegant:~/airi/packages/memory-service$ grep -rl electric-sql/pglite-tools
package.json
dist/index.js
src/api/memory.ts
yjlee@elegant:~/airi/packages/memory-service$ grep -rl elysiajs/cors
package.json
dist/index.js
src/api/server.ts
```
these three is necessary

packages/memory-service/package.json
gg582: those were called by openai. i removed them

packages/memory-service/package.json
gg582: removed now

packages/memory-service/package.json
gg582: Thanks, I added my id.

packages/memory-service/src/api/memory.ts
```diff
@@ -0,0 +1,228 @@
...
+      const envVars = parsePwdEnv(dbUrl)
+
+      const name = (file.name || '').toLowerCase()
+      if (!name.endsWith('.sql')) {
+        set.status = 400
+        return 'For Postgres/Embedded-Postgres, upload a .sql dump.'
+      }
+
+      await new Promise<void>((resolve, reject) => {
+        exec(`psql "${dbUrl}" -f "${tmpPath}"`, { env: envVars }, (error, _so, stderr) => {
```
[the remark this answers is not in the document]
gg582: okay...this is really crucial.
@nekomeowww 
how do you think about this part?
It may take time but I think it must be removed, even if it is not for security reason.

packages/memory-service/drizzle/0000_sharp_iceman.sql
```diff
@@ -0,0 +1,254 @@
...
+CREATE INDEX "memory_items_content_vector_1536_index" ON "memory_fragments" USING hnsw ("content_vector_1536" vector_cosine_ops);--> statement-breakpoint
+CREATE INDEX "memory_items_content_vector_1024_index" ON "memory_fragments" USING hnsw ("content_vector_1024" vector_cosine_ops);--> statement-breakpoint
+CREATE INDEX "memory_items_content_vector_768_index" ON "memory_fragments" USING hnsw ("content_vector_768" vector_cosine_ops);--> statement-breakpoint
+CREATE INDEX "memory_items_memory_type_index" ON "memory_fragments" USING btree ("memory_type");--> statement-breakpoint
+CREATE INDEX "memory_items_category_index" ON "memory_fragments" USING btree ("category");--> statement-breakpoint
+CREATE INDEX "memory_items_importance_index" ON "memory_fragments" USING btree ("importance");--> statement-breakpoint
+CREATE INDEX "memory_items_created_at_index" ON "memory_fragments" USING btree ("created_at");--> statement-breakpoint
+CREATE INDEX "memory_items_last_accessed_index" ON "memory_fragments" USING btree ("last_accessed");--> statement-breakpoint
+CREATE INDEX "memory_items_episode_id_index" ON "memory_fragments" USING btree ("episode_id");--> statement-breakpoint
```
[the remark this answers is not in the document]
gg582: I think that this is just okay..

packages/memory-service/src/index.ts
```diff
@@ -0,0 +1,176 @@
...
+      catch (err) {
+        console.error('[memory-service] Handler error:', err)
+        res.statusCode = 500
+        res.end('Internal Server Error')
+      }
+    })
+
+    server.listen(PORT, () => {
+      console.warn(`[memory-service] HTTP server listening on http://localhost:${PORT}`)
+    })
```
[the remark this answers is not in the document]
gg582: minor but important. key point that we should keep is duplex: 'half' option while rewriting to elysia

packages/memory-service/src/db/schema.ts
```diff
@@ -0,0 +1,293 @@
...
+  mem_llm_provider: text().notNull().default('openai'),
+  mem_llm_model: text().notNull().default('gpt-3.5-turbo'),
+  mem_llm_api_key: text().notNull(),
+  mem_llm_temperature: integer().notNull().default(7), // 0-10 scale, divide by 10 for actual temp
+  mem_llm_max_tokens: integer().notNull().default(2000),
+
+  // Embedding Settings
+  mem_embedding_provider: text().notNull().default('openai'),
+  mem_embedding_model: text().notNull().default('text-embedding-3-small'),
+  mem_embedding_api_key: text().notNull(),
```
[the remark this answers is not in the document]
gg582: @nekomeowww  which solution can be a good method on local application? AWS or other cloud based secrets managers need additional subscription, and cloud subscription is quite expensive to require it as a dependency.

LemonNekoGH:
Thank you, but it's too large, I am a little lost, where should I start to review?

gg582:
Okay...then is it okay to split PR again

gg582:
First, you can review memory-service/src/db.

gg582:
and you can check `llm-providers` that is using XSAI. this part is different from lucas-oma's commit

gg582:
I have coding test on Oct 11th...I will gradually attach a review guide after it

gg582:
And I guess that it will be better to remove all the sources as you mentioned on draft...
However, I am confused if this implementation will be garbaged...if it is not, I should attach package.json, etc first...


And make model card separation logic.

gg582:
First, I will overwrite a commit while removing all files 3-5 files

gg582:
@nekomeowww I guess that it is better to restart with basic idea sketches...
Now basic sketch codes are remained.

gg582:
```
yjlee@yjlee-linuxonmac:~/airi/packages/memory-service/src/services$ tree
.
├── background-trigger.ts
├── context-builder.ts
├── embedding-providers
│   ├── base.ts
│   ├── factory.ts
│   └── xsai.ts
├── llm-memory-manager.ts
├── llm-providers
│   ├── base.ts
│   ├── factory.ts
│   ├── fallback.ts
│   └── xsai.ts
├── memory.ts
├── message-ingestion.service.ts
├── message-processing.ts
├── prompts.ts
└── settings.ts

```

This was a service tree from previous proposal...
But I think that it is better to rewrite whole message ingestion/management while retaining xsai embedding providers/

gg582:
Also, I've checked pull request #636 ; I realized that my commit was also too huge, but in that PR there are no explanations... I think that it is better to check #636's code as good reference and rewrite with XSAI/PGVector implementation that has distinguish Model card using Model Name...So, in my opinion, it is better to share memory between models with same name tag(end user will recognize them as a same character)..
So, I want to hear your opinion...Have a good zhongchujie, and can you look at my message after holidays?

gg582:
Also, after completing memory integration, I propose that making a memory docker image can be a good solution for hosting service/cloud utilization. This can be a good advantage for professional users.

packages/memory-service/tsconfig.json
```diff
@@ -0,0 +1,31 @@
...
+    "lib": [
+      "ESNext"
+    ],
+    "moduleDetection": "force",
+    "baseUrl": ".",
+    "rootDir": "src",
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "paths": {
+      "@telegram-bot/*": ["../telegram-bot/src/*"]
```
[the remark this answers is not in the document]
nekomeowww: What is this package?

PR #646, opened 2025-10-09
feat(nix/stage-tamagotchi): update to electron rewrite

Weathercold:
## Description

currently the app only shows a transparent window with different error messages. But the nix package is in a non-functional state anyway.


nix/package.nix
```diff
@@ -1,152 +1,76 @@
...
+  installPhase = ''
+    runHook preInstall
 
-  postInstall = lib.optionalString stdenv.isDarwin ''
-    mkdir -p "$out/bin"
-    ln -sf "$out/Applications/AIRI.app/Contents/MacOS/airi" "$out/bin/airi"
-  '';
+    mkdir -p "$out/opt"
+    cp -r dist/*-unpacked "$out/opt/AIRI"
+    # The icon is actually 1500x1500... install it anyway
```
nekomeowww, line 70: Do you need other sizes of the icon?

nix/package.nix
```diff
@@ -1,152 +1,76 @@
...
-  ++ lib.optionals stdenv.isLinux [
-    alsa-lib # Used by alsa-sys
-    atk # Used by atk-sys
-    libayatana-appindicator # Used by libappindicator-sys
-    systemdLibs # For libudev used by libudev-sys
-    webkitgtk_4_1
-    xorg.libXtst # Used by x11
+  desktopItems = [
+    (makeDesktopItem {
+      desktopName = "AIRI";
```
nekomeowww, line 26: Will this conflict with executable name change in electron-builder?

Weathercold: no, this is the name of the application as displayed to the user. The executable name is specified by `exec`

Weathercold:
The good news is that with the addition of buttons, I can open settings now. But the canvas is still empty, keyboard shortcuts still don't work, and the console is filled with errors.

image
image
image

nekomeowww:
- Keyboard shortcuts wasn't implemented, later on you can test it under system -> developer -> useMagicKeys:

image

Errors, I know this issue, @kwaa mentioned this privately too, but haven't had time to fix it, perhaps @Iro96 could re-design it when refactoring.

But why the canvas is still empty? Are there any errors? Which driver? Since we are on Electron already yet the canvas issue exists, then I believe it's not something I could help. At least it's working on my Debian + GNOME machine, (4080Super involved).

nekomeowww:
We are stable now for building process. Do you need help to test this PR?

Weathercold:
I'm currently busy; I'll update this PR in a few days. I'm not sure whether it can be finished before v0.8.0 releases though. It's probably not worth it to delay the release for this PR

nekomeowww:
Take your time, we still have small changes WIP.

PR #653, opened 2025-10-15
Emergent

JohnPion-Zhang:
## Description

create RN frontend, remove desktop and web frontends

### DIFF
--- a/apps/stage-mobile/package.json
+++ b/apps/stage-mobile/package.json
@@ -10,7 +10,7 @@
     "url": "https://github.com/moeru-ai"
   },
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


Weathercold:
We have no plans for a react native frontend. Not sure why you think this is a good idea considering all of the frontend is written in vue.

nekomeowww:
> We have no plans for a react native frontend. Not sure why you think this is a good idea considering all of the frontend is written in vue.

Assumption without any clues: it looks like the author is trying to use coding agent to refactor the project not only to React Native but rewrite all, maybe for their own proprietary project without feeding back to the community.

PR #658, opened 2025-10-17
fix(openai-compatible-builder): ignore embed & gemini-2.5-pro model

### LINKED ISSUE #649: Google Gemini API key keeps failing health check
### Describe the bug

During the initial onboarding screen where the users are asked to choose their AI provider of their choice, I am unable to choose Google Gemini even with a valid API key. The health checks keep failing as it attempts to perform it on a model named `embedding-gecko-001` which seems outdated and does not exist in the Gemini model suite anymore. Even if it does, by the looks of it, the code expects an output back as text which I don't think the embedding models are capable of providing. Ideally the user should be able to choose the model provider as well as the model of their choice, and then the health check is to be performed. Models available and providers list's maintenance can probably be outsourced to an external 3rd party library as well

### System Info

```Shell
I tried it out on the hosted website, I haven't setup the repo locally yet
```

### Validations

### Contributions

kwaa:
fix #649 

image

### DIFF
--- a/packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
+++ b/packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
@@ -126,6 +126,8 @@ export function buildOpenAICompatibleProvider(
             [
               // exclude embedding models
               'embed',
+              // exclude tts models, specifically for OpenAI
+              'tts',
               // bypass gemini pro quota
               // TODO: more elegant solution
               'gemini-2.5-pro',


packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
```diff
@@ -122,7 +122,17 @@ export function buildOpenAICompatibleProvider(
             Authorization: `Bearer ${apiKey}`,
           },
         })
-        if (models && models.length > 0)
+          .then(models => models.filter(model =>
+            [
+              // exclude embedding models
+              'embed',
```
nekomeowww, line 128:
```suggestion
              // exclude embedding models
              'embed',
              // exclude tts models, specifically for OpenAI
              'tts',
```

PR #663, opened 2025-10-18
fix(stage-tamagotchi): no drag-and-click for linux

sumimakito:
Previously, [noobfromph/electron-click-drag-plugin](https://github.com/noobfromph/electron-click-drag-plugin) was introduced by https://github.com/moeru-ai/airi/pull/650 to allow the DOM elements with `app-region: drag` to continue receiving mouse events so that tooltips can be displayed.

However, this does not work on some desktop setups in Linux. Although the plugin does seem to have some API usage, we agreed to disable this and fall back to the default `app-region: drag` behavior on Linux.

### DIFF
--- a/apps/stage-tamagotchi/src/preload/index.ts
+++ b/apps/stage-tamagotchi/src/preload/index.ts
@@ -12,6 +12,7 @@ const api = {}
 if (contextIsolated) {
   try {
     contextBridge.exposeInMainWorld('electron', electronAPI)
+    contextBridge.exposeInMainWorld('platform', platform)
     contextBridge.exposeInMainWorld('api', api)
   }
   catch (error) {


apps/stage-tamagotchi/src/renderer/components/Widgets/ControlsIsland/index.vue
```diff
@@ -10,6 +10,7 @@ import ControlButton from './ControlButton.vue'
 import ControlButtonTooltip from './ControlButtonTooltip.vue'
 
 import { electronOpenSettings, electronStartDraggingWindow } from '../../../../shared/eventa'
+import { isLinux } from '../../../utils/platform'
```
nekomeowww, line 13: Wait you can do this in renderer thread?

apps/stage-tamagotchi/src/renderer/utils/platform.ts
```diff
@@ -0,0 +1 @@
+export const isLinux = /^linux/i.test(window.platform)
```
nekomeowww, line 1: Oh, ok.

apps/stage-tamagotchi/src/renderer/components/Widgets/ControlsIsland/index.vue
sumimakito: 🪄

PR #668, opened 2025-10-19
fix(stage-tamagotchi,stage-shared,stage-ui): unify electron-renderer globals and fix env checks in providers

sumimakito:
- Exposed a unified d.ts at `@proj-airi/stage-shared/electron-renderer` for globals like `window` for _preload_ (running in the renderer process) and _renderer_ in Electron
- Corrected type usages in _preload_
- Fixed environment checks in providers
- Made `RUNTIME_ENVIRONMENT` available to all modes

### DIFF
--- a/apps/stage-tamagotchi/src/preload/index.ts
+++ b/apps/stage-tamagotchi/src/preload/index.ts
@@ -20,10 +20,7 @@ if (contextIsolated) {
   }
 }
 else {
-  // @ts-expect-error (define in dts)
   window.electron = electronAPI
-  // @ts-expect-error (define in dts)
   window.platform = platform
-  // @ts-expect-error (define in dts)
   window.api = api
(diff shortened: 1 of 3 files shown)


apps/stage-tamagotchi/electron.vite.config.ts
```diff
@@ -74,13 +74,11 @@ export default defineConfig({
     plugins: [
       {
         name: 'proj-airi:define-runtime-environment',
-        config(ctx) {
-          if (ctx.mode === 'production') {
```
sumimakito, line 78: Unsure if this was intended

nekomeowww: I can't remember it correctly, seems adding more comments help us figuring out the reasons of having code here.

PR #669, opened 2025-10-21
docs: add docs for release versions

reonokiy:
## Description

Add docs for releases, pre-releases and nightly builds to let users download correct versions better.


docs/.vitepress/data/releases.data.ts
nekomeowww, line 1: One problem we may encounter if we use content loader is that, if no recent docs site pushed, while new nightly built generated, there will be no update to the docs. But since new features must come along with new docs... currently it will work without any problems.

docs/content/en/docs/manual/versions/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+**Stable releases are best for:**
+- Production use
+- Users who want a stable experience
+- Live streaming and content creation
+
+**Pre-releases are best for:**
+- Early adopters who want to try new features
+- Testing and providing feedback to developers
+- Users comfortable with occasional bugs
```
nekomeowww: How about turn this into a tooltip, and attach the tooltip component located on the top right corner of ReleaseDownloads cards?

nekomeowww:
I updated some of the styles of the documentation site, it improves drastically I believe, let me help you to bump it.

nekomeowww:
GitHub is crashing... 
image

docs/.vitepress/data/releases.data.ts
reonokiy: Maybe we can just trigger the docs build action after every nightly build. It might be a bit of a waste of resources, but I think one extra build per day is acceptable.

nekomeowww: Not sure how to trigger but sounds reasonable.

docs/content/en/docs/manual/versions/index.md
reonokiy: i just put them on the card, i think it's better for user to compare two versions at once

docs/.vitepress/data/releases.data.ts
reonokiy: netlify has build hooks, just curl webhook endpoint after nightly build finished in  the build action. https://docs.netlify.com/build/configure-builds/build-hooks/

PR #671, opened 2025-10-24
feat(consciousness): Add button to configure new provider even when one exists

Iro96:
Add button to configure new provider even when one exists

### DIFF
--- a/packages/stage-pages/src/pages/settings/modules/consciousness.vue
+++ b/packages/stage-pages/src/pages/settings/modules/consciousness.vue
@@ -67,7 +67,6 @@ function updateCustomModelName(value: string) {
               :title="metadata.localizedName || 'Unknown'"
               :description="metadata.localizedDescription"
             />
-          </fieldset>
           <RouterLink
             to="/settings/providers"
             border="2px solid"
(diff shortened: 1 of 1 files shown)


Iro96:
what will happen when i use more than 1 provider at same time? Should we replace current provider when we switch to another provider? cc @nekomeowww @kwaa

nekomeowww:
> what will happen when i use more than 1 provider at same time? Should we replace current provider when we switch to another provider? cc @nekomeowww @kwaa

Multiple providers can be used simultaneously. For example, you may wish to use OpenAI's vision model for images specifically with Ollama to run small models to work with body control and game control. This is even more critical for embedding models, since embedding compatibility is bad for existing models, we cannot just simply swap to another embedding model for existing database, therefore, multiple embedding models may work together).

Iro96:
> > what will happen when i use more than 1 provider at same time? Should we replace current provider when we switch to another provider? cc @nekomeowww @kwaa
> 
> Multiple providers can be used simultaneously. For example, you may wish to use OpenAI's vision model for images specifically with Ollama to run small models to work with body control and game control. This is even more critical for embedding models, since embedding compatibility is bad for existing models, we cannot just simply swap to another embedding model for existing database, therefore, multiple embedding models may work together).

thanks for explaining, everything is fine now

Issue #652, opened 2025-10-14
A more clear error message for LMstudio


AdairLi2504:
### Clear and concise description of the problem

According to #432, LMstudio server needs to enable CORS. However, it is not mentioned on the docs or error message and LMstudio does not enable CORS by default. 

### Suggested solution / Ideas

Could I add a new sentence to error message such as 
`Also please make sure you open Enable CORS on the settings according to the issue 432`

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Iro96:
lemme add it this pr https://github.com/moeru-ai/airi/pull/648

AdairLi2504:
OK thank you

PR #687, opened 2025-10-30
refactor(validation): improve provider validation and error handling

Iro96:
## Description

-  Show "API Key is required" first when if there aren't any api key provided before instead of health check directly
- Add error message formatter to normalize API validation errors
- Improve DeepSeek authentication error messages
- Clean up JSON error responses in UI
- Centralize error handling in providers.ts
- Add consistent error formatting across all providers

some images:
image
image
image
image

## Linked Issues

- https://github.com/moeru-ai/airi/issues/652

### DIFF
--- a/packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
+++ b/packages/stage-ui/src/stores/providers/openai-compatible-builder.ts
@@ -4,7 +4,11 @@ import { generateText } from '@xsai/generate-text'
 import { listModels } from '@xsai/model'
 import { message } from '@xsai/utils-chat'
 
-import { formatErrorForUser, buildValidationResult } from '../../utils/providerValidation'
+import {
+  formatErrorForUser,
+  buildValidationResult,
+  buildFetchErrorResult,
+} from '../../utils/providerValidation'
 
 type ProviderCreator = (apiKey: string, baseUrl: string) => any
 
@@ -44,7 +48,7 @@ export function buildOpenAICompatibleProvider(
     ...rest
   } = options
 
-  // Default capabilities if provider does not define its own
+  // --- Default capabilities if provider does not define its own ---
   const finalCapabilities = capabilities || {
     listModels: async (config: Record<string, unknown>) => {
       // Safer casting of apiKey/baseUrl (prevents .trim() crash if not a string)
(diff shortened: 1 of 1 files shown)


packages/stage-ui/src/utils/error-formatter.ts
```diff
@@ -0,0 +1,151 @@
...
+const redactPatterns: RedactRule[] = [
+  // Mask OpenAI-like secret keys (sk-...)
+  { pattern: /sk-[\w.-]{8,}/g, replacement: 'sk-(redacted)' },
+
+  // Mask Bearer tokens
+  { pattern: /Bearer\s+[\w.\-=:]+/g, replacement: 'Bearer (redacted)' },
+
+  // Mask long hex-like or secret-looking tokens
+  { pattern: /([\w-]{20,})/g, replacement: (m: string) => (m.length > 8 ? '(redacted)' : m) },
+]
```
nekomeowww: Emmm, it seems I didn't commented out. We are the only one understand what are secrets, we could use https://effect.website/docs/data-types/redacted/ to help masking the secrets and sensitive data.

packages/stage-ui/src/utils/error-formatter.ts
```diff
@@ -0,0 +1,151 @@
...
+          small[k] = String(v)
+      }
+      return JSON.stringify(small)
+    }
+    return String(obj)
+  }
+  catch {
+    return String(obj)
+  }
+}
```
nekomeowww: Why?

packages/stage-ui/src/utils/error-formatter.ts
```diff
@@ -0,0 +1,151 @@
...
+      const v = c[k]
+      if (typeof v === 'string' && v.trim()) return v
+      if (typeof v === 'number' || typeof v === 'boolean') return String(v)
+    }
+
+    return null
+  }
+
+  return null
+}
```
nekomeowww: I think it's not our responsibility for parsing this. For OpenAI compatible ones, let's follow the OpenAI error object pattern, extract error message for that is enough, for rest of the incompatible-errors, extract the object. And sometimes, the providers use Cloudflare/Vercel as their CDN, if their service went down, the Cloudflare responds `text/html` or HTML as response, there is no reason to handle this.

If the provider is well-known and they decides not to follow the OpenAI standards, then we can add special error handlers to the builder or Provider object to handle them case by case.

Iro96:
I think I need to think again

PR #694, opened 2025-10-31
refactor(ui): remove unocss attributify for `FieldRange`

BeanDz:
… names


packages/ui/src/components/Form/Field/FieldRange.vue
```diff
@@ -17,29 +17,29 @@ const modelValue = defineModel<number>({ required: true })
 </script>
 
 <template>
-  <props.as flex="~ col gap-4">
-    <div flex="~ row" items-center gap-2>
-      <div flex="1">
-        <div class="flex items-center gap-1 text-sm font-medium">
+  <props.as :class="['flex', 'flex-col', 'gap-4']">
```
LemonNekoGH:
```suggestion
  <props.as :class="['flex flex-col gap-4']">
```

They are in the same group, no need to separate to multiple strings.

PR #714, opened 2025-11-02
fix(stage-ui|chat): restore streaming output; wrap streamText so finish/error truly gate completion

s3d-i:
## Description

**Bug (before)**

* `useLLM().stream()` resolved immediately after registering handlers, so `send()`’s `finally` ran too early and set `sending=false`.
  Result: the ChatBox **stopped showing tokens mid-stream**

**What changed**

* Wrap `streamFrom()` in a Promise that **resolves only on `onEvent('finish')`** and **rejects on `onEvent('error')`**; also `await onStreamEvent('finish')` before resolving to preserve ordering.
* Move finalization (`parser.end()`, persisting the assistant message, TTS flush, hooks) to **after `await stream(...)`**, giving a single, race-free exit point.
* Remove the “drain & drop” of `res.textStream`; use the **finish event as the sole completion signal**.
* Add `sanitizeMessages()` so any `{ role: 'error' }` entries are safely converted to `{ role: 'user', content: 'User encountered error: ...' }`. This preserves the original “error-as-intent” behavior without unsafe assertions.

**Fix & benefits**

* ✅ **Streaming restored**: `sending` is no longer flipped early; tokens render continuously until the true end.
* ✅ **Correctness**: `await stream()` now aligns with the stream lifecycle (finish/error), eliminating premature `finally`.
* ✅ **Fewer races**: single finalization point avoids double writes and late tail truncation.
* ✅ **Less overhead**: no meaningless textStream draining.
* ✅ **Type safety**: remove unsafe `messages.map(... as Message)`; only targeted sanitize for `error`.

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/components/ChatHistory.vue
+++ b/apps/stage-tamagotchi/src/renderer/components/ChatHistory.vue
@@ -3,7 +3,7 @@ import { MarkdownRenderer } from '@proj-airi/stage-ui/components'
 import { useChatStore } from '@proj-airi/stage-ui/stores/chat'
 import { useBroadcastChannel } from '@vueuse/core'
 import { storeToRefs } from 'pinia'
-import { nextTick, ref, watch } from 'vue'
+import { ref, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 const chatHistoryRef = ref<HTMLDivElement>()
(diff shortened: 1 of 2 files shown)


s3d-i:
corrected effect timing: replaced `nextTick()` using nested `requestAnimationFrame` to fix the following (as shown in video):
1. occasional rebounce issue at the end of streaming
2. occasional airi's sending placeholder invisible


https://github.com/user-attachments/assets/7f61015d-d006-4f8a-85ae-6ef15aeb7318



https://github.com/user-attachments/assets/8f858c59-369f-42c2-a6ba-f882ae53e5cc

PR #715, opened 2025-11-03
fix(stage-tamagotchi): fix the behavior when config is invalid

typed-sigterm:
There is an extra `}` in my local config (idk why)

```
{"windows":[{"title":"AIRI","tag":"main","x":995,"y":295,"width":451,"height":603}]}}
```

This patch avoids the error caused by invalid config to throw error in the promise, resulting to unhandled promise rejection and no windows are shown. But I'm not sure if it's okay to `await` here.


apps/stage-tamagotchi/src/main/index.ts
```diff
@@ -151,7 +151,7 @@ app.whenReady().then(async () => {
     callback: noop,
   })
 
-  injecta.start()
+  await injecta.start()
```
nekomeowww:
```suggestion
  injecta.start().catch(err => console.error(err))
```

typed-sigterm: Should it show an error dialog or terminate the process?

nekomeowww: Nope. It should be able to auto recover.

PR #739, opened 2025-11-14
feat(i18n): BeatSync zh , fix helpinfo

liliumneko:
## Description
Added Chinese i18n for beat_sync, Fixed broken helpinfo i18n.


packages/i18n/src/locales/zh-Hans/settings.yaml
```diff
@@ -328,6 +375,9 @@ pages:
           title: 声音配置
     description: LLM，语音合成，语音识别服务来源等
     provider:
+      aliyun-nls:
+        title: 阿里云 NLS
+        description: nls-console.aliyun.com
```
nekomeowww:
```suggestion
```

PR #743, opened 2025-11-18
feat(stage-ui): improve animation delay in provider settings

Mr-Quin:
## Description

I just found this lovely project, while setting it up I found the provider settings' animation takes too long and is annoying when going back and forth between settings because the animation always cascades down from the top. This PR tries to address that by having the animation start at the previously visited provider and ripple outwards.

Before
https://github.com/user-attachments/assets/d6632f88-f434-4d4c-97ab-ee99aa9a3f79

After
https://github.com/user-attachments/assets/fd41fb89-eb6e-47a3-a47f-1d7165e65b99

Not sure about where to put the store so I left it in the page's folder.


nekomeowww:
This is the most innovative UI changes that happen recently! Never thought it could be designed like this.

Mr-Quin:
@nekomeowww i can generalize it to other pages if you'd like

PR #751, opened 2025-11-20
feat(stage-*): apply ripple grid to other pages

Mr-Quin:
## Description

Lifted the previously introduced "ripple" animation #743 into a component and applied it to these pages:
- /settings
- /settings/modules
- /settings/system
- /settings/providers

Added a story for this component
Adjusted header animation to be in line with this change


nekomeowww:
Conflict.

Mr-Quin:
@nekomeowww resolved

PR #752, opened 2025-11-20
feat(stage-tamagotchi): memorize scroll position

Mr-Quin:
## Description

On electron, since the scroll element is not `document`, scroll position is lost when navigating between pages.
This PR fixes that.

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/App.vue
+++ b/apps/stage-tamagotchi/src/renderer/App.vue
@@ -3,24 +3,33 @@ import { defineInvoke, defineInvokeHandler } from '@moeru/eventa'
 import { useDisplayModelsStore } from '@proj-airi/stage-ui/stores/display-models'
 import { useOnboardingStore } from '@proj-airi/stage-ui/stores/onboarding'
 import { useSettings } from '@proj-airi/stage-ui/stores/settings'
+import { useDark } from '@vueuse/core'
 import { storeToRefs } from 'pinia'
 import { onMounted, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
-import { RouterView, useRouter } from 'vue-router'
+import { RouterView, useRoute, useRouter } from 'vue-router'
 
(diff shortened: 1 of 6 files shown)


apps/stage-tamagotchi/src/renderer/composables/use-restore-scroll.ts
```diff
@@ -0,0 +1,30 @@
...
+
+      const savedPosition = scrollPositions.get(newPath) || 0
+      scrollContainer.value.scrollTop = savedPosition
+    },
+  )
+
+  return {
+    scrollContainer,
+  }
+}
```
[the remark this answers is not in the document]
nekomeowww: I think it's reasonable for this suggestion. @Mr-Quin would you mind take a look?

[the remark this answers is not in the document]
Mr-Quin: @nekomeowww implemented this suggestion

apps/stage-tamagotchi/src/renderer/composables/use-restore-scroll.ts
```diff
@@ -0,0 +1,36 @@
...
+      }
+
+      await nextTick()
+
+      if (!scrollContainer.value) {
+        return
+      }
+
+      const savedPosition = scrollPositions.get(newPath) || 0
+      scrollContainer.value.scrollTop = savedPosition
```
nekomeowww, line 29: https://vueuse.org/core/useScroll/#usescroll

Will it be better if we use `useScroll` from VueUse?

Mr-Quin: I don't think so, we can pull `y` from `useScroll` instead of reading `scrollTop`, but I don't really see a point. 
It also adds a bunch of event listeners to the element that we don't need.

nekomeowww: Sure.

PR #753, opened 2025-11-20
chore(stage-*): improve use of `useDark`, remove `nprogress` for tamagotchi

Mr-Quin:
## Description

Noticed a subtle performance issue with quick easy fix.

What this PR does:
- Wraps `useDark` in a composable so `useDark` is only called once, the returned ref is reused.
- Set `disableTransition` to true everywhere
- Removed nprogress (opinionated, but I don't see the progress bar and it's not used in the electron app)

Why:

When navigating, nprogress modifies class on the document `html` element (because it's calling `NProgress.start` at the router level). This causes selectors selecting `.dark` (because it's also on the `html` element) to invalidate and this slows things down. 

Similarly, each call to `useDark` (those without `disableTransition`) ends up injects a <style> tag temporarily which also causes a css invalidation. Wrapping `useDark` means we only call it once in the app.

This improvement is noticeable if your device is slower (I notice it in 4x slow down, noticeably snappier after the changes)

image

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -44,7 +44,7 @@ catalogs:
       version: 0.4.0-beta.9
     '@xsai/shared':
       specifier: ^0.4.0-beta.9
-      version: 0.4.0-beta.9
+      version: 0.4.0-beta.10
     '@xsai/shared-chat':
       specifier: ^0.4.0-beta.9
       version: 0.4.0-beta.9
@@ -355,7 +355,7 @@ importers:
         version: 0.4.0-beta.9
       '@xsai/shared':
         specifier: 'catalog:'
-        version: 0.4.0-beta.9
+        version: 0.4.0-beta.10
       '@xsai/shared-chat':
         specifier: 'catalog:'
         version: 0.4.0-beta.9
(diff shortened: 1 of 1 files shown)


packages/ui/src/composables/use-theme.ts
nekomeowww, line 1: Use `use-theme.ts`

apps/stage-web/src/main.ts
```diff
@@ -32,15 +31,6 @@ if (import.meta.env.VITE_APP_TARGET_HUGGINGFACE_SPACE)
...
 
-router.beforeEach((to, from) => {
-  if (to.path !== from.path)
-    NProgress.start()
-})
-
-router.afterEach(() => {
-  NProgress.done()
-})
-
```
nekomeowww, line 43: I guess we can keep it for `stage-web`?

packages/ui/src/composables/use-theme.ts
Mr-Quin: done

apps/stage-web/src/main.ts
Mr-Quin: reverted.
maybe nprogress can be scoped to a separate container?

PR #754, opened 2025-11-22
perf(stage-web): improve chat responsiveness & loading order

Mr-Quin:
## Description

On the website (desktop), if chat has a lot of content, navigating to the home page gets pretty slow. For example if I'm on settings page and I click go back, it freezes for a bit before the home page renders, because chat takes time to render.

Before:
https://github.com/user-attachments/assets/555d487c-7b8a-4e25-ba02-002f323bf49b

The PR makes it so that chat rendering is deferred a little bit to allow navigation to complete, and show a loading indicator while chat is rendering. Overall render time doesn't change, but now navigation is almost immediate.

After:
https://github.com/user-attachments/assets/dfbbbb1d-5763-4c8d-9402-979590d55bdd

Changes:
- Introduced deferred rendering
- Refactored InteractiveArea into separate components to make the logic a bit cleaner, although this isn't strictly necessary for this change
- Added a chat loading indicator
- Doesn't affect MobileInteractiveArea

Recorded at 4x slow down.

The PR doesn't address the slowness with the markdown renderer itself.

### DIFF
--- a/packages/ui/src/composables/use-deferred-mount.ts
+++ b/packages/ui/src/composables/use-deferred-mount.ts
@@ -4,6 +4,9 @@ export function useDeferredMount() {
   const isReady = ref(false)
 
   onMounted(() => {
+    // NOTICE: yield the main thread to the browser to paint.
+    // Other approaches work too, e.g. double RAF like here, or a setTimeout 0, or Suspense with async component, etc. 
+    // Virtualization somewhat helps, however, virtual list may break Ctrl-F and Ctrl-A.
     requestAnimationFrame(() => {
       requestAnimationFrame(() => {
         isReady.value = true


packages/ui/src/composables/use-deferred-mount.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function useDeferredMount() {
+  const isReady = ref(false)
+
+  onMounted(() => {
+    requestAnimationFrame(() => {
+      requestAnimationFrame(() => {
+        isReady.value = true
+      })
+    })
```
nekomeowww, line 14: I don't get it. Could wait for 2 frame callback solve this completely? What if more frames is required. I think what we need is virtual list nor deferred mount as described here.

Mr-Quin: The thing that this is trying to optimize for is time to paint.

If you have a chat with 5 Markdown blocks, and each takes 1 second to render, when you go to the chat page it's going to block the main thread for 5s before it paints the page, i.e. browser freeze for 5s when I click on that back button, then it show the chat page.

What this does is it defers rendering to AFTER the page has painted, then it freezes for that 5 seconds, the difference is that user can already see the page and the loading indicator is moving so visually it's not frozen.

You can do this with double raf like here, or a setTimeout 0, or Suspense with async component, etc. The point is to yield the main thread to the browser to paint. 

If you virtualize the chat so that only 1 Markdown block renders, it's still going to block rendering for 1 second before the next page appears.

Virtualization is going to help, but not completely resolve it (you can combine the two approaches). Plus idk if you really want to virtualize the chat because then you are going to break things like ctrl-f or ctrl-a.

@nekomeowww

packages/ui/src/composables/use-deferred-mount.ts
```diff
@@ -0,0 +1,17 @@
+import { onMounted, ref } from 'vue'
+
+export function useDeferredMount() {
+  const isReady = ref(false)
+
+  onMounted(() => {
+    requestAnimationFrame(() => {
```
nekomeowww, line 10:
```suggestion
    // NOTICE: yield the main thread to the browser to paint.
    // Other approaches work too, e.g. double RAF like here, or a setTimeout 0, or Suspense with async component, etc. 
    // Virtualization somewhat helps, however, virtual list may break Ctrl-F and Ctrl-A.
    requestAnimationFrame(() => {
```

packages/ui/src/composables/use-deferred-mount.ts
nekomeowww: I get it now, thanks for explaination.

PR #758, opened 2025-11-23
fix(stage-ui): early return for Live2D drop-shadow update hook

s3d-i:
…only computes/applies filters and runs the animation loop when the shadow setting is enabled

 - Default live2dShadowEnabled to true

### DIFF
--- a/packages/stage-ui/src/components/scenes/live2d/Model.vue
+++ b/packages/stage-ui/src/components/scenes/live2d/Model.vue
@@ -488,7 +488,7 @@ function updateDropShadowFilterLoop() {
     dropShadowAnimationId.value = 0
     return
   }
-  
+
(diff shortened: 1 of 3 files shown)


packages/stage-ui/src/components/scenes/live2d/Model.vue
```diff
@@ -483,11 +484,15 @@ watch(() => props.scale, setScaleAndPosition)
 // TODO: This is hacky!
 function updateDropShadowFilterLoop() {
   updateDropShadowFilter()
+  if (!live2dShadowEnabled.value) {
+    dropShadowAnimationId.value = 0
+    return
+  }
```
nekomeowww, line 490:
```suggestion
  }
  
```

PR #760, opened 2025-11-24
feat(stage-ui): implement OPFS as cache for zipped Live2D models

Mr-Quin:
## Description

Loading l2d model right now is pretty slow because we are unzipping the zip file each time we load a model.
This PR adds a OPFS layer to cache the unzipped files so the next load is faster. 
Overall this is about 50% faster load time compared to before.

This works by injecting a middleware into live2d-display with a bit of hack to work around some of the quirks of the library.

Before
Screenshot 2025-11-23 152945

After
Screenshot 2025-11-23 152933

### DIFF
--- a/packages/stage-ui/src/utils/live2d-opfs-registration.ts
+++ b/packages/stage-ui/src/utils/live2d-opfs-registration.ts
@@ -4,7 +4,10 @@ import { OPFSCache } from './opfs-loader'
 
 const zipLoaderIndex = Live2DFactory.live2DModelMiddlewares.indexOf(ZipLoader.factory)
 
-if (zipLoaderIndex !== -1) {
+if (Live2DFactory.live2DModelMiddlewares.includes(OPFSCache.checkMiddleware)) {
+  // Middlewares already registered.
+}
+else if (zipLoaderIndex !== -1) {
   // Insert Check before ZipLoader
   Live2DFactory.live2DModelMiddlewares.splice(zipLoaderIndex, 0, OPFSCache.checkMiddleware)
   // Insert Save after ZipLoader
--- a/packages/stage-ui/src/utils/opfs-loader.ts
+++ b/packages/stage-ui/src/utils/opfs-loader.ts
@@ -22,8 +22,9 @@ export class OPFSCache {
           value: pathPrefix + file.name,
         })
         files.push(file)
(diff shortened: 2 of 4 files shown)


packages/stage-ui/src/utils/live2d-opfs-registration.ts
```diff
@@ -0,0 +1,15 @@
...
+
+if (zipLoaderIndex !== -1) {
+  // Insert Check before ZipLoader
+  Live2DFactory.live2DModelMiddlewares.splice(zipLoaderIndex, 0, OPFSCache.checkMiddleware)
+  // Insert Save after ZipLoader
+  Live2DFactory.live2DModelMiddlewares.splice(zipLoaderIndex + 2, 0, OPFSCache.saveMiddleware)
+}
+else {
+  console.warn('[OPFS] ZipLoader not found in middlewares, caching disabled')
+}
```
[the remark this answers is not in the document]
Mr-Quin: ok

packages/stage-ui/src/utils/live2d-opfs-registration.ts
```diff
@@ -0,0 +1,18 @@
+import { Live2DFactory, ZipLoader } from 'pixi-live2d-display/cubism4'
```
nekomeowww, line 1:
```suggestion
import { Live2DFactory, ZipLoader } from 'pixi-live2d-display/cubism4'
```

Cool! I never knew we could write like this!

packages/stage-ui/src/utils/opfs-loader.ts
```diff
@@ -0,0 +1,181 @@
...
+  opfsKey?: string
+}
+
+declare global {
+  interface FileSystemDirectoryHandle {
+    values(): AsyncIterableIterator<FileSystemDirectoryHandle | FileSystemFileHandle>
+  }
+}
+
+export class OPFSCache {
```
nekomeowww, line 13: Last year OPFS wasn't even ready, but it seems ok now. We failed many times when working on https://github.com/proj-airi/duckdb-wasm 🎉

https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system

nekomeowww (approved):

Issue #757, opened 2025-11-23
failed to render on arch linux


astro-angelfish:
### Describe the bug

I downloaded the deb package [here](https://github.com/moeru-ai/airi/actions/workflows/release-tamagotchi.yml)

then used [debtap](https://github.com/helixarch/debtap) to convert it to an arch package and then I installed it.

But after trying start it, it shows nothing but a tray icon, and settings ui failed to render too.

The stdout spams things like:
```
[46732:1123/144755.178813:ERROR:media/gpu/vaapi/vaapi_wrapper.cc:1631] vaInitialize failed: unknown libva error
[46732:1123/144755.180084:ERROR:ui/ozone/platform/wayland/gpu/wayland_surface_factory.cc:250] '--ozone-platform=wayland' is not compatible with Vulkan. Consider switching to '--ozone-platform=x11' or disabling Vulkan
[46694:1123/144755.389584:ERROR:ui/ozone/platform/wayland/host/wayland_wp_color_manager.cc:273] Unable to set image transfer function.
[46694:1123/144755.389600:ERROR:ui/ozone/platform/wayland/host/wayland_wp_color_manager.cc:191] Failed to populate image description for color space {primaries:BT709, transfer:SRGB, matrix:RGB, range:FULL}
```

I've also tried to use `prime-run`, installed [intel-media-driver](https://archlinux.org/packages/extra/x86_64/intel-media-driver/) and [libva-nvidia-driver](https://archlinux.org/packages/extra/x86_64/libva-nvidia-driver/) to resolve `vaInitialize failed: unknown libva error`, but the rest of errors still persists and preventing anything from being rendered.

<img width="1455" height="1500" alt="Image" src="https://github.com/user-attachments/assets/81e6ac81-a715-4274-8be9-2aa8e4bcfa91" />

### System Info

```Shell
Need to install the following packages:
envinfo@7.20.0
Ok to proceed? (y) y


  System:
    OS: Linux 6.17 Arch Linux
    CPU: (24) x64 Intel(R) Core(TM) i7-14650HX
    Memory: 36.66 GB / 46.77 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 25.2.1 - /usr/bin/node
    npm: 11.6.3 - /usr/bin/npm
  Browsers:
    Chromium: 142.0.7444.175
    Firefox: 145.0.1
    Firefox Developer Edition: 145.0.1
```

### Validations

[...]

PR #761, opened 2025-11-24
feat(stage-tamagotchi): wayland support

astro-angelfish:
## Description

airi used to fail to render anything in wayland session

adding `--enable-features=UseOzonePlatform` command option to electron seems to be working, and i copied rest of options from [here](https://github.com/mmaura/poe2linuxcompanion/blob/90664607a147ea5ccea28df6139bd95fb0ebab0e/electron/main/index.ts#L28-L46).

## Linked Issues

https://github.com/moeru-ai/airi/issues/757

### DIFF
--- a/apps/stage-tamagotchi/src/main/index.ts
+++ b/apps/stage-tamagotchi/src/main/index.ts
@@ -52,6 +52,11 @@ if (isLinux) {
   app.commandLine.appendSwitch('enable-unsafe-webgpu')
   app.commandLine.appendSwitch('enable-features', 'Vulkan')
 
+  // NOTICE: we need UseOzonePlatform, WaylandWindowDecorations for working on Wayland.
+  // Partially related to https://github.com/electron/electron/issues/41551, since X11 is deprecating now, 
+  // we can safely remove the feature flags for Electron once they made it default supported.
+  // Fixes: https://github.com/moeru-ai/airi/issues/757
+  // Ref: https://github.com/mmaura/poe2linuxcompanion/blob/90664607a147ea5ccea28df6139bd95fb0ebab0e/electron/main/index.ts#L28-L46
   if (env.XDG_SESSION_TYPE === 'wayland') {
     app.commandLine.appendSwitch('enable-features', 'GlobalShortcutsPortal')
 


apps/stage-tamagotchi/src/main/index.ts
```diff
@@ -51,6 +51,13 @@ if (isLinux) {
   app.commandLine.appendSwitch('enable-features', 'SharedArrayBuffer')
   app.commandLine.appendSwitch('enable-unsafe-webgpu')
   app.commandLine.appendSwitch('enable-features', 'Vulkan')
+
+  if (env.XDG_SESSION_TYPE === 'wayland') {
+    app.commandLine.appendSwitch('enable-features', 'GlobalShortcutsPortal')
+
+    app.commandLine.appendSwitch('enable-features', 'UseOzonePlatform')
+    app.commandLine.appendSwitch('enable-features', 'WaylandWindowDecorations')
+  }
```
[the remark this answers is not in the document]
astro-angelfish: nope, using lists would make these `enable-feature` flags useless. things breaks again.

apps/stage-tamagotchi/src/main/index.ts
```diff
@@ -51,6 +51,13 @@ if (isLinux) {
   app.commandLine.appendSwitch('enable-features', 'SharedArrayBuffer')
   app.commandLine.appendSwitch('enable-unsafe-webgpu')
   app.commandLine.appendSwitch('enable-features', 'Vulkan')
+
+  if (env.XDG_SESSION_TYPE === 'wayland') {
```
nekomeowww, line 60:
```suggestion
  // NOTICE: we need UseOzonePlatform, WaylandWindowDecorations for working on Wayland.
  // Partially related to https://github.com/electron/electron/issues/41551, since X11 is deprecating now, 
  // we can safely remove the feature flags for Electron once they made it default supported.
  // Fixes: https://github.com/moeru-ai/airi/issues/757
  // Ref: https://github.com/mmaura/poe2linuxcompanion/blob/90664607a147ea5ccea28df6139bd95fb0ebab0e/electron/main/index.ts#L28-L46
  if (env.XDG_SESSION_TYPE === 'wayland') {
```

nekomeowww: Remember to always add citations, references, and adding contexts for these special things like magic numbers, or feature flags, this helps people to find the answer or understand why we wrote it in the very first place.

PR #780, opened 2025-11-30
feat(pinia-broadcast): use BroadcastChannel & SharedWorker for syncing states across windows & tabs

nekomeowww:
## Description

- Currently, `stage-tamagotchi` is multi-window / multi-tab architecture, means Chat window / Inlay window doesn't share the same Pinia with main window (which model renders at), this causes issue when syncing states between Pinia stores across windows.
- The situation becomes more complex when it comes to play TTS audios if we use BroadcastChannel to send the syncing state patches, if no `SharedWorker` or election mechanism like Kubernetes involves, multiple windows / tabs with TTS configured, both chat streaming, TTS will play in the same time across these windows / tabs.
- Here we introduce a `BroadcastChannel` & `SharedWorker` channel implemented for syncing states for windows / tabs first, then we should have another Pull Request, introduce another plugin or option to distinguish those Pinia stores with only one instance live, and use `SharedWorker` to dispatch messages only for the first constructor caller.


0xSelenicDove:
Closing this PR because its global state-sync design no longer matches the current architecture.

Since this PR opened, `main` added scoped cross-window synchronization for chat, speech, model settings, rendering, and other domains. These paths now use dedicated Eventa or `BroadcastChannel` contracts with explicit ownership.

The proposed `SharedWorker` coordinator also conflicts with current platform requirements. AIRI supports Chromium-based Android environments where `SharedWorker` is unavailable.

Rebasing this branch would require an architectural redesign. If a specific state-sync problem remains, please open a focused PR in the module that owns that state.

Thank you for the original design work and investigation.

Issue #255, opened 2025-07-02
proposal: Plugin system & Plugin API


nekomeowww:
### Clear and concise description of the problem

I would love to add a new proposal for everyone (including official maintainers like us) to be able to add new things into Project AIRI easily, to build up better and widely benefitial community, the **Plugin APi**.

### Background

Applications in VTuber field like [OBS](https://obsproject.org), Warudo and [VTubeStudio](https://github.com/DenchiSoft/VTubeStudio/wiki/Plugins) provides exclusive API system for users to integrate their own modules and scripts inside the system. [Obsidian](https://docs.obsidian.md/Plugins/Getting+started/Build+a+plugin), [VSCode](https://code.visualstudio.com/api/get-started/your-first-extension), so do libraries like [code-mirror](https://codemirror.net/docs/extensions/), [Vue](https://vuejs.org), and [Tauri](https://tauri.app). I recently read the blog post from [Figma](https://www.figma.com/plugin-docs/), they got their own plugin system too. Games like [Minecraft](https://minecraft.fandom.com/wiki/Tutorials/Creating_Forge_mods ), and [Factorio](https://wiki.factorio.com/Modding) which AIRI original supported got their own Modding capabilities too. 

The Project AIRI has gone far quite a while, the internal system became a bit messy. Even for easy things like adding and integrating new LLM/AI provider, right now it requires us to maintaining different of modules where we would need to modify many files in order to get things done, the process became much more complicated for Minecraft, Discord, Telegram, and Factorio agent too, they are not easy to write and maintain within both the scope of Project AIRI and the agents within our ecosystem. 

Another motivation of having Plugin system is that, new contributors, users, may not need to spend days or weeks to understand our repository with TypeScript and Vue knowledge, but easy to add new building blocks or actual stuff into the system.

I know some could argue that this is due to the design flaw of our architecture. But I have to say, I was progressively learning, and constantly improving this project with faster iteration speed and doing all different things already...

### What is Plugin exactly? 

In detail, the concept of Plugin in [Warudo](https://warudo) was divided into [Blueprint](https://docs.warudo.app/docs/blueprints/understanding-blueprints), [Mod](https://docs.warudo.app/docs/modding/mod-sdk), and [Scripting](https://docs.warudo.app/docs/scripting/overview). OBS got [Plugins](https://docs.obsproject.com/scripting) and [Scripts](https://obsproject.com/kb/scripting-guide) divided too. 

[...]

PR #800, opened 2025-12-16
feat: add serverless memory API functions and memory system package

inoribea:
### Summary
This PR adds serverless memory API functions and memory system package to provide both short-term and long-term memory capabilities for AIRI.

### Details
- Implements serverless functions for memory operations (save, search, clear)
- Provides support for multiple memory providers (Vercel KV, Upstash Redis, PostgreSQL with pgvector, Qdrant)
- Adds memory embedding functionality with support for OpenAI and Cloudflare embeddings
- Creates a unified memory system package with provider abstraction
- Includes migration scripts for database setup

### Testing
1. Verify that memory save API endpoint works correctly
2. Test memory search functionality with different providers
3. Confirm short-term memory operations (Vercel KV, Upstash Redis)
4. Verify long-term memory operations (PostgreSQL, Qdrant)
5. Test embedding functionality with various providers

### DIFF
--- a/api/_lib/memory.ts
+++ b/api/_lib/memory.ts
@@ -12,7 +12,13 @@ import { QdrantClient } from '@qdrant/js-client-rest'
 import { Redis } from '@upstash/redis'
 import { kv as vercelKv } from '@vercel/kv'
 import { sql } from '@vercel/postgres'
+import { embed } from '@xsai/embed'
 import { Pool } from 'pg'
+
+// Constants
+const DEFAULT_OPENAI_EMBEDDING_MODEL = 'text-embedding-3-small'
+const DEFAULT_CLOUDFLARE_EMBEDDING_MODEL = '@cf/baai/bge-base-en-v1.5'
+const DEFAULT_CLOUDFLARE_API_BASE_URL = 'https://api.cloudflare.com/client/v4'
 
 interface SqlQueryExecutor {
   query: <T extends QueryResultRow = QueryResultRow>(text: string, params?: unknown[]) => Promise<QueryResult<T>>
@@ -298,33 +304,24 @@ async function generateEmbedding(text: string): Promise<number[]> {
     return response.data[0].embedding
   }
   else if (provider === 'cloudflare') {
-    const { cloudflare } = config.longTerm.embedding
-    if (!cloudflare) {
+    const cfConfig = embeddingConfig.cloudflare
+    if (!cfConfig) {
       throw new Error('Cloudflare configuration is missing')
     }
 
-    const response = await fetch(
-      `https://api.cloudflare.com/client/v4/accounts/${cloudflare.accountId}/ai/run/${cloudflare.model}`,
-      {
-        method: 'POST',
-        headers: {
-          'Authorization': `Bearer ${cloudflare.apiToken}`,
-          'Content-Type': 'application/json',
-        },
-        body: JSON.stringify({ text }),
-      },
-    )
+    // Use Cloudflare Workers AI OpenAI-compatible endpoint
+    // Format: https://api.cloudflare.com/client/v4/accounts/{account_id}/ai/v1
+    const baseURL = `${DEFAULT_CLOUDFLARE_API_BASE_URL}/accounts/${cfConfig.accountId}/ai/v1`
+    const embeddingModel = cfConfig.model || DEFAULT_CLOUDFLARE_EMBEDDING_MODEL
 
-    if (!response.ok) {
-      throw new Error(`Cloudflare API error: ${response.statusText}`)
-    }
+    const { embedding } = await embed({
+      apiKey: cfConfig.apiToken,
+      baseURL,
+      input: text,
+      model: embeddingModel,
+    })
 
-    const payload = await response.json() as Record<string, any>
-    console.info('[Memory Debug] Cloudflare API response structure:', JSON.stringify(payload, null, 2).substring(0, 500))
-    const embedding = extractCloudflareEmbedding(payload)
-
-    if (!embedding || !Array.isArray(embedding) || embedding.length === 0) {
-      console.error('[Memory Debug] Failed to extract embedding. Full payload:', JSON.stringify(payload))
+    if (!embedding || !embedding.length) {
       throw new Error('Cloudflare embedding response did not include an embedding vector.')
     }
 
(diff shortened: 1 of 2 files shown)


shinohara-rin:
As a heads-up there're plans to implement the memory system as plugins, please refer to #255 and #520

api/_lib/memory.ts
nekomeowww, line 1: I think it'd be better if we could provide such API in both Electron and remote API way?

api/_lib/memory.ts
```diff
@@ -0,0 +1,748 @@
+// Standalone memory service for Vercel serverless functions
+// Supports Vercel KV and Upstash Redis for short-term memory
+// Supports Postgres+pgvector and Qdrant for long-term memory
+
+import type { QueryResult, QueryResultRow } from 'pg'
+
+import { randomUUID } from 'node:crypto'
+
+import OpenAI from 'openai'
```
nekomeowww, line 9: Should switch to xsai.js.org

api/_lib/memory.ts
```diff
@@ -0,0 +1,748 @@
...
+
+import { QdrantClient } from '@qdrant/js-client-rest'
+import { Redis } from '@upstash/redis'
+import { kv as vercelKv } from '@vercel/kv'
+import { sql } from '@vercel/postgres'
+import { Pool } from 'pg'
```
nekomeowww, line 16: I think we should have something like `storage-backends` or abstractions for these, to support different storages. Do LlamaIndex provide enough design reference for this? At least they are company and hopefully their package will last one enough.

api/_migrations/001_create_memory_embeddings.sql
```diff
@@ -0,0 +1,35 @@
...
+    user_id TEXT NOT NULL,
+    content TEXT NOT NULL,
+    embedding vector(768),  -- Adjust dimension based on your embedding model
+    metadata JSONB,
+    created_at TIMESTAMPTZ DEFAULT NOW(),
+    
+    -- Indexes for performance
+    INDEX idx_memory_user_id (user_id),
+    INDEX idx_memory_created_at (created_at DESC)
+);
```
nekomeowww, line 19: The design here isn't the optimal. But we can go first with this.

api/_migrations/README.md
nekomeowww, line 1: Remove this file.

api/memory/clear.ts
```diff
@@ -0,0 +1,37 @@
...
+    return res.status(200).json({ success: true })
+  }
+  catch (error) {
+    console.error('Error in /api/memory/clear:', error)
+    return res.status(500).json({
+      success: false,
+      error: error instanceof Error ? error.message : String(error),
+    })
+  }
+}
```
nekomeowww, line 37: If so, why don't we use Nitro or Hono?

api/memory/config.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
+  // Set CORS headers
+  res.setHeader('Access-Control-Allow-Origin', '*')
+  res.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS')
+  res.setHeader('Access-Control-Allow-Headers', 'Content-Type')
+
+  if (req.method === 'OPTIONS') {
+    return res.status(200).end()
+  }
```
nekomeowww, line 13: It seems CORS is occurring in multiple files, middleware should be used.

api/memory/search.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  try {
+    const body = req.body as { query?: string, userId?: string, limit?: number }
+
+    console.info('[Memory Debug Search] 收到搜索请求:', {
+      query: body?.query,
+      userId: body?.userId,
+      limit: body?.limit,
+    })
```
nekomeowww, line 26: Debug console? Should be removed?

api/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@proj-airi/api",
+  "version": "0.7.2-beta.3",
+  "private": true,
+  "description": "Vercel serverless API functions for AIRI",
```
nekomeowww, line 5: So it works only on Vercel?

packages/memory/src/providers/long-term/postgres-pgvector.provider.ts
```diff
@@ -0,0 +1,383 @@
...
+    await this.pool.query(`CREATE INDEX IF NOT EXISTS ${this.tableName}_user_id_idx ON ${this.tableName} (user_id);`)
+    await this.pool.query(`CREATE INDEX IF NOT EXISTS ${this.tableName}_session_id_idx ON ${this.tableName} (session_id);`)
+    await this.pool.query(`CREATE INDEX IF NOT EXISTS ${this.tableName}_created_at_idx ON ${this.tableName} (created_at DESC);`)
+    await this.pool.query(
+      `CREATE INDEX IF NOT EXISTS ${this.tableName}_embedding_idx
+       ON ${this.tableName}
+       USING ivfflat (embedding vector_l2_ops)
+       WITH (lists = 100);`,
+    )
+  }
```
nekomeowww, line 185: Why do we need to take care of this? Shouldn't drizzle helped us?

packages/memory/src/providers/long-term/qdrant.provider.ts
```diff
@@ -0,0 +1,348 @@
...
+    if (!embedding.length) {
+      throw new Error('Failed to generate embedding for memory content.')
+    }
+
+    if (embedding.length !== this.vectorSize) {
+      throw new Error(`Embedding dimension mismatch. Expected ${this.vectorSize}, received ${embedding.length}.`)
+    }
+
+    return embedding.map((value: number | string) => Number(value))
+  }
```
nekomeowww, line 240: Try switching to https://xsai.js.org/docs/packages/embed

packages/memory/src/providers/long-term/qdrant.provider.ts
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    return {
+      provider,
+      apiKey,
+      model,
+      baseUrl,
+      accountId,
+    } satisfies EmbeddingProviderConfiguration
+  }
```
nekomeowww, line 293: This should be done before application start?

packages/memory/src/providers/long-term/qdrant.provider.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    const payload = await response.json() as Record<string, any>
+    const embedding = this.extractCloudflareEmbedding(payload)
+
+    if (!embedding || !Array.isArray(embedding) || embedding.length === 0) {
+      throw new Error('Cloudflare embedding response did not include an embedding vector.')
+    }
+
+    return embedding.map((value: number | string) => Number(value))
+  }
```
nekomeowww, line 322: https://xsai.js.org/docs/packages-ext/providers got Cloudflare supported well and enough.

api/_lib/memory.ts
inoribea: I've switched to xsai, which can be used directly in Electron. The api/ directory is designed as a supplementary solution for remote API scenarios. Users can choose either local or remote approach based on their needs.

api/_lib/memory.ts
inoribea: We already have `IMemoryProvider` as the storage abstraction layer. I'm more familiar with LangChain. do you have specific design suggestions?

api/memory/clear.ts
inoribea: The current code has too much CORS boilerplate. I'll migrate to Hono in a follow-up PR - it's a small change since. business logic is already separated in _lib/memory.ts.

api/package.json
inoribea: Yes, it's Vercel-oriented. Old habit from my serverless days. Now that we're using xsai, local deployment should be easier to support. Will work on it in a follow-up PR.

packages/memory/src/providers/long-term/postgres-pgvector.provider.ts
inoribea: Not using Drizzle elsewhere in the project yet. Current focus is getting full functionality working in serverless first. Will consider adding Drizzle later for better maintainability.

packages/memory/src/providers/long-term/qdrant.provider.ts
inoribea: Switched.

packages/memory/src/providers/long-term/qdrant.provider.ts
inoribea: For serverless, runtime check feels safer since cold starts can happen anytime. But yeah, maybe could add a separate setup script for traditional deployments.

packages/memory/src/providers/long-term/qdrant.provider.ts
inoribea: Updated to use xsai now. The Cloudflare implementation now uses xsai's embed function with Cloudflare Workers AI's OpenAI-compatible API endpoint (/accounts/{accountId}/ai/v1), instead of manual fetch calls.

nekomeowww:
New designed finalized for memory layer, in both #879 and https://github.com/moeru-ai/plast-mem. We will support adding additional memory layer to the AIRI ecosystem in runtime level and you will be able to implement your own memory layer on top of it, instead of adding into the system directly. So closing now.

PR #801, opened 2025-12-16
feat: add memory system frontend components and settings UI

inoribea:
### Summary
This PR adds memory system frontend components and settings UI to allow users to configure and manage memory settings in the AIRI application.

### Details
- Adds memory store implementation for frontend state management
- Creates memory settings page with UI components for configuration
- Implements memory provider selection and configuration
- Adds UI components for short-term and long-term memory configuration
- Provides configuration options for embedding providers

### Testing
1. Verify that memory settings page renders correctly
2. Test memory provider selection functionality
3. Confirm embedding configuration options work properly
4. Verify short-term and long-term memory configuration UI
5. Ensure all memory-related settings are properly saved and applied


nekomeowww:
New designed finalized for memory layer, in both #879 and https://github.com/moeru-ai/plast-mem. We will support adding additional memory layer to the AIRI ecosystem in runtime level and you will be able to implement your own memory layer on top of it, instead of adding into the system directly. So closing now.

PR #805, opened 2025-12-16
feat: Implement a Memory Journal system and enhance performance

ineedcakeahhhhhhh:
feat: Implement a Memory Journal system and enhance performance

 ## Memory Journal: Include useMemoryStore for search/delete journal items.

 Add the following five LLM tools: write_journal, read_journal, delete_journal, check_journal_status, and journal_stats

 Include JournalModal.vue with animations, search, and remove.

 - Full i18n support (9 languages) - Smart detection: Personal vs. Technical/Dev Log entries

 ## Enhancements for Performance

 Include composable GPU detection (WebGPU/WebGL layer)

 Include a performance monitor (memory, draw calls, FPS).

 Enhance ThreeScene through adaptive DPR

 Improve chunking and terser to optimize the Vite build.

### DIFF
--- a/packages/stage-ui/src/stores/memory.ts
+++ b/packages/stage-ui/src/stores/memory.ts
@@ -1,10 +1,11 @@
+import { generateObject } from '@xsai/generate-object'
 import { type ChatProvider } from '@xsai-ext/shared-providers'
 import { useLocalStorage } from '@vueuse/core'
 import { defineStore, storeToRefs } from 'pinia'
 import { computed } from 'vue'
+import * as v from 'valibot'
 
 import { useChatStore } from './chat'
-import { useLLM } from './llm'
 import { useConsciousnessStore } from './modules/consciousness'
 import { useProvidersStore } from './providers'
 
@@ -25,12 +26,16 @@ export interface JournalEntry {
   tags?: string[]
 }
 
-interface LLMJournalResponse {
-  content: string
-  mood: JournalMood
-  type: JournalType
-  tags?: string[]
-}
+// ─────────────────────────────────────────────────────────────────────────────
+// Schema (Valibot for validation)
+// ─────────────────────────────────────────────────────────────────────────────
+
+const JournalResponseSchema = v.object({
+  content: v.string(),
+  mood: v.picklist(['happy', 'sad', 'excited', 'tired', 'calm', 'curious', 'annoyed', 'focus', 'neutral']),
+  type: v.picklist(['personal', 'technical']),
+  tags: v.optional(v.array(v.string())),
+})
 
 // ─────────────────────────────────────────────────────────────────────────────
 // Constants
@@ -60,7 +65,6 @@ export const useMemoryStore = defineStore('memory', () => {
   
   // Dependencies
   const { messages } = storeToRefs(useChatStore())
-  const { stream } = useLLM()
   const { activeProvider, activeModel } = storeToRefs(useConsciousnessStore())
   const providersStore = useProvidersStore()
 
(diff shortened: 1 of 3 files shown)


apps/stage-web/vite.config.ts
```diff
@@ -69,6 +69,39 @@ export default defineConfig({
...
+            if (modules.some(module => id.includes(module))) {
+              return chunk
+            }
+          }
+        },
+      },
+      treeshake: {
+        moduleSideEffects: 'no-external',
+      },
+    },
```
nekomeowww, line 77: we should not configure this...

packages/stage-ui-three/src/composables/useGPUDetect.ts
nekomeowww, line 1: Are there any existing library we could borrow for this?

nekomeowww:
Conflicted, need to rebase.

apps/stage-web/vite.config.ts
```diff
@@ -69,6 +69,22 @@ export default defineConfig({
...
+          }
+          if (id.includes('three') || id.includes('@tresjs')) {
+            return 'three-vendor'
+          }
+          if (id.includes('drizzle') || id.includes('localforage')) {
+            return 'db-vendor'
+          }
+        },
+      },
+    },
```
nekomeowww, line 77: Again, your agent should follow to now to write things like this.

apps/stage-web/src/components/Journal/JournalModal.vue
```diff
@@ -0,0 +1,323 @@
+<script setup lang="ts">
+import { useMemoryStore } from '@proj-airi/stage-ui/stores/memory'
+import type { JournalEntry, JournalMood } from '@proj-airi/stage-ui/stores/memory'
+import { storeToRefs } from 'pinia'
+import { computed, ref, watch } from 'vue'
+import { useI18n } from 'vue-i18n'
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Props & Emits
+// ─────────────────────────────────────────────────────────────────────────────
```
nekomeowww: Can we not to have this kind of markers?

packages/stage-ui/src/stores/memory.ts
```diff
@@ -0,0 +1,289 @@
...
+CONVERSATION:
+${conversation}
+
+RESPOND WITH ONLY A VALID JSON OBJECT:
+{
+  "content": "Your journal entry here (markdown supported)",
+  "mood": "one of the mood options",
+  "type": "personal" or "technical",
+  "tags": ["tag1", "tag2"]
+}`
```
nekomeowww: Should use Valibot + `@xsai/generate-object` now.

packages/stage-ui-three/src/composables/useGPUDetect.ts
nekomeowww, line 1: Why do we need GPU related things?

packages/stage-ui-three/src/components/ThreeScene.vue
```diff
@@ -88,6 +88,10 @@ const {
   skyBoxIntensity,
 } = storeToRefs(modelStore)
 
+
+const optimizedDPR = computed(() => optimalSettings.value.pixelRatio)
+const enablePostProcessing = computed(() => optimalSettings.value.postProcessing)
```
nekomeowww: In most cases this should not hurt the performance, what is the reason of this? 

If what you are trying to do is to prevent the 3D scene to not to lag once when loaded the first time, then this doesn't solve this problem completely.

nekomeowww:
All build failed.

packages/stage-ui-three/src/composables/useGPUDetect.ts
nekomeowww: So do we have any answers? This is not solved, do not click the resolve conversation button.

PR #808, opened 2025-12-16
fix(stage-ui): AIRI in full resolution! fixed pixi.js resolution issue

shinohara-rin:
No more blurry live2d models on HDPI screens!

image
image

Yes it also works on tamagotchi

Screenshot 2025-12-17 at 10 12 21 AM


shinohara-rin:
However... this also introduces more render load, and it kinda lags on low-power devices. It's a good idea to make this toggleable in the settings.

nekomeowww (approved):
Oh I didn't try this option composition... I tried with many other combos...

shinohara-rin:
> Oh I didn't try this option composition... I tried with many other combos...

This is also 4 hours of consecutive retrying :)

PR #819, opened 2025-12-19
fix(stage-ui): prevent race conditions in model loading

shinohara-rin:
This PR fixes the funny render issue when loading certain models.

https://github.com/user-attachments/assets/cfca55d6-9801-4bef-9910-1b696c0d9752


packages/stage-ui/src/components/scenes/live2d/Model.vue
```diff
@@ -76,6 +76,8 @@ const modelLoading = ref(false)
 // NOTICE: boolean is sufficient; this flag is only used inside loadModel to bail out if the component unmounts mid-load.
 let isUnmounted = false
 
+const currentLoadRequestId = ref(0)
```
nekomeowww: Try to use https://es-toolkit.dev/reference/promise/Mutex.html.

We have it installed by default, most of the utility calls are based on this, you can ask the agent to refer to Context7  with https://context7.com/toss/es-toolkit

shinohara-rin: can't find it in `package json` though... is there something I've missed?

nekomeowww: Sorry I just added it.

PR #826, opened 2025-12-22
(feat/integrated-memory) Simple SQLite Attachment for Short-Term, Long-Term memory as an alternative of browser-cached memory

gg582:
## Description
Recycling drafts from mainline, and attach SQLite extension to export

## Linked Issues
Currently, draft


gg582:
This is yet rough, though it is actively trying integration....Especially, for memory import/export.

PR #827, opened 2025-12-22
fix(chat): serialize sends and reset on cleanup

s3d-i:
## Description

- Adds a per-chat FIFO send queue and generation guard so rapid consecutive sends are serialized instead of interleaving on the shared streamingMessage. Each send runs end-to-end before the next begins; stale callbacks are dropped if the session was cleared.
- Updates cleanupMessages to increment generation, clear pending sends, reset sending, and zero streamingMessage, preventing ghost placeholders or late tokens from repopulating a cleared history.
- Leaves UI unchanged; Enter spam is handled transparently by the queue.

Rationale: previously, multiple send calls in one window could overlap and mix responses, and trashing mid-stream left sending true and allowed placeholder to remain. The queue + generation reset makes the send pipeline single-flight per store and makes cleanup authoritative.

## Additional Context

Another issue is duplicated messages in stage-tamagotchi. Related to https://github.com/moeru-ai/airi/pull/780


s3d-i:
added optimistic textarea cleanup feat; optimised v-for rendering to prevent flickering   
doubt whether I should split it into different PRs

Issue #829, opened 2025-12-24
Nix devshell does not contain python


ReStranger:
### Describe the bug

Missing Python causes an error during postinstallation

```bash
[restranger@pc:/mnt/sda1/Develop/airi]$ pnpm i
Scope: all 37 workspace projects
Lockfile is up to date, resolution step is skipped
Already up to date
. postinstall$ npx simple-git-hooks && pnpm run build:packages
[269 lines collapsed]
│ @proj-airi/font-cjkfonts-allseto:build: ℹ config file: /mnt/sda1/Develop/airi/packages/font-cjkfonts-allseto/tsd…
│ @proj-airi/font-cjkfonts-allseto:build: ℹ entry: src/index.ts
│ @proj-airi/font-cjkfonts-allseto:build: ℹ tsconfig: tsconfig.json
│ @proj-airi/font-cjkfonts-allseto:build: ℹ Build start
│ @proj-airi/font-cjkfonts-allseto:build: ℹ dist/index.mjs  0.04 kB │ gzip: 0.05 kB
│ @proj-airi/font-cjkfonts-allseto:build: ℹ 1 files, total: 0.04 kB
│ @proj-airi/font-cjkfonts-allseto:build: ✔ Build complete in 29ms
│  Tasks:    13 successful, 13 total
│ Cached:    13 cached, 13 total
│   Time:    881ms >>> FULL TURBO
└─ Done in 3.2s
plugins/airi-plugin-web-extension postinstall$ wxt prepare
│ WXT 0.20.13
│ ℹ Generating types...
│ ✔ Finished in 4.736 s
└─ Done in 9.9s
plugins/airi-plugin-vscode prepare$ pnpm run update
│ > airi-vscode@0.8.0-beta.6 update /mnt/sda1/Develop/airi/plugins/airi-plugin-vscode
│ > tsx ./scripts/vscode-ext-gen.ts
└─ Done in 735ms
apps/stage-tamagotchi postinstall$ electron-builder install-app-deps
│   • electron-builder  version=26.4.0
│   • loaded configuration  file=/mnt/sda1/Develop/airi/apps/stage-tamagotchi/electron-builder.yml
│   • packageManager not detected by file, falling back to environment detection  resolvedPackageManager=pnpm dete…
│   • detected workspace root for project using packageManager field  pm=pnpm config=pnpm@10.25.0 resolved=/mnt/sd…
[...]

PR #830, opened 2025-12-24
feat(nix): Add python3 in devShell

ReStranger:
## Description

Just adding python3 to devShell

## Linked Issues

#829 

## Additional Context

You can set a specific python version (like python313/python314) for reliability

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -42,7 +42,7 @@
               nixfmt-rfc-style
               nixfmt-tree
               pnpm
-              python3
+              python314
             ];
           };
         }


flake.nix
```diff
@@ -42,6 +42,7 @@
               nixfmt-rfc-style
               nixfmt-tree
               pnpm
+              python3
```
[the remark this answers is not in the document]
ReStranger: I would do this, but I need to decide which version to take as a default

nekomeowww:
```suggestion
              python314
```

nekomeowww: How about 314?

nekomeowww: /gemini review

Can I write here `python314`?

ReStranger: Okay, I'll pin 3.14

ReStranger: Okay, I'll pin 3.14

Issue #817, opened 2025-12-18
Provider validation: `health` and `chat_completions` checks appear redundant?


shinohara-rin:
The two validations `health` and `chat_completions` seem to be doing the very same check
https://github.com/moeru-ai/airi/blob/55f3019737982969fe75272c8ee8e597e67f918a/packages/stage-ui/src/stores/providers/openai-compatible-builder.ts#L189-L205

https://github.com/moeru-ai/airi/blob/55f3019737982969fe75272c8ee8e597e67f918a/packages/stage-ui/src/stores/providers/openai-compatible-builder.ts#L230-L246

Is chat_completions intended to test something different from health?

Currently, they're used from `novita-ai` and `modelscope`, but `health` seems to be sufficient already?
https://github.com/moeru-ai/airi/blob/55f3019737982969fe75272c8ee8e597e67f918a/packages/stage-ui/src/stores/providers.ts#L1586-L1597

https://github.com/moeru-ai/airi/blob/55f3019737982969fe75272c8ee8e597e67f918a/packages/stage-ui/src/stores/providers.ts#L1697-L1707

Iro96:
lemme check

Iro96:
i think you're right, it just a code dupe of `health_check` and the `chat_completions` validation did not have a special purpose.

shinohara-rin:
I suppose we'll remove it whenever convenient

nekomeowww:
Done.

PR #831, opened 2025-12-25
feat(stage-ui): remove redundant chat completions validation

Iro96:
## Description

The chat_completions validation check was a duplicate of the health check in the openai-compatible-builder, causing a redundant API call during provider validation.

This affects novita-ai and modelscope providers which were using both.

This change removes the superfluous chat_completions check and updates the provider configurations to rely only on the health check, which already validates the required endpoint.

## Linked Issues

- https://github.com/moeru-ai/airi/issues/817


Iro96:
i think there are some issues in github workflows cc @nekomeowww

shinohara-rin:
it's that `pnpm-lock.yaml` again

nekomeowww (approved):

PR #807, opened 2025-12-16
feat(server): with api server, service-lize

luoling8192:

PR #843, opened 2025-12-29
feat(user): Add user login/signup and user profile UI

Iro96:


luoling8192:
Thanks for your contribution! We have already basic Auth system in https://github.com/moeru-ai/airi/issues/807 with better-auth

Please stay tuned, or we can contribute it together!

PR #844, opened 2025-12-29
feat(user): add login/signup and user profile UI

Iro96:


nekomeowww:
We have the login page here in https://github.com/moeru-ai/airi/blob/0.9.0/apps/stage-web/src/pages/auth/login.vue.

If you need to add email & password option, implement this in 0.9.0, reopen after modified.

PR #854, opened 2025-12-30
refactor(stage-ui): fix vad, prompt hygiene, lifecycle hooks

asyadil:
## what was changed
- fix VAD/transcription flow: interrupt streaming on speech start, WAV conversion, transcription pipe, silence stop
- harden chat prompt/context: sanitize context keys, render system prompt placeholders, prevent prototype pollution
- clean lifecycle & types: add disposers for update hooks, extract model-store types/lighting presets, export TrackingMode; fix contributors avatars fallback
- ensure lint/typecheck pass across workspaces

_______________________
thx!
HAPPY HOLIDAY ....!!!!

### DIFF
--- a/packages/stage-ui/src/components/scenarios/settings/model-settings/VRM.vue
+++ b/packages/stage-ui/src/components/scenarios/settings/model-settings/VRM.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import type { TrackingMode } from '@proj-airi/stage-ui-three'
+
 import { useModelStore } from '@proj-airi/stage-ui-three'
 import { Button } from '@proj-airi/ui'
 import { storeToRefs } from 'pinia'
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2621,7 +2621,7 @@ importers:
         version: 66.5.11
       '@wxt-dev/module-vue':
         specifier: ^1.0.3
-        version: 1.0.3(vite@8.0.0-beta.5(@types/node@24.10.4)(esbuild@0.25.12)(jiti@2.6.1)(less@4.5.1)(terser@5.44.1)(tsx@4.21.0)(yaml@2.8.2))(vue@3.5.25(typescript@5.9.3))(wxt@0.20.13(@types/node@24.10.4)(jiti@2.6.1)(less@4.5.1)(lightningcss@1.30.2)(rollup@4.54.0)(terser@5.44.1)(tsx@4.21.0)(yaml@2.8.2))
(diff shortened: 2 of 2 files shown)


apps/stage-tamagotchi/src/renderer/components/InteractiveArea.vue
```diff
@@ -28,6 +29,58 @@ const { t } = useI18n()
...
+
+  let offset = WAV_HEADER_SIZE
+  for (let i = 0; i < buffer.length; i++) {
+    const s = Math.max(-1, Math.min(1, buffer[i]))
+    view.setInt16(offset, s < 0 ? s * 0x8000 : s * 0x7FFF, true)
+    offset += BYTES_PER_SAMPLE
+  }
+
+  return new File([arrayBuffer], 'recording.wav', { type: 'audio/wav' })
+}
```
nekomeowww, line 83: No longer needed.

apps/stage-tamagotchi/src/renderer/components/InteractiveArea.vue
```diff
@@ -102,33 +155,34 @@ function removeAttachment(index: number) {
...
+  try {
+    const recordingFile = float32ToWavFile(buffer)
+    const text = await transcribeForRecording(recordingFile)
+    if (typeof text === 'string' && text.trim())
+      messageInput.value = text
+  }
+  catch (err) {
+    console.error('Transcription failed', err)
+  }
 }
```
nekomeowww, line 186: No longer needed.

docs/.vitepress/contributors.ts
```diff
@@ -44,8 +48,51 @@ function createLinks(tm: CoreTeam): CoreTeam {
...
+    github: 'junkwarrior87',
+    title: 'Engineer',
+  },
+  {
+    name: 'Menci',
+    avatar: resolveAvatar('Menci'),
+    github: 'Menci',
+    title: 'Engineer',
+  },
+]
```
nekomeowww, line 95: Uh... we don't use this field for now, perhaps undo this change?

nekomeowww:
I think many changes in this PR should not be done like this without discussion and context sync. 

In fact we have unified VAD & ASR in https://github.com/moeru-ai/airi/blob/95d7612fedbe51662c0c4ad871c43d17cc19b713/apps/stage-tamagotchi/src/renderer/pages/index.vue#L119-L132. It's just... not applied to chat window. If you need, you can up level this to components, and make it a AudioInputProvider or something, and so that both stage.vue input and chat input could benefit from VAD & ASR.

asyadil:
Sorry for the scope creep. I’ll drop the VAD/InteractiveArea and plainTeamMembers changes; will resubmit a smaller PR for the hygiene bits if that’s acceptable.

Issue #835, opened 2025-12-27
Feature(Consciousness Model Compatibility)


TECozens:
### Clear and concise description of the problem

I think that the chat / consciousness module could benefit greatly from some added compatibility for models which include response structures that have `<think> </think>` related tags (at least for common flavours of models). It would be great if we could filter these responses (make them collapse-able) or store them for use in memories and avoid it in TTS responses where it is not needed.

### Suggested solution / Ideas

From my playground exercises with Ollama models I often found that models with this response structure leak into the TTS which is not Ideal... we don't want to hear our Waifu's thoughts but rather keep the conversational speech or at least have the option to be selective.

### Alternative

We could go as far as re-using the extracted `<think> / <thought>` from response for memories.

### Additional context

In the chat / consciousness Module we experience TTS for parts of a response which are not speech.

Don't want (TTS reads the whole thing):
`<think>` maybe I should post this by saying I'm posting this `</think>` "I'm posting this" 

What we want:
"I'm posting this"

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Iro96:
Thank you for your idea, it's great, and through these tags in model responses we can categorize the model's response content, which is very useful for memory functions for contexts (need more dicussion) and also significantly improved the model's output. Feel free if you can help

shinohara-rin:
*nods*
In addition to this, I think non-verbal expressions wrapped in `*` should also be skipped for TTS.
*tilts my head*

TECozens:
> Thank you for your idea, it's great, and through these tags in model responses we can categorize the model's response content, which is very useful for memory functions for contexts (need more dicussion) and also significantly improved the model's output. Feel free if you can help

@Iro96  I'm interested in exploring this further, my help may be limited to start with... as of when I'll update here as I get insights... think I'll start with looking at model responses, possible categories to deconstruct the problem / solution. (gives me the opportunity to learn more about these models too 😸)

In my local fork I have messy branch  version I put together I'll try to come up with a clean initial extraction method for response categorisation then use that categorisation to add a response-category filter to TTS (an allowed list of responses from the model for TTS).

[...]

PR #859, opened 2026-01-01
feat(stage-ui): process for reasoning models

TECozens:
### Before
image

### After
image

## New Features
This PR will provide a collapsible reasoning box to the assistant chat, it renders only the speech in the chat
TTS ignores reasoning

Reasoning is defined by any response which is wrapped in tags `<think>` , `<thoughts>` etc, this excludes and does not have an affect on SPECIAL_TOKENS

Provided is `use-chat-session/`  `summary.ts` containing helper functions to grab the reasoning from a chat session, I believe we can use this to create summaries and store those are memories or use the reasoning directly with little processing.

## Linked Issues

#835 

## Additional Context

There were a lot of challenges in retrospect so I have kept a couple of the ideas in mind for future feature requests and PRs

I think the biggest challenge is determining if the input comes from a reasoning model and what parts of text were reasoning.

I think that if a toggle is needed I could add one to enable/disable reasoning filter, but at present it moves reasoning content if found to a collapsible box and TTS ignores it during stream.

Still i found the streaming and tts to be quite difficult to understand at first like if the response was completed and we performed TTS on the final response that might be easier than doing it per chunk we parse as we do.

Open to suggestions but I found limitations in trying to work around the current approach, if all if fine then we can push this!

I also think that this burns tokens like crazy  (Current Release) Significantly less now with only speech parts
image

## Testing
Unit Test Response Categoriser

Example Test Input

```
    // Special tokens like <|EMOTE_HAPPY|> and <|DELAY:1|> should be included in reasoning
    const text = 'Hello <reasoning>thinking <|EMOTE_HAPPY|> about this <|DELAY:1|></reasoning> world!'
```

```
# Run only the response-categoriser tests
pnpm exec vitest run packages/stage-ui/src/composables/response-categoriser.test.ts

# Check the console logs for more verbose output
pnpm exec vitest watch packages/stage-ui/src/composables/response-categoriser.test.ts --reporter=verbos   
```
Sample Output:

```
 Test: should handle <think> with special tokens like <|EMOTE_HAPPY|>
  Input text: Hello <think>thinking <|EMOTE_HAPPY|> about this <|DELAY:1|> and that</think> world!
  Segments found: 1
  Tag name: think
  Segment content: thinking <|EMOTE_HAPPY|> about this <|DELAY:1|> and that
  Has <|EMOTE_HAPPY|>: true
  Has <|DELAY:1|>: true
  Reasoning: thinking <|EMOTE_HAPPY|> about this <|DELAY:1|> and that
  Speech: Hello world!
```

### DIFF
--- a/nix/pnpm-deps-hash.txt
+++ b/nix/pnpm-deps-hash.txt
@@ -1,1 +1,1 @@
-sha256-3rW39/L+VHWslw+bAPS/bMSkYNepxYMhQbds3kbA5uo=
+sha256-+bVZrOtAwghcsxq9oVMjNvaAWbbLTiy5vX94AQPqH20=


packages/stage-ui/src/composables/responseCategorizer.ts
```diff
@@ -0,0 +1,312 @@
...
+    startIndex: number
+    endIndex: number
+  }> = []
+
+  // Match any XML-like tag: <tagname>content</tagname>
+  // This regex finds opening tags, captures the tag name, content, and closing tag
+  const tagPattern = /<([a-z_][\w-]*)>([\s\S]*?)<\/\1>/gi
+
+  let match
+  while ((match = tagPattern.exec(response)) !== null) {
```
nekomeowww: I am wondering, is that possible to use rehype or hast-util-from-string could help this?

TECozens: Added Rehype as it was less hassle to do 👍🏼 and I trust it more in the chat pipeline.

packages/stage-ui/src/composables/responseCategorizer.ts
```diff
@@ -0,0 +1,312 @@
...
+
+    // If stack is non-empty, we have unclosed tags
+    return stack.length > 0
+  }
+
+  return {
+    consume(chunk: string) {
+      buffer += chunk
+      // Re-categorize on each chunk
+      categorized = categorizeResponse(buffer, providerId)
```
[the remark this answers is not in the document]
TECozens: Integrated with rehype solution and added a incremental parser

packages/stage-ui/src/stores/chat.ts
```diff
@@ -391,33 +395,64 @@ export const useChatStore = defineStore('chat', () => {
...
+          // Categorize the full text stream
+          const finalCategorization = categorizeResponse(fullText, activeProvider.value)
+
+          // Always store categorization (even if empty) for consistency and memory features
+          streamingMessage.value.categorization = {
+            speech: finalCategorization.speech,
+            reasoning: finalCategorization.reasoning,
+          }
+        },
         minLiteralEmitLength: 24, // Avoid emitting literals too fast. This is a magic number and can be changed later.
```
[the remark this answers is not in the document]
TECozens: Not  a super huge problem.

packages/stage-ui/src/composables/responseCategorizer.ts
```diff
@@ -0,0 +1,465 @@
...
+          const tree = unified().use(rehypeParse, { fragment: true }).parse(fullText) as Root
+          let closingOffset = -1
+
+          visit(tree, 'element', (node: Element) => {
+            const position = node.position
+            if (position?.end && closingOffset === -1) {
+              const endOffset = getOffsetFromPosition(fullText, position.end)
+              // Check if this element actually has a closing tag in the source
+              const elementSource = fullText.slice(
+                getOffsetFromPosition(fullText, position.start) ?? 0,
```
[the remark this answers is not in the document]
TECozens: Checked locally this works.

TECozens:
Updated 👍🏼 ready for review

packages/stage-ui/src/stores/chat.ts
```diff
@@ -443,14 +478,16 @@ export const useChatStore = defineStore('chat', () => {
...
         if (rawMessage.role === 'assistant') {
-          const { slices: _, tool_results, ...rest } = rawMessage as ChatAssistantMessage
-          return {
+          const { slices: _, tool_results, categorization: __categorization, ...rest } = rawMessage as ChatAssistantMessage
+          // Create a clean, serializable message object
+          return JSON.parse(JSON.stringify({
             ...toRaw(rest),
             tool_results: toRaw(tool_results),
-          }
+          })) as Message
```
nekomeowww: Uh wait, why are we doing this here?

packages/stage-ui/src/composables/response-categoriser.ts
nekomeowww, line 1: Sorry for my legacy naming, update this file to `packages/stage-ui/src/composables/response-categorizer.ts`

packages/stage-ui/src/composables/use-chat-session/summary.ts
nekomeowww, line 1: How about `packages/stage-ui/src/composables/use-chat-session/summary.ts`?

packages/stage-ui/src/components/scenarios/chat/ChatAssistantItem.vue
```diff
@@ -66,6 +67,12 @@ const boxClasses = computed(() => [
         </template>
       </div>
       <div v-else-if="showLoader" i-eos-icons:three-dots-loading />
+
+      <ChatResponseCategorization
```
nekomeowww:
```suggestion
      <ChatResponsePart
```

How... about calling your proposed PR related sub-system `chat response part`? It seems many things could be added into this `Part` component, `reasoning` is one of the possible part.

packages/stage-ui/src/composables/response-categoriser.ts
```diff
@@ -0,0 +1,466 @@
...
+    reasoning,
+    raw: response,
+  }
+}
+
+/**
+ * Note: This receives literal text from useLlmmarkerParser (special tokens <|...|> are already extracted).
+ * Only XML/HTML tags like <think>, <reasoning> need to be parsed here.
+ */
+export function createStreamingCategorizer(
```
nekomeowww, line 214: Do you have test cases for the possible inputs? For example `<think>...<|Special in between|>...</think>`, and probably, may I ask you to create a unit test file for this implementation? You may refer to how `llmmarketParser.test.ts` to understand how unit tests works.

packages/stage-ui/src/components/scenarios/chat/ChatAssistantItem.vue
TECozens: Absolutely!

packages/stage-ui/src/composables/response-categoriser.ts
TECozens: Good point, I'll do that!

packages/stage-ui/src/stores/chat.ts
TECozens: I was skeptical of this for a while, I had a few issues before while figuring my way around things, I understand we don't need this, so I reverted that back to what it was.

packages/stage-ui/src/composables/response-categoriser.ts
TECozens: To run the End to end and Function you can use these.

Example Test Input

```
    // Special tokens like <|EMOTE_HAPPY|> and <|DELAY:1|> should be included in reasoning
    const text = 'Hello <reasoning>thinking <|EMOTE_HAPPY|> about this <|DELAY:1|></reasoning> world!'
```

```
# Run only the response-categoriser tests
pnpm exec vitest run packages/stage-ui/src/composables/response-categoriser.test.ts
```
image

TECozens:
> Generally LGTM.

I think now it should (Hooray, shall be my first open source contribution)

packages/stage-ui/src/components/scenarios/chat/ChatResponsePart.vue
```diff
@@ -0,0 +1,53 @@
...
+  <div v-if="hasReasoning" :class="containerClasses" flex="~ col" gap-1>
+    <Collapsable :default="false">
+      <template #trigger="slotProps">
+        <button
+          class="w-full flex items-center justify-between rounded-lg bg-neutral-100/50 px-2 py-1 text-xs text-neutral-600 outline-none transition-all duration-200 dark:bg-neutral-800/50 hover:bg-neutral-200/50 dark:text-neutral-400 dark:hover:bg-neutral-700/50"
+          @click="slotProps.setVisible(!slotProps.visible)"
+        >
+          <div flex="~ items-center" gap-1.5>
+            <div i-solar:lightbulb-bolt-bold-duotone size-3.5 text-amber-500 dark:text-amber-400 />
+            <span font-medium>Reasoning</span>
```
shinohara-rin: We could use i18n for this

TECozens: > We could use i18n for this

I'll take a look at that ;)

TECozens: @shinohara-rin added it in 👍🏼

PR #861, opened 2026-01-01
chore: hygiene and model-store cleanup

asyadil:
Summary:
- Chat: render system prompt placeholder, sanitize context keys to prevent prototype pollution.
- Lifecycle: onShouldUpdateView now returns disposer; Stage.vue cleans up hooks; live2d hook adjusted.
- LLM: async executor lint fix.
- Hearing: add onActivity/onSilence options to stop streaming on silence/idle.
- Model-store: extracted shared types, per-environment lighting presets, cleaned returns; export TrackingMode; add vite devDep.
- VRM settings: type-only TrackingMode import sorted.

Notes:
- Scoped redo; excludes VAD/InteractiveArea and contributors list per maintainer feedback.

_________________________________________________
- sorry for the earlier scope creep; this PR keeps to the requested scope.


packages/stage-ui/src/stores/chat.ts
```diff
@@ -464,11 +478,9 @@ export const useChatStore = defineStore('chat', () => {
           {
             role: 'user',
             content: [
-            // TODO: use prompt render & i18n system later
-            // TODO: Module should have description & context length management
               { type: 'text', text: ''
-                + 'These are the contextual information retrieved or on-demand updated from other modules, you may use them as context for chat, or reference of the next action, tool call, etc.:\n'
-                + `${Object.entries(activeContexts.value).map(([key, value]) => `Module ${key}: ${JSON.stringify(value)}`).join('\n')}\n` },
+                + 'Context from other modules (sanitized):\n'
```
nekomeowww, line 519:
```suggestion
                + 'These are the contextual information retrieved or on-demand updated from other modules, you may use them as context for chat, or reference of the next action, tool call, etc.:\n'
```

packages/stage-ui-three/src/stores/model-store.ts
```diff
@@ -109,38 +68,36 @@ export const useModelStore = defineStore('modelStore', () => {
...
+
+  // Per-environment directional light presets.
+  const directionalLightSettingsByEnv = useLocalStorage<Record<'hemisphere' | 'skyBox', { intensity: number, color: HexColor }>>(
+    'settings/stage-ui-three/scenes/scene/directional-light/settingsByEnv',
+    {
+      hemisphere: defaultHemisphereLightSettings,
+      skyBox: { intensity: 1.2, color: '#ffffff' },
+    },
+  )
+  const directionalLightIntensity = computed(() => directionalLightSettingsByEnv.value[envSelect.value]?.intensity ?? defaultHemisphereLightSettings.intensity)
```
nekomeowww, line 91:
```suggestion
  // TODO: Manual directional light intensity will not work for other
  //       scenes with different lighting setups. But since the model
  //       is possible to have MeshToonMaterial, and MeshBasicMaterial
  //       without envMap to be able to inherit lighting from HDRI map,
  //       we will have to figure out a way to make this work to apply
  //       different directional light and other lighting setups
  //       for different environments.
  // WORKAROUND: To achieve the rendering style of Warudo for anime style
  //             Genshin Impact, or so called Cartoon style rendering with
  //             harsh shadows and bright highlights.
  // REVIEW: This is a temporary solution, and will be replaced with
  //         a more flexible lighting system in the future.
  const directionalLightIntensity = computed(() => directionalLightSettingsByEnv.value[envSelect.value]?.intensity ?? defaultHemisphereLightSettings.intensity)
```

packages/stage-ui-three/package.json
```diff
@@ -40,6 +40,7 @@
   "devDependencies": {
     "@types/culori": "^4.0.1",
     "@types/three": "^0.182.0",
+    "vite": "^5.0.0",
```
nekomeowww: Why do we have v5 Vite here?

packages/stage-ui/src/stores/llm.ts
```diff
@@ -52,40 +52,42 @@ async function streamFrom(model: string, chatProvider: ChatProvider, messages: M
...
-            reject(err)
-          }
-        },
-      })
-    }
-    catch (err) {
-      reject(err)
-    }
+  return new Promise<void>((resolve, reject) => {
+    void (async () => {
```
nekomeowww:
```suggestion
    void (async () => {
```

Why do we have to use IIFE here?

packages/stage-ui/src/stores/modules/hearing.ts
```diff
@@ -370,7 +397,10 @@ export const useHearingSpeechInputPipeline = defineStore('modules:hearing:speech
       const session = await createAudioStreamFromMediaStream(
         stream,
         options?.sampleRate ?? DEFAULT_SAMPLE_RATE,
-        () => bumpIdle(),
+        {
+          onActivity: () => bumpIdle(),
+          onSilence: () => void stopStreamingTranscription(false, providerId),
+        },
```
nekomeowww, line 402: I don't quite understand this change here, could you please explain why calling `stopStreamingTranscription` onSilence? What is the context of doing this?

nekomeowww:
I am fine with other changes.

packages/stage-ui-three/package.json
```diff
@@ -40,6 +40,7 @@
   "devDependencies": {
     "@types/culori": "^4.0.1",
     "@types/three": "^0.182.0",
+    "vite": "^7.3.0",
```
nekomeowww, line 44:
```suggestion
```

packages/stage-ui-three/src/stores/model-store.ts
```diff
@@ -109,38 +68,36 @@ export const useModelStore = defineStore('modelStore', () => {
...
-
-  const hemisphereSkyColor = useLocalStorage('settings/stage-ui-three/scenes/scene/hemisphere-light/sky-color', '#FFFFFF')
-  const hemisphereGroundColor = useLocalStorage('settings/stage-ui-three/scenes/scene/hemisphere-light/ground-color', '#222222')
+  const defaultHemisphereLightSettings = { intensity: 2.02, color: '#fffbf5' as HexColor }
+
+  const directionalLightPosition = useLocalStorage<Vec3>('settings/stage-ui-three/scenes/scene/directional-light/position', { x: 0, y: 0, z: -1 })
+  const directionalLightTarget = useLocalStorage<Vec3>('settings/stage-ui-three/scenes/scene/directional-light/target', { x: 0, y: 0, z: 0 })
+  const directionalLightRotation = useLocalStorage<Vec3>('settings/stage-ui-three/scenes/scene/directional-light/rotation', { x: 0, y: 0, z: 0 })
+
+  // Per-environment directional light presets.
```
nekomeowww, line 83:
```suggestion
  // TODO: Manual directional light intensity will not work for other
  //       scenes with different lighting setups. But since the model
  //       is possible to have MeshToonMaterial, and MeshBasicMaterial
  //       without envMap to be able to inherit lighting from HDRI map,
  //       we will have to figure out a way to make this work to apply
  //       different directional light and other lighting setups
  //       for different environments.
  // WORKAROUND: To achieve the rendering style of Warudo for anime style
  //             Genshin Impact, or so called Cartoon style rendering with
  //             harsh shadows and bright highlights.
  // REVIEW: This is a temporary solution, and will be replaced with
  //         a more flexible lighting system in the future.
```

packages/stage-ui/src/stores/chat.ts
nekomeowww, line 519:
```suggestion
                + 'These are the contextual information retrieved or on-demand updated from other modules, you may use them as context for chat, or reference of the next action, tool call, etc.:\n'
```

packages/stage-ui/src/stores/llm.ts
```diff
@@ -52,40 +52,44 @@ async function streamFrom(model: string, chatProvider: ChatProvider, messages: M
...
+            }
+          },
+        })
+      }
+      catch (err) {
+        reject(err)
+      }
     }
+
+    run().catch(reject)
```
nekomeowww: I don't quite understand this one, what is the difference of wrapping in another async run than the old one? For caller, they should all receive the same `Promise`, if you need to do this concurrently, you should handle the Promise in the place where it got used to add `.catch(...)`.

packages/stage-ui/src/stores/modules/hearing.ts
nekomeowww: > The VAD worklet only enqueues audio; without this stop, idle sessions linger. We added a silenceTriggered guard so onSilence fires once and resets when activity resumes.

This is not true, we will wait for another 15 seconds to automatically shutdown the graph & API use: https://github.com/moeru-ai/airi/blob/07711006e6a3465e1d007f914f5c78e006319957/packages/stage-ui/src/stores/modules/hearing.ts#L350-L358

If you need quicker strategy, tune the https://github.com/moeru-ai/airi/blob/07711006e6a3465e1d007f914f5c78e006319957/packages/stage-ui/src/stores/modules/hearing.ts#L217 to `5000` or `2500` should be better than calling `stopStreamingTranscription`. Stop means literally stop, should be used only for unmounted tear down: https://github.com/moeru-ai/airi/blob/07711006e6a3465e1d007f914f5c78e006319957/packages/stage-ui/src/stores/modules/hearing.ts#L275-L327

packages/stage-ui-three/package.json
```diff
@@ -41,6 +41,7 @@
...
```
nekomeowww, line 44: Again, this should not be added, workspace root already has Vite installed in https://github.com/moeru-ai/airi/blob/76a9f8c58f06bc0d284c8348366a53f9e4e75899/package.json#L79. What is the reason here you wanted to install this?

packages/stage-ui/src/stores/chat.ts
```diff
@@ -503,11 +515,9 @@ export const useChatStore = defineStore('chat', () => {
           {
             role: 'user',
             content: [
-            // TODO: use prompt render & i18n system later
-            // TODO: Module should have description & context length management
               { type: 'text', text: ''
-                + 'These are the contextual information retrieved or on-demand updated from other modules, you may use them as context for chat, or reference of the next action, tool call, etc.:\n'
```
nekomeowww: Do not change this line, what is the purpose?

packages/stage-ui/src/stores/llm.ts
```diff
@@ -52,41 +52,31 @@ async function streamFrom(model: string, chatProvider: ChatProvider, messages: M
...
+      : undefined,
+    async onEvent(event) {
+      await options?.onStreamEvent?.(event as StreamEvent)
+      if (event.type === 'error')
+        streamError = event.error ?? new Error('Stream error')
+    },
   })
+
+  if (streamError)
+    throw streamError
```
nekomeowww, line 79: What is the purpose of this change? They look identical...

PR #869, opened 2026-01-03
Feat/tts parallel processor

frigidom1024:
## Description
**Key improvements:**
- Pre-generates audio chunks concurrently (configurable limit)
- Maintains correct playback order despite concurrent execution
- Reduces playback gaps from 2000-7000ms to <2ms
- Fixed race condition in concurrent task management
- Improved type safety with explicit nullable `audioBuffer`

### Changes

**`packages/stage-ui/src/utils/tts.ts`**
- Added `createTTSOrderlyParallelProcessor` factory function
- Concurrency control via `maxConcurrent` option (default: 5)
- Result ordering via indexed pending map + `drain()` function
- Fixed race condition: increment `running` counter synchronously in `startNext()`
- Simplified task counting with counter instead of array
- Updated types to allow explicit `null` for `audioBuffer` (special tokens)

**`packages/stage-ui/src/components/scenes/Stage.vue`**
- Integrated parallel processor with `maxConcurrent: 3`
- Replaced serial queue with concurrent generation
- Added type-safe null checks for audioBuffer

### Performance Impact

Tests show dramatic improvements:
- **Average playback gap:** 3447.60ms → 0.90ms (**3830x faster**)
- **Smooth playback rate:** 20% → 100%
- **Max gap:** 7813.20ms → 2.50ms

[TTS_PERFORMANCE_TEST_REPORT.md](https://github.com/user-attachments/files/24416524/TTS_PERFORMANCE_TEST_REPORT.md)

## Linked Issues

None

### How It Works

```typescript
const processor = createTTSOrderlyParallelProcessor<TTSChunkItem>({
  maxConcurrent: 3,      // Max concurrent API requests
  debug: false,          
  processItem: async (item) => {
    // Generate audio for single chunk (runs concurrently)
  },
  onResult: (result) => {
    // Receive results in ORIGINAL order despite concurrent execution
  }
})

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -113,3 +113,6 @@ outputs/
 
 # MediaPipe task assets
 packages/model-driver-mediapipe/tasks/assets/*
+
+# Private documentation
+docs/private_docs/
```
nekomeowww: What is the purpose of this?

frigidom1024: This is a temporary directory for private documentation notes that I created locally. It's not intended for production and can be removed from the final merge.

nekomeowww:
Too complicated, try using Semaphore from `async-mutex`

PR #880, opened 2026-01-05
feat(stage-ui): support custom model

Lstmxx:
## Description

This PR adds support for custom user input in the `RadioCardManySelect` component and enables it for the Onboarding Model Selection step.

**Motivation:**
When using Zhipu as a provider, I wanted to use the `glm-4.5-flash` model, but it was not returned in Zhipu's model list API. This limitation prevented users from selecting valid models that exist but are not included in the provider's API response.

**Changes:**
- Updated [RadioCardManySelect.vue](cci:7://file:///Users/lstmxx/code/lstmxx/ai/airi/packages/stage-ui/src/components/menu/RadioCardManySelect.vue:0:0-0:0) to accept a new `allowCustom` prop.
  - When enabled, if a user searches for a term that doesn't match any existing items, it is presented as a selectable custom option.
  - If the currently selected model value is not present in the predefined list, it is now correctly displayed as a selected item.
- Enabled `:allow-custom="true"` in [step-model-selection.vue](cci:7://file:///Users/lstmxx/code/lstmxx/ai/airi/packages/stage-ui/src/components/scenarios/dialogs/onboarding/step-model-selection.vue:0:0-0:0) to allow users to manually input model IDs that are not in the provider's default list.

This solves the issue where users could not use valid models that were missing from the static or fetched provider lists.

## Additional Context

The custom option logic handles two cases:
1. **Pre-selection**: If `v-model` has a value not in `items`, it's temporarily added to the list so it can be verified visually.
2. **Search-to-add**: When searching, if `allowCustom` is true, the search query itself acts as a candidate for selection.

### DIFF
--- a/packages/stage-pages/src/pages/settings/modules/consciousness.vue
+++ b/packages/stage-pages/src/pages/settings/modules/consciousness.vue
@@ -179,6 +179,7 @@ function handleDeleteProvider(providerId: string) {
             v-model:search-query="modelSearchQuery"
             :items="providerModels.sort((a, b) => a.id === activeModel ? -1 : b.id === activeModel ? 1 : 0)"
             :searchable="true"
+            :allow-custom="true"
             :search-placeholder="t('settings.pages.modules.consciousness.sections.section.provider-model-selection.search_placeholder')"
             :search-no-results-title="t('settings.pages.modules.consciousness.sections.section.provider-model-selection.no_search_results')"
             :search-no-results-description="t('settings.pages.modules.consciousness.sections.section.provider-model-selection.no_search_results_description', { query: modelSearchQuery })"


shinohara-rin:
What about the consciousness setting page?

It would be a good idea to also add support for custom models here for consistency.

image

Lstmxx:
> What about the consciousness setting page?
> 
> It would be a good idea to also add support for custom models here for consistency.
> 
> image

Sounds good, I'll take a look at the consciousness settings page and apply the same enhancement for consistency. 
Will update the PR once it's ready for review.

Lstmxx:
@shinohara-rin Done. The consciousness settings page now supports custom models too.

nekomeowww:
Rebase needed.

Lstmxx:
> Rebase needed.

@nekomeowww done. pls check it

PR #882, opened 2026-01-05
feat(server): characters

luoling8192:


apps/server/src/schemas/providers.ts
```diff
@@ -0,0 +1,36 @@
+import type { InferInsertModel, InferSelectModel } from 'drizzle-orm'
+
+import { relations } from 'drizzle-orm'
+import { boolean, jsonb, pgTable, text, timestamp } from 'drizzle-orm/pg-core'
+
+import { nanoid } from '../utils/id'
+import { user } from './accounts'
+
+export const providerConfigs = pgTable(
+  'provider_configs',
```
nekomeowww, line 10: `user_settings_provider_configs`

PR #884, opened 2026-01-06
feat: Display character name in chat and add character name settings

ozekimasaki:
Display the active card's character name in the chat screen

Add assistant character name configuration feature in settings

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -92,7 +92,7 @@ catalogs:
       version: 1.4.10
     builder-util-runtime:
       specifier: ^9.3.1
-      version: 9.3.1
+      version: 9.5.1
     drizzle-valibot:
       specifier: ^0.4.2
       version: 0.4.2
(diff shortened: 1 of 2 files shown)


packages/stage-ui/src/stores/settings.ts
```diff
@@ -115,6 +115,8 @@ export const useSettings = defineStore('settings', () => {
 
   const [allowVisibleOnAllWorkspaces, resetAllowVisibleOnAllWorkspaces] = createResettableLocalStorage('settings/allow-visible-on-all-workspaces', true)
 
+  const [assistantCharacterName, resetAssistantCharacterName] = createResettableLocalStorage('settings/chat/assistant-character-name', '')
```
nekomeowww, line 118: I would not recommend doing this as Character Card already got cutomizable name in it.

nekomeowww:
This is invalid as you should create and edit your own character name in character card.

PR #885, opened 2026-01-06
feat: Add plugin store and management page

ozekimasaki:
Add plugin store for managing plugin state

Add plugins management page with loading and error states


packages/stage-ui/src/stores/mods/api/context-bridge.ts
```diff
@@ -40,6 +46,43 @@ export const useContextBridgeStore = defineStore('mods:api:context-bridge', () =
         broadcastContext(event.data as ContextMessage)
       }))
 
+      // Handle input:text events from plugins (e.g., YouTube Live Chat)
+      disposeHookFns.value.push(serverChannelStore.onEvent('input:text', async (event) => {
```
nekomeowww: While your implementation is correct, I would rather to suggest to use spark:notify for this. So we can implement not only individual chat message processing, but also throttled, summarized ones, and yet still open the possibility of using `priority: immediate` with `spark:notify` for Super Chat.

`input:text` should be used for direct user input only (actual user using AIRI on desktop or other platforms which triggers the `chatStore.send` directly).

packages/stage-pages/src/pages/settings/plugins/[pluginId].vue
```diff
@@ -0,0 +1,267 @@
...
+})
+
+const metadata = computed<PluginMetadata | null>(() => pluginInfo.value?.metadata ?? null)
+const connected = computed(() => pluginInfo.value?.connected ?? false)
+
+// Plugin settings
+const enabled = ref(true)
+const configFields = ref<Record<string, string>>({})
+
+// Known plugin-specific fields
```
Comment (nekomeowww) [inline packages/stage-pages/src/pages/settings/plugins/[pluginId].vue:37]: These are not yet ready, let's ignore these for now.
They should be able to run independently or built-in with Electron, and register capability to server-runtime in order to process in multi-agent pattern instead of single roundtrip for each chat.

packages/stage-layouts/src/components/Layouts/ViewControls/Inputs.vue
```diff
@@ -54,6 +67,8 @@ const viewControlsValueY = computed({
         return -live2dPosition.value.y
       case 'vrm':
         return vrmPosition.value.y * 100
+      case 'pngtuber':
+        return pngtuberPosition.value.y
```
nekomeowww: Let's remove these.

ozekimasaki:
## Review comments addressed

I've addressed all the review comments in two separate commits:

### Commit 1: Review comments fixes
- ✅ Fixed invalid syntax in plugin settings pages (Critical)
  - Removed invalid `class="..."` attributes in `[pluginId].vue:234` and `index.vue:81`
  - Fixed template syntax error (`top="[calc(...)]"` to `style` attribute)
- ✅ Removed `input:text` handler, use `spark:notify` instead
  - Removed `input:text` event handler from `context-bridge.ts`
  - Plugins should use `spark:notify` for text input as per @nekomeowww's comment
- ✅ Removed unimplemented plugin fields configuration
  - Commented out `knownPluginFields` with explanation (not ready yet)
- ✅ Removed pngtuber-related code as requested
  - Removed from `Inputs.vue`, `models/index.vue`, and `model-settings/index.vue`
- ✅ Improved CORS policy with helper function and comments
  - Added `setApiCorsHeaders` helper function to reduce duplication
  - Added explanatory comments about wildcard usage for plugin architecture

### Commit 2: Plugin metadata automation
- ✅ Replaced hardcoded plugin metadata with automatic generation
  - Added `vite-plugin-plugins-metadata` to scan plugins directory
  - Generates `plugins-metadata.generated.ts` from `package.json` files
  - Added `plugins-metadata.json` for icon and i18n key mapping

All changes have been tested and lint checks pass. Ready for re-review.

PR #899, opened 2026-01-07
fix(ui): color jittering

### LINKED ISSUE #864: bug: Dynamic theme hue causes color jitter on Context Flow devtool toggles
### Describe the bug

Behaviour:

  - In stage-tamagotchi, open Devtools → Context Flow with theme color set to “dynamic”.
  - Toggling any switch/tab (e.g., Show incoming/outgoing, channel filters, direction tabs) causes the accent
    color on the control to jitter/lag instead of transitioning smoothly.
  - The issue reproduces even when the Context Flow list is empty (no entries rendered), so it’s not tied to
    list re-rendering.

https://github.com/user-attachments/assets/7ea59717-d1f1-443c-bb59-1c7ab88c31d5

Possible root cause(unverified):

 - The app-level dynamic-hue animation (--chromatic-hue keyframes on document.documentElement in apps/stage-
    tamagotchi/src/renderer/App.vue) continuously animates the hue used by primary-* tokens.
 - Form controls (Checkbox.vue, SelectTab.vue) also run 200–250ms color transitions on toggle. The
    overlapping per-frame hue animation and the control’s own transition cause visible hue jumps when toggled.

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (32) x64 13th Gen Intel(R) Core(TM) i9-13900HX
    Memory: 6.81 GB / 31.73 GB
 Binaries:
    Node: 24.9.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.6.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.26.2 - C:\Program Files\nodejs\pnpm.CMD
 Browsers:
    Chrome: 143.0.7499.170
    Edge: Chromium (141.0.3537.57)
    Firefox: 143.0.4 - C:\Program Files\Mozilla Firefox\firefox.exe
    Firefox Developer Edition: 135.0 - C:\Program Files\Firefox Developer Edition\firefox.exe
    Internet Explorer: 11.0.26100.7309
```

### Validations

### Contributions

gg582:
## Description

Close #864 

Hello, I've noticed that bug #864 may happen because of a common UI component's class, which is widely shared.
In my limited, and biased opinion this bug might be "a hidden nail" when running into the `stage-web` mode because of WebGL's fast rendering performance; but I understand this is a mere hypothesis without verified benchmark between `stage-tamagotchi` and `stage-web`; so if you have different opinion- or more accurate approach to mine, please feel free to tell me.

This patch adds `is-interacting` class to Checkbox.vue and SelectTab.vue to keep minimal changes in source.
This class is declared as 
```css
.is_interacting {
  transition-property: none !important;
}
```
at fallback.css; while keeping other components.

Although this couldn't stand as `the best approach`, it is considered that this will not interfere with your team's code quality. Hence, this patch is easily removable if this makes unnecessary noise when releasing this software as a stable edition.

## Linked Issues

#864 

## Additional Context

Since I am an amateur and I have no relation with commercial development yet, my patch can be messy.
I shall be grateful if you consider this pull request as reasonable choice for preserving previous codes.
Thank you.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2829,7 +2829,7 @@ importers:
         version: 66.5.11
       '@wxt-dev/module-vue':
         specifier: ^1.0.3
-        version: 1.0.3(vite@8.0.0-beta.5(@types/node@24.10.4)(esbuild@0.25.12)(jiti@2.6.1)(less@4.5.1)(terser@5.44.1)(tsx@4.21.0)(yaml@2.8.2))(vue@3.5.25(typescript@5.9.3))(wxt@0.20.13(@types/node@24.10.4)(jiti@2.6.1)(less@4.5.1)(lightningcss@1.30.2)(rollup@4.54.0)(terser@5.44.1)(tsx@4.21.0)(yaml@2.8.2))
+        version: 1.0.3(vite@7.3.0(@types/node@24.10.4)(jiti@2.6.1)(less@4.5.1)(lightningcss@1.30.2)(terser@5.44.1)(tsx@4.21.0)(yaml@2.8.2))(vue@3.5.25(typescript@5.9.3))(wxt@0.20.13(@types/node@24.10.4)(jiti@2.6.1)(less@4.5.1)(lightningcss@1.30.2)(rollup@4.54.0)(terser@5.44.1)(tsx@4.21.0)(yaml@2.8.2))
       vue-tsc:
         specifier: ^3.1.8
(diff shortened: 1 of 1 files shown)


gg582:
Thanks for merging my patch, I was not sure if I can contribute, and help this project.
In my memory, I remember that my last pull request that was merged, was in 2025.
Happy New Year, and have a good day.

PR #902, opened 2026-01-07
chore(model-driver-mediapipe): streamline asset prep with retries

s3d-i:
### Bug Context
Pasted image 20260107173558

 - pnpm install postinstall failed while fetching MediaPipe task binaries from storage.googleapis.com with
    FetchError … ECONNRESET, leaving assets missing and aborting the install.

### Potential Causes (WSL + Clash Meta TUN)
 - TUN/transparent proxy resets long TLS streams or IPv4/IPv6 path changes mid-request.
 - DNS split-brain between Windows and WSL leading to flaky resolution for storage.googleapis.com.

### Fix
 - Sequential download flow with up to 3 retries and incremental backoff, using temp files + atomic rename to
 avoid partials.
 - Skip downloads when a usable file already exists; validate immediately after download.
 - Keep WASM assets copied/verified after task assets are ensured.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -109,23 +109,23 @@ export default defineConfig({
         ],
         outline: {
           level: 'deep',
-          label: 'このページの内容',
+          label: 'On this page',
         },
         docFooter: {
-          prev: '前のページ',
-          next: '次のページ',
+          prev: 'Previous page',
+          next: 'Next page',
         },
         editLink: {
           pattern: 'https://github.com/moeru-ai/airi/edit/main/docs/content/:path',
-          text: 'GitHub でこのページを編集',
+          text: 'Edit this page on GitHub',
         },
         lastUpdated: {
-          text: '最終更新',
+          text: 'Last updated',
         },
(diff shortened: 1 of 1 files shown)


packages/model-driver-mediapipe/tasks/prepare-tasks.ts
```diff
@@ -6,6 +6,7 @@ import type { VisionTaskAssets } from './tasks'
 import fs from 'node:fs/promises'
 
 import { Buffer } from 'node:buffer'
+import { setTimeout as delay } from 'node:timers/promises'
```
nekomeowww: Either one of this should be used.

- https://es-toolkit.dev/reference/util/attemptAsync.html#attemptasync-func
- https://std.moeru.ai/docs/packages/std/utils/with-retry

nekomeowww:
cc @sumimakito

packages/model-driver-mediapipe/tasks/prepare-tasks.ts
s3d-i: thank you, useful utils

PR #908, opened 2026-01-07
fix(stage-ui): fix notify agent toolcall

shinohara-rin:
tool calling in notifyagent should now work properly

image


Cargo.toml
```diff
@@ -1,5 +1,12 @@
 [workspace]
-members = [ "crates/tauri-plugin-ipc-audio-transcription-ort", "crates/tauri-plugin-ipc-audio-vad-ort", "crates/tauri-plugin-mcp", "crates/tauri-plugin-rdev", "crates/tauri-plugin-window-pass-through-on-hover", "crates/tauri-plugin-window-router-link" ]
+members = [
```
nekomeowww, line 2: Why you again.

shinohara-rin: autofix is certainly up to something :D

PR #924, opened 2026-01-09
fix(stage-pages): fix the cancel text display problem in settings -> data

leaft:
## Description

- Fix the cancel text display problem in settings-data, should be "Cancel Action"
image


nekomeowww:
Nice catch

PR #928, opened 2026-01-09
fix(stage-ui,server-*): implement handling for `input:text` from anywhere, support for Discord

skyline624:
## Description
  This PR fixes the Discord bot which was ignoring text mentions and improves how external messages are handled in the web interface.

### Key Changes
- **Discord Bot:** Added `input:text` handling in `context-bridge.ts`. Previously, the bot sent the event but the stage-web client wasn't listening.
 - **User Identification:** Updated the Discord adapter to send the user's `displayName`. Added a marker `(From Discord user X):` to messages so the LLM can identify the speaker.
 - **Session Isolation:**
     - Modified `chatStore.ts` to support background/targeted sessions (`targetSessionId`).
     - Discord messages are now routed to a dedicated `'discord'` session.
 - **UI & Filtering:**
     - Prevented Discord messages from cluttering the main web chat UI by using background processing.
     - Added a filter to ensure AI responses are only sent back to Discord if the triggering message originated from there.

 ## Motivation
 Makes the Discord integration fully functional for text chat without polluting the primary user's web interface experience.

### DIFF
--- a/packages/stage-ui/src/stores/chat.ts
+++ b/packages/stage-ui/src/stores/chat.ts
@@ -379,6 +379,8 @@ export const useChatStore = defineStore('chat', () => {
     // Use a local object for building the message to avoid polluting the UI for background sessions
     const buildingMessage: StreamingAssistantMessage = { role: 'assistant', content: '', slices: [], tool_results: [], createdAt: Date.now() }
 
+    // NOTICE: Clone into reactive state only for the foreground session
+    // to avoid background streams mutating UI.
     const updateUI = () => {
       if (isForegroundSession()) {
         streamingMessage.value = JSON.parse(JSON.stringify(buildingMessage))


apps/stage-web/vite.config.ts
```diff
@@ -61,6 +61,7 @@ export default defineConfig({
     },
   },
   server: {
+    host: true,
```
nekomeowww:
```suggestion
```

packages/stage-ui/src/stores/mods/api/context-bridge.ts
```diff
@@ -40,61 +59,100 @@ export const useContextBridgeStore = defineStore('mods:api:context-bridge', () =
...
+            id,
+            contextId: id,
+            source: event.source,
+            strategy: ContextUpdateStrategy.AppendSelf,
+            createdAt: Date.now(),
+            content: `The input is coming from Discord channel ${discord.channelId} (Guild: ${discord.guildId}).`,
+            metadata: {
+              discord,
+            },
+          })
```
nekomeowww: I am so sorry that the `server-shared`/`channel-server`'s type is not finalized and still experimenting. Is `context:update` ok for consuming the `input:text` event from Discord? Should we do `spark:notify` instead?

packages/stage-ui/src/stores/mods/api/context-bridge.ts
```diff
@@ -40,61 +59,100 @@ export const useContextBridgeStore = defineStore('mods:api:context-bridge', () =
...
+        if (activeProvider.value && activeModel.value) {
+          const chatProvider = await providersStore.getProviderInstance<ChatProvider>(activeProvider.value)
+
+          let messageText = text
+          let targetSessionId: string | undefined
+
+          if (discord?.guildMember?.displayName) {
+            messageText = `(From Discord user ${discord.guildMember.displayName}): ${text}`
+            targetSessionId = 'discord'
+          }
```
nekomeowww: Nice work here. This reminds me that having a context formatter is essential to all integrations.

packages/stage-ui/src/stores/mods/api/context-bridge.ts
```diff
@@ -40,61 +59,100 @@ export const useContextBridgeStore = defineStore('mods:api:context-bridge', () =
...
+
+          await chatStore.send(messageText, {
+            model: activeModel.value,
+            chatProvider,
+          }, targetSessionId)
```
nekomeowww: Oh I see, so here use as `context:update` is fine since you use `chatStore.send(...)` for reacting to stage.

plugins/airi-plugin-web-extension/package.json
```diff
@@ -5,14 +5,14 @@
...
-    "zip:firefox": "wxt zip -b firefox",
+    "dev": "NODE_OPTIONS='--localstorage-file=/tmp/airi-wxt-ls' wxt",
+    "dev:firefox": "NODE_OPTIONS='--localstorage-file=/tmp/airi-wxt-ls' wxt -b firefox",
+    "build": "NODE_OPTIONS='--localstorage-file=/tmp/airi-wxt-ls' wxt build",
+    "build:firefox": "NODE_OPTIONS='--localstorage-file=/tmp/airi-wxt-ls' wxt build -b firefox",
+    "zip": "NODE_OPTIONS='--localstorage-file=/tmp/airi-wxt-ls' wxt zip",
+    "zip:firefox": "NODE_OPTIONS='--localstorage-file=/tmp/airi-wxt-ls' wxt zip -b firefox",
     "compile": "vue-tsc --noEmit",
-    "postinstall": "wxt prepare"
+    "postinstall": "NODE_OPTIONS='--localstorage-file=/tmp/airi-wxt-ls' wxt prepare"
```
nekomeowww: Perhaps let's not change this for now?

services/discord-bot/src/adapters/airi-adapter.ts
nekomeowww, line 1: Much appreciated for implementing this. Thanks.

packages/stage-ui/src/stores/mods/api/context-bridge.ts
```diff
@@ -40,61 +48,107 @@ export const useContextBridgeStore = defineStore('mods:api:context-bridge', () =
...
 
         chatStore.onAssistantMessage(async (message, _messageText, context) => {
+          let discord
+          const inputContent = Array.isArray(context.input.content)
+            ? context.input.content.map(c => c.type === 'text' ? c.text : '').join('')
+            : context.input.content
+
+          if (typeof inputContent === 'string' && inputContent.includes('(From Discord user')) {
+            discord = Object.values(context.contexts).flat().find(c => c.metadata?.discord)?.metadata?.discord
+          }
```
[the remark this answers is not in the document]
nekomeowww: /gemini review

Having `getDiscord...(...)` implementation here and handles specifically for Discord is not extendable for future integrations, what approach would you suggest to have both Discord context builder and other source/module integrated in? Perhaps, can we ask the input:text to build essential messages / context for consumer side from the module that initially emits? What is the best option here?

plugins/airi-plugin-web-extension/package.json
skyline624: Oops, unintended change. Reverted.

packages/stage-ui/src/stores/mods/api/context-bridge.ts
skyline624: Yes exactly, I used `input:text` to trigger a direct chat interaction which feels more natural for a direct reply pattern.

skyline624:
Done! I've applied the latest code review suggestions and fixed the remaining points. It should be ready for another look/approval
  for deployment. Thanks!

packages/stage-ui/src/stores/mods/api/channel-server.ts
```diff
@@ -45,7 +45,7 @@ export const useModsServerChannelStore = defineStore('mods:channels:proj-airi:se
     initializing.value = new Promise<void>((resolve) => {
       client.value = new Client({
         name: isStageWeb() ? WebSocketEventSource.StageWeb : isStageTamagotchi() ? WebSocketEventSource.StageTamagotchi : WebSocketEventSource.StageWeb,
-        url: import.meta.env.VITE_AIRI_WS_URL || 'ws://localhost:6121/ws',
+        url: import.meta.env.VITE_AIRI_WS_URL || `ws://${window.location.hostname}:6121/ws`,
```
nekomeowww:
```suggestion
url: import.meta.env.VITE_AIRI_WS_URL || 'ws://localhost:6121/ws',
```

plugins/airi-plugin-web-extension/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@proj-airi/airi-plugin-web-extension",
   "type": "module",
-  "version": "0.8.1-beta.8",
+  "version": "0.8.1-beta.7",
```
nekomeowww:
```suggestion
  "version": "0.8.1-beta.8",
```

services/discord-bot/src/adapters/airi-adapter.ts
```diff
@@ -129,16 +134,71 @@ export class DiscordAdapter {
...
+      try {
+        const { message, discord } = event.data as {
+          message: { content: string }
+          discord?: { channelId: string }
+        }
+
+        if (discord?.channelId) {
+          const channel = await this.discordClient.channels.fetch(discord.channelId)
+          if (channel?.isTextBased() && 'send' in channel) {
+            await (channel as any).send(message.content)
```
nekomeowww: Can we not to use any here?

pnpm-workspace.yaml
```diff
@@ -37,6 +37,7 @@ catalog:
   '@pinia/testing': ^1.0.3
   '@proj-airi/drizzle-duckdb-wasm': ^0.4.29
   '@proj-airi/iconify-meteocons': ^0.1.4
+  '@snazzah/davey': ^0.1.9
```
nekomeowww:
```suggestion
```

packages/server-runtime/src/bin/run.ts
```diff
@@ -12,4 +12,5 @@ serve(app, {
   // @ts-expect-error - the .crossws property wasn't extended in types
   plugins: [ws({ resolve: async req => (await app.fetch(req)).crossws })],
   port: env.PORT ? Number(env.PORT) : 6121,
+  hostname: env.HOST || '0.0.0.0',
```
nekomeowww:
```suggestion
```

services/discord-bot/src/adapters/airi-adapter.ts
skyline624: Absolutely. I'll refactor this to use a property check instead of any to ensure the channel is sendable while keeping it type-safe.

pnpm-workspace.yaml
skyline624: Regarding the change in pnpm-workspace.yaml, @snazzah/davey is a dependency required by the new Discord bot service. I understand the preference to keep the global catalog clean of dependencies used by only a single service.
As an alternative, I can remove it from the root pnpm-workspace.yaml and specify the version explicitly in services/discord-bot/package.json instead of using catalog:.
Would you prefer this approach, or is there another library you'd rather have me use for the bot's functionality?

skyline624:
Addressed all review comments:
- Reverted changes to vite.config.ts, server-runtime, and pnpm-workspace.yaml.
- Moved @snazzah/davey dependency definition directly into services/discord-bot to keep the root workspace clean.
- Refactored airi-adapter.ts to use a proper Type Guard instead of any, ensuring type safety for channel.send.

Tests are passing locally. Ready for another look!

PR #931, opened 2026-01-10
feat(stage-tamagotchi): disable back button on settings home page

leaft:
## Description
我发现 `layout: settings` 实际上使用的是 `packages/stage-layouts/src/layouts/settings.vue`，而非 `apps/stage-tamagotchi/src/renderer/layouts/settings.vue`，因此现在设置页首页存在一个无用的返回按钮。
或许我们应该把后者移除。
顺便，`apps/stage-tamagotchi/src/renderer/pages/settings/index.vue` 也没有被使用，现在用的是 `packages/stage-pages/src/pages/settings/index.vue`。

I noticed that `layout: settings` is actually using `packages/stage-layouts/src/layouts/settings.vue` instead of `apps/stage-tamagotchi/src/renderer/layouts/settings.vue`. This is why the settings home page now has an unnecessary back button. Maybe we should remove the latter.
BTW, apps/stage-tamagotchi/src/renderer/pages/settings/index.vue isn't being used either; it’s currently using packages/stage-pages/src/pages/settings/index.vue instead.

image


packages/stage-layouts/src/layouts/settings.vue
```diff
@@ -166,6 +166,7 @@ onMounted(() => updateThemeColor())
       <PageHeader
         :title="routeHeaderMetadata?.title"
         :subtitle="routeHeaderMetadata?.subtitle"
+        :disable-back-button="route.path === '/settings'"
```
nekomeowww, line 173:
```suggestion
        :disable-back-button="isStageTamagotchi() && route.path === '/settings'"
```

Perhaps...

nekomeowww:
> I noticed that layout: settings is actually using packages/stage-layouts/src/layouts/settings.vue instead of apps/stage-tamagotchi/src/renderer/layouts/settings.vue. This is why the settings home page now has an unnecessary back button. Maybe we should remove the latter.
> BTW, apps/stage-tamagotchi/src/renderer/pages/settings/index.vue isn't being used either; it’s currently using packages/stage-pages/src/pages/settings/index.vue instead.

cc @LemonNekoGH can we? I remember that tamagotchi has different layout setting than web & pocket

PR #940, opened 2026-01-12
feat(stage-tamagotchi): add icon size settings for tamagotchi

leaft:
## Description

- 添加侧边控制栏图标大小设置项-自动/默认/小图标。
- 修复 `ComboboxContent` 导致的重叠问题，使用 `ComboboxPortal`，问题如下图所示。

- Add sidebar controls island icon size settings: Auto(Responsive) / Large(Default) / Small.
- Fix overlapping issues caused by `ComboboxContent` by using `ComboboxPortal` (see the image below).

IMG_20260112_235921

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/components/WithScreenCapture.vue
+++ b/apps/stage-tamagotchi/src/renderer/components/WithScreenCapture.vue
@@ -1,0 +1,111 @@
+<script setup lang="ts">
+import type { SourcesOptions } from 'electron'
+
+import { useElectronScreenCapture } from '@proj-airi/electron-screen-capture/vue'
+import { Button } from '@proj-airi/ui'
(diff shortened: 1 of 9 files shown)


packages/stage-ui/src/stores/settings.ts
```diff
@@ -274,3 +274,16 @@ export const useSettingsAudioDevice = defineStore('settings-audio-devices', () =
     resetState,
   }
 })
+
+export const useControlsIslandIconSize = defineStore('controls-island-icon-size', () => {
```
nekomeowww: Perhaps, `useControlsIslandSettings`? At least icon size should not be considered as dedicated store?

nekomeowww (approved):

nekomeowww:
You are right, we can later split it down into different settings for each modules (appearance, language, general, etc.)

PR #942, opened 2026-01-12
feat(discord): add guild segmentation, DM support and context enrichment

skyline624:
This PR significantly enhances the Discord integration by introducing multi-server memory segmentation, native Direct Message (DM) support, and richer context propagation.

  Previously, AIRI treated all Discord interactions as a single monolithic session ("discord"), causing conversation history to bleed between different servers and users. This update ensures that each Guild (and each DM user) gets its own isolated memory context.

  Key Features

   * Guild-Based Memory Segmentation:
       * Interactions in servers are now isolated by sessionId: 'discord-guild-{GUILD_ID}'.
       * Interactions in DMs are isolated by sessionId: 'discord-dm-{USER_ID}'.
       * This prevents context leaking between different communities.

   * Native Direct Message Support:
       * Added DirectMessages gateway intent and Partials.Channel support.
       * The bot now responds to DMs even without explicit mentions, allowing for a more natural conversation flow.

   * Context Enrichment:
       * Added guildName to the Discord metadata interface.
       * The LLM prompt is now prefixed with contextual info:
           * Server: (From Discord user Bob on server 'Gaming Community'): ...
           * DM: (From Discord user Bob in Direct Message): ...
       * This allows the persona to adapt its tone based on where the conversation is happening.

  Technical Changes

   * packages/server-shared: Updated Discord interface to include guildName.
   * services/discord-bot:
       * Updated airi-adapter.ts to capture and send guildName.
       * Implemented logic to handle DM events and removed mention requirement for private chats.
   * packages/stage-ui:
       * Updated context-bridge.ts to generate dynamic targetSessionId based on the Discord context.
       * Enhanced message prefix generation for better LLM awareness.

  Verification
   - Tested on multiple Discord servers (memory is isolated).
   - Tested in DMs (bot responds without mentions).
   - Verified context propagation to the LLM via input:text events.

### DIFF
--- a/services/discord-bot/src/adapters/airi-adapter.ts
+++ b/services/discord-bot/src/adapters/airi-adapter.ts
@@ -164,7 +164,33 @@ export class DiscordAdapter {
         if (message?.content && discordContext?.channelId) {
           const channel = await this.discordClient.channels.fetch(discordContext.channelId)
           if (channel?.isTextBased() && 'send' in channel && typeof channel.send === 'function') {
-            await channel.send(message.content)
+            const content = message.content
+            if (content.length <= 2000) {
+              await channel.send(content)
+            }
+            else {
+              let remaining = content
+              while (remaining.length > 0) {
+                let chunkSize = 2000
+                if (remaining.length > 2000) {
+                  // Try to split at the last newline before 2000
+                  const lastNewline = remaining.lastIndexOf('\n', 2000)
+                  if (lastNewline > -1) {
+                    chunkSize = lastNewline
+                  }
+                  else {
+                    // Fallback to last space
+                    const lastSpace = remaining.lastIndexOf(' ', 2000)
+                    if (lastSpace > -1)
+                      chunkSize = lastSpace
+                  }
+                }
+
+                const chunk = remaining.slice(0, chunkSize)
+                await channel.send(chunk)
+                remaining = remaining.slice(chunkSize).trim()
+              }
+            }
           }
         }
       }


nekomeowww (dismissed):

packages/stage-ui/src/stores/mods/api/context-bridge.ts
```diff
@@ -92,8 +93,30 @@ export const useContextBridgeStore = defineStore('mods:api:context-bridge', () =
...
+            if (discord.guildId) {
+              targetSessionId = `discord-guild-${discord.guildId}`
+            }
+            else {
+              targetSessionId = `discord-dm-${discord.guildMember?.id || 'unknown'}`
+            }
+          }
+          else if (overrides?.messagePrefix) {
             messageText = `${overrides.messagePrefix}${text}`
+          }
```
nekomeowww: These should be calculated in Discord...

skyline624: I completely understand. Implementing it here first allowed me to quickly validate the context enrichment and ensure the LLM responses were consistent. Now that the logic is verified, I'll refactor this to move it directly into the Discord service.

skyline624:
Okay, that should be better now. I tried to follow best practices to ensure everything isclean and robust, so I hope I haven't forgotten anything.
There are still two or three things to look into, such as the bot remaining online even when the web version isn't open and therefore not responding, or even when the bot is disabled in the configuration. One thing at a time.

skyline624:
This seems relevant to me unless there is a limit on token generation.

Issue #866, opened 2026-01-02
feature request: add a HomeBrew Tap


Somiona:
### Clear and concise description of the problem

It would be convenient to create a new repository moeru-ai/tap for the homebrew to auto install on macos. 

### Suggested solution / Ideas

Here is an example brew file, I assume it would be working :D
```ruby
cask "airi" do
  arch arm: "arm64", intel: "x64"

  version "0.8.1-beta.2"
  sha256 arm:   "f71dd085be89608d8236cf184d70016cb7e43db491443b30467a696c56882c4c",
         intel: "a6772ec3c6632251ab99b13c7bf38033883f046eea4a8090393f19e19c3b5be8"

  url "https://github.com/moeru-ai/airi/releases/download/v#{version}/AIRI-#{version}-darwin-#{arch}.dmg",
      verified: "github.com/moeru-ai/airi/"
  name "AIRI"
  desc "AI VTuber application, similar to Neuro-sama"
  homepage "https://airi.moeru.ai/"

  app "AIRI.app"

  # The application is not notarized by Apple yet, so you may need to clear the quarantine attribute.
  caveats do
    <<~EOS
      If you receive a "damaged" or "can't be opened" error, run the following command:
        sudo xattr -c '/Applications/AIRI.app'
    EOS
  end

  zap trash: [
    "~/Library/Application Support/airi",
    "~/Library/Preferences/moeru-ai.airi.plist",
    "~/Library/Saved Application State/moeru-ai.airi.savedState",
  ]
end
```

### Alternative

Maybe submitting an uptate to the homebrew casks tap would be good, however I dont know how well this would went through as it is a cask.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

LemonNekoGH:
So cool! Thanks for submit this idea!

mvanhorn:
Opened #1431 with a ready-to-go `homebrew/` directory containing the Homebrew cask (arm64 + x64), a daily auto-bump CI workflow, and a README. Maintainers can copy these files into a new `moeru-ai/homebrew-tap` repo to enable `brew install --cask airi`.

Garfield550:
[...]

PR #943, opened 2026-01-13
feat: add GitHub Actions workflow for automated Homebrew tap updates

lockrush-dev:
Add a workflow that automatically updates the Homebrew Cask in the moeru-ai/tap repository whenever a new release is published.

The workflow:
- Triggers on release.published events
- Downloads DMG files for both arm64 and x64 architectures
- Calculates SHA256 checksums automatically
- Generates/updates the Cask file with the correct format
- Commits and pushes changes to the tap repository

## Linked Issues

#866

### DIFF
--- a/.github/workflows/update-homebrew-tap.yml
+++ b/.github/workflows/update-homebrew-tap.yml
@@ -1,0 +1,126 @@
+name: Update Homebrew Tap
+
+on:
+  release:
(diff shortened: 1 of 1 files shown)


nekomeowww:
We don't have the tap repository, and not prepared for code sign, how can this script work?

PR #951, opened 2026-01-14
feat(stage-ui,server): use optimistic request for provider catalog

luoling8192:


packages/stage-ui/src/composables/use-optimistic.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (typeof rollback === 'function') {
+        await rollback()
+      }
+      if (onError) {
+        await onError(err)
+      }
+      throw err
+    }
+  }, { immediate: true })
+}
```
nekomeowww:
```suggestion
export interface UseOptimisticOptions<T, R, E = unknown> {
  /**
   * The optimistic update logic.
   * Should return a rollback function.
   */
  apply: () => Promise<(() => Promise<void> | void)> | (() => Promise<void> | void)
  /**
   * The actual async task (e.g., API call).
   */
  action: () => Promise<T>
  /**
   * Optional callback after successful action to refine state (e.g., replacing temp IDs).
   */
  onSuccess?: (result: T) => Promise<R> | R
  /**
   * Optional callback on error. Rollback is handled automatically.
   */
  onError?: (error?: E | null) => void | Promise<void>
}

/**
 * A wrapper for performing optimistic updates with automatic rollback.
 * Integrates with useAsyncState for loading/error tracking.
 */
export function useOptimistic<T, R = T, E = unknown>(options: UseOptimisticOptions<T, R>) {
  const { apply, action, onSuccess, onError } = options

  return useAsyncState(async () => {
    const rollback = await apply()

    try {
      const result = await action()
      if (onSuccess) {
        return await onSuccess(result)
      }
      return result as unknown as R
    }
    catch (err: E) {
      if (typeof rollback === 'function') {
        await rollback()
      }
      if (onError) {
        await onError(err)
      }
      throw err
    }
  }, { immediate: true })
}
```

packages/stage-ui/src/composables/use-optimistic.ts
```diff
@@ -0,0 +1,50 @@
...
+    catch (err) {
+      if (typeof rollback === 'function') {
+        await rollback()
+      }
+      if (onError) {
+        await onError(err)
+      }
+      throw err
+    }
+  }, { immediate: true })
```
nekomeowww: I think `immediate` should be option too, perhaps... `lazy`?

packages/stage-ui/src/stores/provider-catalog.ts
```diff
@@ -43,78 +45,135 @@ export const useProviderCatalogStore = defineStore('provider-catalog', () => {
...
+    const provider: ProviderCatalogProvider = {
+      id,
+      definitionId,
+      name: definition.name,
+      config: initialConfig,
+      validated: false,
+      validationBypassed: false,
+    }
+
+    return useOptimistic<any, any>({
```
nekomeowww: Cool, love this.

packages/stage-ui/src/composables/use-optimistic.ts
```diff
@@ -0,0 +1,50 @@
...
+   * Optional callback on error. Rollback is handled automatically.
+   */
+  onError?: (error: unknown) => void | Promise<void>
+}
+
+/**
+ * A wrapper for performing optimistic updates with automatic rollback.
+ * Integrates with useAsyncState for loading/error tracking.
+ */
+export function useOptimistic<T, R = T>(options: UseOptimisticOptions<T, R>) {
```
nekomeowww: Pehaps... `useMutateOptimistic`? Or `useOptimisticMutate`?

luoling8192: Could you explain why should we use `mutate`?

packages/stage-ui/src/composables/use-optimistic.ts
```diff
@@ -17,14 +17,14 @@ export interface UseOptimisticOptions<T, R> {
...
-  onError?: (error: unknown) => void | Promise<void>
+  onError?: (error?: E | null) => void | Promise<void>
 }
 
 /**
  * A wrapper for performing optimistic updates with automatic rollback.
  * Integrates with useAsyncState for loading/error tracking.
  */
-export function useOptimistic<T, R = T>(options: UseOptimisticOptions<T, R>) {
+export function useOptimistic<T, R = T, E = unknown>(options: UseOptimisticOptions<T, R>) {
```
nekomeowww:
```suggestion
export function useOptimistic<T, R = T, E = unknown>(options: UseOptimisticOptions<T, R, E>) {
```

packages/stage-ui/src/composables/use-optimistic.ts
nekomeowww: - https://pinia-colada.esm.dev/guide/mutations.html
- https://tanstack.com/query/latest/docs/framework/react/guides/mutations
- https://swr.vercel.app/docs/mutation

Everyone calls this `mutate`

luoling8192: However, what inspired me is this React implementation, https://react.dev/reference/react/useOptimistic

nekomeowww: Then should we divide this into two composables?

PR #968, opened 2026-01-17
feat(stage-ui): add n1n provider

n1n-api:
Add n1n.ai as an OpenAI-compatible AI provider. Includes English and Simplified Chinese translations.

### DIFF
--- a/packages/stage-ui/src/libs/providers/providers/n1n/index.ts
+++ b/packages/stage-ui/src/libs/providers/providers/n1n/index.ts
@@ -40,7 +40,7 @@ export const providerN1N = defineProvider<N1NConfig>({
         }),
     }),
     createProvider(config) {
-        return createOpenAI(config.apiKey as string, config.baseUrl)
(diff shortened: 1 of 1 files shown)


n1n-api:
I have read the CLA Document and I hereby sign the CLA

shinohara-rin:
As of the current UI, the chat providers already take up a whole page.

I don't feel like it's a great idea to keep adding cards for every single provider out there, at least not for the OpenAI-compatible ones like this, since users can already use it with a trivial baseurl setup.

I think this PR can wait until we have a better UI to accommodate it

n1n-api:
> As of the current UI, the chat providers already take up a whole page.
> 
> I don't feel like it's a great idea to keep adding cards for every single provider out there, at least not for the OpenAI-compatible ones like this, since users can already use it with a trivial baseurl setup.
> 
> I think this PR can wait until we have a better UI to accommodate it

Thank you for considering us! Can not wait to see the new UI in the soon future :)

PR #981, opened 2026-01-21
Feat/tts parallel processor

frigidom1024:
## Description

This PR introduces parallel TTS (Text-to-Speech) processing capabilities to the `pipelines-audio` package. The implementation embeds concurrency control directly into the `speech-pipeline` system, allowing multiple TTS requests to be processed concurrently while maintaining strict ordering guarantees for callbacks and audio playback.

### Motivation

Previously, TTS requests were processed serially—each segment had to complete before the next one could start. This introduced unnecessary latency, especially when working with cloud-based TTS services where network I/O dominates processing time. By allowing multiple segments to be processed in parallel, we can:

- **Reduce end-to-end latency** — Subsequent segments can be generated while current segments are being played
- **Better resource utilization** — Overlap network I/O with audio playback
- **Improved responsiveness** — Faster response to real-time streaming scenarios

### Implementation Overview

**New Module**: [`parallel-tts.ts`](packages/pipelines-audio/src/parallel-tts.ts)
- Exports `createParallelTts<T>()` factory function
- Uses `Semaphore` from `async-mutex` to limit concurrent requests
- Implements a buffering mechanism that guarantees `onComplete` callbacks are invoked in submission order (0, 1, 2, ...), even when TTS requests complete out of order
- Clean separation of concerns: handles only concurrency and ordering, not business logic

**Integration**: [`speech-pipeline.ts`](packages/pipelines-audio/src/speech-pipeline.ts:92)
- Added `ttsConcurrency` option (default: `3`)
- Parallel processor is created per-intent in `runIntent()`
- TTS requests are submitted via `parallel.submit(index, request, ttsFn)`
- Event emission and playback scheduling remain in the pipeline for clarity

### Configuration

```typescript
const pipeline = createSpeechPipeline({
  ttsConcurrency: 3,  // Number of parallel TTS requests
  // 1 = serial (original behavior)
  // 2-4 = recommended for most use cases
  // ... other options
})
```

### Example Flow

```
Segment 0 ─┐
Segment 1 ─┼─→ Parallel Processor (concurrency: 3) ─→ Ordered Callbacks
Segment 2 ─┤       ├─ Request 0 (TTS takes 200ms) ─────┐
Segment 3 ─┘       ├─ Request 1 (TTS takes 100ms) ─────┼→ onComplete(0, audio0)
Segment 4 ──→       └─ Request 2 (TTS takes 150ms) ─────┼→ onComplete(1, audio1)
                    ...                               └→ onComplete(2, audio2)
```

Even though Request 1 completes first (100ms), its callback fires after Request 0's callback completes, preserving the correct playback order.

---

## Linked Issues

---

### Key Design Decisions

1. **Embedded Implementation**: Rather than a standalone service, parallel processing is embedded in `runIntent`. This keeps the implementation simple and avoids complex state management across intents.

2. **Order Preservation**: The index-based buffering ensures that audio segments are emitted and scheduled in the correct order, preventing playback from becoming garbled.

3. **Per-Intent Processor**: A new parallel processor is created for each intent. This ensures proper isolation and cleanup when intents complete or are canceled.

### Potential Future Enhancements

- Add metrics/observability for parallel processor utilization
- Consider dynamic concurrency adjustment based on TTS service response times
- Add retry logic for transient TTS failures

---
*This PR description was prepared with AI assistance.*

### DIFF
--- a/packages/pipelines-audio/src/parallel-tts.ts
+++ b/packages/pipelines-audio/src/parallel-tts.ts
@@ -20,14 +20,12 @@ export function createParallelTts<T>(options: {
   const buffer = new Map<number, QueuedItem<T>>()
   let nextIndex = 0
 
-  async function tryFlush() {
-    let flushed = 0
+  function tryFlush() {
     while (buffer.has(nextIndex)) {
       const item = buffer.get(nextIndex)!
       buffer.delete(nextIndex)
       options.onComplete(item.index, item.audio, item.request)
       nextIndex++
-      flushed++
     }
   }
 
@@ -41,7 +39,7 @@ export function createParallelTts<T>(options: {
         }
 
         buffer.set(index, { index, audio, request })
-        await tryFlush()
+        tryFlush()
       }
       finally {
         release()
--- a/packages/pipelines-audio/src/speech-pipeline.ts
+++ b/packages/pipelines-audio/src/speech-pipeline.ts
@@ -162,7 +162,10 @@ export function createSpeechPipeline<TAudio>(options: SpeechPipelineOptions<TAud
         const currentIndex = segmentIndex++
 
         parallel.submit(currentIndex, request, () =>
-          options.tts(request, intent.controller.signal))
+          options.tts(request, intent.controller.signal).catch((err) => {
+            logger.warn('TTS generation failed:', err)
+            return null
+          }))
       }
 
       reader.releaseLock()


frigidom1024:
Recreated this PR after fixing an earlier branch history issue.
The previous PR was closed to keep the review clean.

PR #1009, opened 2026-01-29
Release v0.8.4 with build fixes

sebastian-zm:
## Description

This pull request is exactly the same as the previous one (#1008) except it adds a release commit, so that everyone benefits immediately.


nekomeowww:
We cannot use merge commit to release so I think I will just close this as #1008 was merged already.

PR #1019, opened 2026-02-02
fix(nix): nix workflows are more resilient

sebastian-zm:
## Description

Sometimes the nix hash update workflows run in parallel: both commit, one pushes, the other tries to push and runs into a problem: it doesn't have the latest commit (needs to pull again) and the workflow errors without updating the hash. Here I give a fix to that problem.


sebastian-zm:
@gemini-code-assist sorry, I accidentally included a commit I didn't want to include. Please try reviewing again please.

Weathercold:
hmm, I thought the concurrency group would prevent them from running in parallel
also the option doesn't work

Issue #1031, opened 2026-02-12
bug: entire /settings navbar is clickable and redirects to main page in web


Zazzik1:
### Summary
On the `/settings` page (desktop version of browser UI), the top navigation bar contains the AIRI logo in the top-left corner that redirects to the main page `/`.

However, the issue is that the entire navbar area (a wide invisible box) acts as a link, not just the logo.

This means clicking anywhere within the navbar (not only on the logo/icon) triggers a redirect to the main page.

### Steps to Reproduce

1. Navigate to /settings https://airi.moeru.ai/settings
2. Click outside of the "Welcome to AIRI!" modal to close it
3. Click anywhere on the top navbar (outside the icon area)
4. Observe that it redirects to the main page

### Expected Behavior

Only the `AIRI` text and the icon should redirect to the main page.

(Or is the entire navbar intentionally designed to be clickable?)

### Fix

This could be fixed by adding `width: max-content` to the styles of `a` element that wraps the logo. 

#### Before
<img width="1461" height="326" alt="Image" src="https://github.com/user-attachments/assets/a6c1be76-29a3-458d-b721-a05bb932dd66" />

#### After
<img width="1463" height="346" alt="Image" src="https://github.com/user-attachments/assets/4902cb6c-90e5-47b1-a5f9-e21b118ded57" />

### Additional Questions
This issue does not affect mobile devices, since the navbar is hidden there. Would it make sense to add a simple “Back” button on mobile for consistency, or are the browser navigation buttons enough to get back to the main page (the subpages on the settings page have similar "Back" buttons, but the main page doesn't have it on mobile)?

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (12) x64 AMD Ryzen 5 4600H with Radeon Graphics
    Memory: 5.23 GB / 15.37 GB
  Binaries:
    Node: 23.5.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 144.0.7559.133
    Edge: Chromium (140.0.3485.54)
    Internet Explorer: 11.0.26100.1882
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
[...]

PR #1035, opened 2026-02-15
fix(stage-layouts): HeaderLink component no longer makes entire Header clickable

Zazzik1:
# Summary

The `stage-web` uses the `Header` component. On the settings page the entire header is clickable and redirects to the main page (`/`) - users can click on the AIRI button, but they can also click wherever on the wide invisible box and it results in the same effect. This PR fixes this bug by limiting the clickable area to the width of the `HeaderLink` component.

## Linked Issues

Resolves: #1031 

## Extra notes regarding tests

I tried to add tests to check if clicking outside the AIRI logo triggers a page redirect (in `stage-layout`, using Vitest + Playwright), but I ran into constant errors such as:
`Failed to resolve import "~build/time" from "../stage-ui/src/composables/use-build-info.ts"`
and some issues related to Cubism (when I tried to add those tests in `stage-web` and merge the config from vite.config.ts).

It seems that when running tests in browser mode (vitest.config.ts → test → browser), the `@proj-airi/ui` module cannot be mocked with vi.mock(). As far as I understand, vite tries to resolve and bundle the module before the tests run, which causes the `~build/*` alias errors (maybe it would work with proper resolve.alias). Because of this, the tests fail before any mocks can work.

The reason for using browser mode was to test how the CSS is rendered, since it is a bug in the styles that cannot be detected in the default Node environment. I didn't want to simply check the presence of `width: max-content` in test, but rather to check it if clicking works. However, because of these module issues, I wasn’t able to make this work.

I think that the easiest way to test this behavior to prevent regressions would be with e2e tests

Here's the approach that I used:
```sh
cd packages/stage-layout
# added "@vitest/browser-playwright": "catalog:vitest" to dev dependencies in package.json
pnpm install --workspace-root=false
pnpm add -D @vitejs/plugin-vue@^6.0.3 @vue/test-utils
```
config was similar to this:
```ts
import { playwright } from '@vitest/browser-playwright'
// […]
test: {
    globals: true,
    browser: {
      enabled: true,
      provider: playwright(),
      instances: [
        { browser: 'chromium' },
      ],
      headless: false,
    },
    include: ['**/*.test.{ts,js}'],
}
```

### DIFF
--- a/packages/stage-layouts/src/components/Layouts/HeaderLink.vue
+++ b/packages/stage-layouts/src/components/Layouts/HeaderLink.vue
@@ -10,8 +10,8 @@ const { isDark: dark } = useTheme()
 
 <template>
   <RouterLink
-    to="/" flex="~" items-center
-    gap-2 px-2 text-nowrap text-2xl outline-none
+    to="/" flex="~"
+    w-max items-center gap-2 px-2 text-nowrap text-2xl outline-none
   >
     <template v-if="dark">
       <img :src="LogoDark" h-8 w-8 class="theme-colored">


Zazzik1:
Checked - MobileHeaderLink.vue does not appear to be impacted by this bug

Issue #1058, opened 2026-03-01
Feature Request: Allow manual model name input for OpenAI-compatible providers


willwang0202:
### Describe the bug

**Describe the problem**
When using the OpenAI Compatible provider, AIRI tries to auto-fetch the model list via `/v1/models`. However, some providers (e.g. MiniMax at `api.minimax.io`) do not expose this endpoint, resulting in a `404` error under Modules → Consciousness → Model, making it impossible to select a model.

**Expected behavior**
If the model list fails to load, there should be a fallback option to manually type in a model name/ID — similar to how many other AI tools handle this.

**Steps to reproduce**
1. Go to Settings → Providers → Add OpenAI Compatible
2. Set Base URL to `https://api.minimax.io/v1` with a valid MiniMax API key
3. Go to Modules → Consciousness
4. Select the OpenAI Compatible provider
5. Model section shows: `Error loading models — Remote sent 404 response: 404 page not found`
6. There is no way to manually enter a model name

**Suggested fix**
Add a text input field as a fallback when the `/v1/models` endpoint returns an error, so users can manually type their desired model ID (e.g. `MiniMax-M2.5`).

### System Info

```Shell
- AIRI version: 0.8.4
- OS: macOS (Apple Silicon)
- Provider: MiniMax (`https://api.minimax.io/v1`)
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

NJX-njx:
I'd like to work on this. The plan is to add a fallback text input field for manual model name/ID entry when the `/v1/models` endpoint returns an error or is unavailable. This will involve modifying the provider configuration UI to detect fetch failures and show the manual input option.

PR #1065, opened 2026-03-01
fix(onboarding): allow manual model entry when list is empty

liuxiaopai-ai:
## Summary
- add manual model input fallback in onboarding model-selection step when provider model list is empty
- keep existing model list picker unchanged when models are available
- reuse existing model placeholder i18n key for consistent copy

Related to #1058

## Testing
- `pnpm exec moeru-lint --fix packages/stage-ui/src/components/scenarios/dialogs/onboarding/step-model-selection.vue`
- `pnpm test:run -- packages/stage-ui/src/stores/provider-catalog.test.ts`


nekomeowww:
Rebase is needed.

PR #1076, opened 2026-03-02
refactor: replace unsafe any types with unknown

fvngs:
## Description

- replaces unsafe `any` types with `unknown` in 12 files where it is safe to do so
- targets type guard parameters, pass-through return types, and fields that are never
accessed for properties
- leaves all `as any` casts, game-engine objects, zod internals, and event handler
patterns unchanged

## Additional Context

  - `services/telegram-bot` - return type and json-serialized params
  - `services/satori-bot` - request body and interface fields
  - `services/minecraft` - type guard params with proper `in (x as object)` narrowing
  - `packages/stream-kit` - drain task `promise<void>`
  - `packages/tauri-plugin-mcp` - schema default field
  - `packages/stage-ui` - vad debug event, voice key type, validator helper
  - `docs/scripts/avif.ts` - transform return type


nekomeowww (approved):

shinohara-rin:
ci failing

nekomeowww (approved):

Issue #1087, opened 2026-03-03
[Feat]: Reset Main window option


kiminus:
### Clear and concise description of the problem

windows: add system tray option to force reset the window. This is particularly useful when the Tamagotchi main window gets stuck off-screen or out of bounds after resizing and moving

### Suggested solution / Ideas

<img width="222" height="210" alt="Image" src="https://github.com/user-attachments/assets/8275788b-199e-47c7-ae2d-01031d62c535" />

tested on windows only. 

### Additional context

I have created the branch here: [feat/reset-main-window-position](https://github.com/moeru-ai/airi/compare/main...kiminus:feat/reset-main-window-position?expand=1)

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

nekomeowww:
Implemented.

PR #1088, opened 2026-03-03
feat(tamagotchi): add reset main window position to system tray

kiminus:
## Description

sometimes the window gets stuck and you cannot click on the move button to drag the window, so it might be beneficial to add a system tray option to force reset the window when stuck. 

## Linked Issues

https://github.com/moeru-ai/airi/issues/1087

### DIFF
--- a/apps/stage-tamagotchi/src/main/tray/index.ts
+++ b/apps/stage-tamagotchi/src/main/tray/index.ts
@@ -37,7 +37,7 @@ export function setupTray(params: {
       { label: 'Show', click: () => toggleWindowShow(params.mainWindow) },
       { type: 'separator' },
       {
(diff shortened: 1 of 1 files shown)


apps/stage-tamagotchi/src/main/tray/index.ts
```diff
@@ -36,6 +36,15 @@ export function setupTray(params: {
     const contextMenu = Menu.buildFromTemplate([
       { label: 'Show', click: () => toggleWindowShow(params.mainWindow) },
       { type: 'separator' },
+      {
+        label: 'Reset main window',
```
sumimakito:
```suggestion
        label: 'Center and Reset Size',
```

PR #1096, opened 2026-03-04
feat(stage-ui): optimize card many select component layout styles

horizonzzzz:
## Description

- Add `columns` prop to support configurable column count
- Unify column width calculation for both collapsed and expanded states
- Fix Y-axis scroll snap behavior in expanded mode
- Add visibility control for the expand/collapse button
- Fix model bindings in voice-card-many-select story
- Remove reordering in the consciousness model selector to prevent layout shift in collapsed state
- Add explicit UI text to indicate the currently selected model

### Before

ScreenShot_2026-03-04_171417_890

### After

ScreenShot_2026-03-04_170513_762


horizonzzzz:
Personally I feel this layout looks cleaner, but it’s not a required change. Totally fine to keep the current version.

PR #1100, opened 2026-03-04
fix(providers): use native ElevenLabs API for TTS, bypass unspeech proxy 401 on free-tier users

Hanfeng-Lin:
## Description

ElevenLabs TTS synthesis was routed through the public `unspeech.hyp3r.link` proxy, which triggers ElevenLabs' Free Tier abuse detection because many users share the same proxy IP, resulting in HTTP 401 errors for free-tier users:

```
Remote sent 401 response: {"detail": {"status": "detected_unusual_activity",
"message": "Unusual activity detected. Free Tier usage disabled..."}}
```

## Root Cause

`createProvider` used `createUnElevenLabs(apiKey, baseUrl)` from the `unspeech` SDK, which routes all TTS requests through an unspeech proxy instance. When using the default public instance (`unspeech.hyp3r.link`), ElevenLabs sees many requests originating from a single shared IP and flags it as abuse, blocking free-tier users entirely.

## Fix

Replace `createUnElevenLabs` with a native provider that calls the ElevenLabs API directly.

`@xsai/generate-speech` supports an `options.fetch` override. We use this to intercept the xsai OpenAI-format request and transform it into ElevenLabs' native `POST /v1/text-to-speech/{voice_id}` format before it leaves the app — no proxy involved.

This is consistent with how other providers in this file are implemented:
- `openai-audio-speech` → uses `createOpenAI` directly (no proxy)
- `index-tts-vllm` → custom fetch, direct REST call

### DIFF
--- a/packages/stage-ui/src/stores/providers.ts
+++ b/packages/stage-ui/src/stores/providers.ts
@@ -861,7 +861,7 @@ export const useProvidersStore = defineStore('providers', () => {
       description: 'elevenlabs.io',
       icon: 'i-simple-icons:elevenlabs',
       defaultOptions: () => ({
-        baseUrl: 'https://unspeech.hyp3r.link/v1/',
+        baseUrl: 'https://api.elevenlabs.io/v1/',
         voiceSettings: {
           similarityBoost: 0.75,
           stability: 0.5,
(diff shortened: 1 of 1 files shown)


nekomeowww:
ElevenLabs API doesn't support requesting from frontend directly, this should not be implemented.

PR #1101, opened 2026-03-04
fix: fallback to in-window settings route when IPC open fails

shesl-meow:
## Summary
- add fallback when opening settings via IPC fails
- route to /settings in current window to avoid no-response behavior

## Test


nekomeowww:
This should not be implemented, why would IPC fail?

PR #1107, opened 2026-03-04
fix(providers): use native ElevenLabs API on desktop to avoid unspeech proxy 401

Hanfeng-Lin:
This is a follow-up to the previous PR #1100 , addressing the rejection reason.

## Problem

The previous PR called ElevenLabs API directly from all platforms, which broke the **web version** due to CORS — browsers block cross-origin requests to `api.elevenlabs.io` since it doesn't return `Access-Control-Allow-Origin` headers.

The underlying issue remains: the public `unspeech.hyp3r.link` proxy causes HTTP 401 errors for Free Tier users because many users share the same proxy IP, triggering ElevenLabs' abuse detection.

## Solution

Use `isStageTamagotchi()` to branch by platform:

- **Desktop**: Electron's renderer is not subject to browser CORS rules, so direct cross-origin fetch to `api.elevenlabs.io` works, avoiding the shared-IP 401 issue.
- **Web**: Keeps routing through the configured unspeech proxy — required for CORS compliance.


nekomeowww:
But the request still happens right inside of the browser - provided by Electrom, which is based on Chromium, the CORS issue still exists...

Hanfeng-Lin:
The desktop version has already been tested and confirmed working without CORS issue. The network requests reach api.elevenlabs.io and receive valid responses, which would be impossible if browser-level CORS were enforced. The web end was also tested and the proxy is working as normal. 
I think Electron has more relaxed security policies which could bypass the CORS restrictions, either by `webPreferences: { webSecurity: false }` in BrowserWindow or `session.webRequest.onHeadersReceived` to inject CORS headers into responses

shinohara-rin:
$ curl -sI -X OPTIONS https://api.elevenlabs.io/v1/text-to-speech/test \
  -H "Origin: http://localhost" \
  -H "Access-Control-Request-Method: POST"

access-control-allow-origin: *
access-control-allow-headers: *




strange things, looks like elevenlabs return CORS headers as of now(or is it just me???)

Hanfeng-Lin:
Thanks @shinohara-rin for pointing out! I've updated the branch to bypass the unspeech proxy and call ElevenLabs' native API directly across all platforms. Previously this was Desktop-only due to CORS issues, but ElevenLabs now returns 'Access-Control-Allow-Origin: *'. 

Both platforms were tested successfully

shinohara-rin (commented):
It looks a bit dirty to me to handcraft an http client here, this might only work as a temporary fix.

shall we consider supporting the elevenlabs native api? @nekomeowww 
with CORS out of the way, we still have the anti-abuse problem to deal with

Hanfeng-Lin:
If we're sticking with the native ElevenLabs API (because this solves the 401 redirect issue caused by shared IPs once and for all), I can extract the dirty work related to fetching and create a dedicated adapter file in src/stores/providers/elevenlabs/native.ts, exposing a clean createNativeElevenLabsProvider(). This way, providers.ts can now function like before, simply calling a single line of code.

PR #1124, opened 2026-03-05
fix(stage-tamagotchi): guard stdout/stderr against EPIPE

cat1949:
## Summary
Fix Electron main-process crash caused by broken output streams (EPIPE / ERR_STREAM_DESTROYED) when console streams are detached or closed.

## Changes
- Add installStreamErrorGuards() in apps/stage-tamagotchi/src/main/index.ts
- Register guarded error listeners on process.stdout and process.stderr
- Initialize stream guards early during main process bootstrap

## Testing
- Start desktop app with pnpm dev:tamagotchi
- Verify app no longer crashes on broken-pipe style stream errors


nekomeowww:
In what cases this will happen? How to reproduce this?

cat1949:
> 这种情况会在哪些情况下发生？如何重现这种情况？-----------谷歌翻译----------什么情况下会发生这种情况？如何重置这个？
刚部署在打开设置界面进入二级界面时候会出现报错，最后进行修复后就正常了
屏幕截图 2026-03-05 062709

cat1949:
Fixed in `ac6482b7`.

The issue occurred right after deployment: when opening the Settings page and then entering a second-level settings page, the app would throw this error. After this fix, it works normally again.

I also updated the stream error guard so that it only ignores `EPIPE` / `ERR_STREAM_DESTROYED` and rethrows unexpected stream errors.

nekomeowww:
Root cause wasn't this, it should be fixed.

PR #1125, opened 2026-03-05
feat(providers): manual model ping and selective validation checks

cat1949:
## Summary
Improve provider configuration UX and safety for OpenAI-compatible chat providers:
- avoid automatic chat ping unless model is explicitly selected
- allow manual model ping from provider settings

## Changes
- Add model select/manual input in provider settings page
- Add manual validation trigger (runValidationNow) and UI button (??????)
- Update OpenAI-compatible validators:
  - connectivity/model list checks use model-list request path
  - chat-completions check runs only when config.model is set

## Why
Prevents accidental, uncontrolled chat ping behavior during provider setup and reduces risk of rate-limit spikes.

## Testing
- pnpm -rF @proj-airi/stage-ui typecheck
- pnpm -rF @proj-airi/stage-pages typecheck


nekomeowww:
Could you please include screenshots?

cat1949:
> Could you please include screenshots?-----------谷歌翻译-----------您能附上屏幕截图吗？

当时没有进行截图，具体情景就是。当我输入完接口和key过后，立马跳出达到速率限制了，我这个速率限制是5 rpm

cat1949:
To clarify my earlier reply in English: I did not take screenshots at the time.

The exact scenario was that after I entered the endpoint URL and API key, the app immediately hit the rate limit. My configured rate limit was 5 RPM.

This PR is intended to avoid that automatic validation / ping behavior until a model is explicitly selected, and to allow validation to be triggered manually.

cat1949:
Addressed in `3d6f5a95`.

Changes made:
- guard provider access with provider metadata checks so invalid route keys like `__proto__` are not used to read/write provider config or model-loading state
- centralize provider field updates in the shared validation composable to reduce duplicated setter logic
- replace the hardcoded validation button text with i18n keys

Validation:
- `pnpm -rF @proj-airi/stage-ui typecheck`
- `pnpm -rF @proj-airi/stage-pages typecheck`
- `pnpm typecheck`

PR #1148, opened 2026-03-05
feat(ui,stage-ui,stage-pages,i18n): transcription confidence filter

### LINKED ISSUE #871: feature request: Disregard transcription with low confidence
### Clear and concise description of the problem

Transcription models like `whisper-large-v3` can hallucinate a lot. Although we have a local VAD to gate transcriptions, some noises would still get through. In those cases, the transcription service would tend to produce gibberish outputs.

### Suggested solution / Ideas

The returned `avg_logprob` from the API response should be a fairly good indicator of potential hallucinations; a reliable prediction generally has a logprob higher than `-1`, thus, we could drop transcriptions with lower confidence.

### Alternative

Instead of relying on the `avg_logprob` returned by providers (which might not be universal), we could also try to use a small, distilled whisper solely for gating transcription API calls.

### Additional context

_No response_

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Reisenbug:
## Description

Filter out low-confidence transcription segments based on `avg_logprob` to reduce Whisper hallucinations caused by background noise passing through VAD.

A new Confidence Threshold slider is added to Settings → Modules → Hearing.

Only effective for providers supporting the Whisper API (e.g., OpenAI, Groq). Streaming transcription (Aliyun NLS, Web Speech API) is not affected.

Also fixes FieldRange, || to ??

## Linked Issues

Fixes #871 

## Additional Context

Filtering is segment-level, not result-level

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
+++ b/apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
@@ -119,7 +119,7 @@ function refreshWindow() {
       >
         <div v-if="expanded" border="1 neutral-200 dark:neutral-800" mb-2 flex flex-col gap-1 rounded-2xl p-2 backdrop-blur-xl class="bg-neutral-100/80 shadow-2xl shadow-black/20 dark:bg-neutral-900/80">
           <div grid grid-cols-3 gap-2>
-            <ControlButtonTooltip>
+            <ControlButtonTooltip disable-hoverable-content>
               <ControlButton :button-style="adjustStyleClasses.button" @click="openSettings">
                 <div i-solar:settings-minimalistic-outline :class="adjustStyleClasses.icon" text="neutral-800 dark:neutral-300" />
               </ControlButton>
@@ -128,7 +128,7 @@ function refreshWindow() {
               </template>
             </ControlButtonTooltip>
 
-            <ControlButtonTooltip>
+            <ControlButtonTooltip disable-hoverable-content>
               <ControlButton :button-style="adjustStyleClasses.button" @click="openChat">
                 <div i-solar:chat-line-line-duotone :class="adjustStyleClasses.icon" text="neutral-800 dark:neutral-300" />
               </ControlButton>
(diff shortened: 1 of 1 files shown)


Reisenbug:
> Thanks for the PR, however, a few things need to be addressed:
> 
> * Magic number -3 duplicated, should be a constant
> * Don't display the slider for streaming providers where it has no effect

Solved.
Extracted -3 as `CONFIDENCE_THRESHOLD_DISABLED` constant
Hidden the slider when the active provider supports streaming input

shinohara-rin:
Have you tested this on providers that do not support verbose json?

This might cause confusion if used on said providers without clear error surfacing and user guidance?

Reisenbug:
> Have you tested this on providers that do not support verbose json?
> 
> This might cause confusion if used on said providers without clear error surfacing and user guidance?

 Now when a provider does not return verbose_json segments, a warning is surfaced in the UI to inform the user that confidence filtering had no effect on the last transcription
Also added a note when Confidence Threshold is enabled: If your provider does not support verbose_json responses, this setting will have no effect

packages/stage-ui/src/stores/modules/hearing.ts
```diff
@@ -49,6 +49,15 @@ interface HearingTranscriptionInvokeOptions {
   providerOptions?: Record<string, unknown>
 }
 
+export const CONFIDENCE_THRESHOLD_DISABLED = -3
+
+export function filterTranscriptionByConfidence(
+  segments: Array<{ text: string, avg_logprob: number }>,
```
nekomeowww:
```suggestion
  segments: Array<{ text?: string, avg_logprob?: number }>,
```

packages/stage-ui/src/stores/modules/hearing.ts
```diff
@@ -49,6 +49,15 @@ interface HearingTranscriptionInvokeOptions {
...
+  threshold: number,
+): string {
+  return segments.filter(s => s.avg_logprob >= threshold).map(s => s.text).join('').trim()
```
nekomeowww:
```suggestion
  // if provider side doesn't implement this correctly
  if (!segments.some(s => s?.avg_logprob) || !segments.some(s => s?.text)) {
    return segments
  }
  
  return segments.filter(s => s?.avg_logprob >= threshold).map(s => s?.text).join('').trim()
```

PR #1151, opened 2026-03-05
fix(stage-ui,stage-web): TTS should play on iOS even when silent mode is on

### LINKED ISSUE #894: bug: cannot hear playback of text-to-speech audio on PWA
### Describe the bug

For iOS devices, if Silent mode toggled on, websites & PWA audio will be played but no volume.

### System Info

```Shell
Mobile. iOS 26.1
```

### Validations

### Contributions

NJX-njx:
## Summary

Fixes TTS audio not playing on iOS when the mute/silent switch is on. Addresses #894.

## Root Cause

On iOS, the hardware mute switch causes all Web Audio API output to be silent because `AudioContext` is treated as the "ambient" audio session category. This is a known WebKit limitation:
- https://bugs.webkit.org/show_bug.cgi?id=237322
- https://stackoverflow.com/questions/21122418/ios-webaudio-only-works-on-headphones

## Solution

The standard workaround (used by YouTube, Spotify Web, and other production apps) is to play a silent HTML5 `<audio>` element alongside the Web Audio API. This forces iOS to assign the "playback" audio session category, which is **not** affected by the mute switch.

### Changes

- **`packages/stage-ui/src/utils/ios-audio-unmute.ts`** (new):
  - `enableIOSPlayback()`: Creates and loops a tiny silent WAV via HTML5 `<audio>`, forcing iOS into "playback" mode
  - `disposeIOSPlayback()`: Cleanup function
  - `isIOS()`: Platform detection (includes iPad with desktop UA)
  - No-ops safely on non-iOS platforms

- **`packages/stage-ui/src/stores/audio.ts`**:
  - Registers one-time `touchend`/`click` document listener that calls `enableIOSPlayback()` on first user interaction
  - User gesture is required for iOS to allow `<audio>.play()`

## Testing Notes

- Safe no-op on non-iOS platforms (Android, desktop browsers)
- The silent WAV is a data URI (44 bytes), no network request needed
- Only one silent `<audio>` element is created (singleton pattern)

Closes #894


packages/stage-ui/src/utils/ios-audio-unmute.ts
```diff
@@ -0,0 +1,58 @@
...
+ * - https://github.com/nickvdp/unmute
+ * - https://bugs.webkit.org/show_bug.cgi?id=237322
+ */
+
+let silentAudio: HTMLAudioElement | undefined
+let disposed = false
+
+// Tiny silent WAV (44 bytes header + 1 sample of silence), base64-encoded
+const SILENT_WAV_DATA_URI
+  = 'data:audio/wav;base64,UklGRiQAAABXQVZFZm10IBAAAAABAAEARKwAAIhYAQACABAAZGF0YQAAAAA='
```
nekomeowww, line 20: /gemini review

Is that possible to avoid using pre-defined base64 payload? I am not sure whether this will be harmful...

packages/stage-ui/src/utils/ios-audio-unmute.ts
```diff
@@ -0,0 +1,58 @@
...
+
+function isIOS(): boolean {
+  if (typeof navigator === 'undefined')
+    return false
+  return /iPad|iPhone|iPod/.test(navigator.userAgent)
+    || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1)
+}
```
nekomeowww, line 27: /gemini review

Must we use such method to detect iOS?

packages/stage-ui/src/utils/ios-audio-unmute.ts
nekomeowww: Can't we dynamically generate it? I think the encoded base64 will have issues when againsting cyber security issues, we should never do this, should we?

packages/stage-ui/src/utils/ios-audio-unmute.ts
nekomeowww: Should we use https://www.npmjs.com/package/ua-parser-js?

PR #1170, opened 2026-03-06
chore(docs): reorganize content

### LINKED ISSUE #1158: Feature Request: Add GUI entry point for minecraft
# Core Request
We need either an **actual GUI entry point** OR **a clear declaration/notice in the GUI and docs** for the minecraft module.

Currently, it can only be run directly from the source code, which is not very user-friendly for everyone.

Additionally, it would be great if we could have a unified GUI entry point for other platform adapters like Discord, Telegram, and Satori in the future.

NashChennc:
# Description
* reorganized existing info while preserving the original wording; added more entries.
* Added a feature availability statement.
* Proofread Chinese, Japanese, and English versions.

Closes #1158

### DIFF
--- a/docs/content/zh-Hans/docs/chronicles/version-v0.0.1/index.md
+++ b/docs/content/zh-Hans/docs/chronicles/version-v0.0.1/index.md
@@ -9,7 +9,7 @@ title: 编年史 v0.0.1
   - [x] 模型下载
     - [x] 桃濑日和（Neuro 第一版本的模型）Pro 版本（中小企业可免费商用）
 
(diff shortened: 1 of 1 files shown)


NashChennc:
Removing `(Desktop)` from the frontmatter description immediately fixed the startup crash. It seems the vite-frontmatter-assets plugin mistakenly parsed this plain text as a local file path and threw an ENOENT error. Such a weird bug.

NashChennc:
I'm not too familiar with frontend stuff, so I might need some help when updating the docs & gui in the future.

PR #1187, opened 2026-03-07
fix(server-runtime): use superjson.parse for incoming WebSocket messages

SakataRyuichi:
## Summary

- **Root cause**: `server-runtime` was deserializing incoming WebSocket messages using `message.json()` (standard `JSON.parse`), but SDK clients (`@proj-airi/server-sdk`) serialize all outgoing events with `superjson.stringify`.
- This caused every event from SDK clients to be parsed as the raw superjson wrapper `{ json: {...}, meta: {...} }`, making `event.type` always `undefined` and silently breaking all event routing (`input:text`, `module:announce`, etc.).
- Any service using the SDK client (e.g. `discord-bot`, `minecraft-bot`) would receive no responses because the server never dispatched their events to other peers.

## Changes

- `packages/server-runtime/src/index.ts`: Import `parse` from `superjson` and replace `message.json()` with `parse<WebSocketEvent>(message.text())`. Added a `NOTICE` comment explaining the reason.
- `packages/stage-ui/src/stores/mods/api/context-bridge.ts`: Add `try/catch` around `getProviderInstance` to surface provider lookup errors instead of silently swallowing them. Also add an early return guard when `activeProvider` or `activeModel` is not configured.
- `services/discord-bot/src/adapters/airi-adapter.ts`: Minor: add `try/catch` in the `output:gen-ai:chat:message` handler (was already partially guarded).

## How to reproduce (before fix)

1. Start `server-runtime` and connect a `discord-bot` (or any SDK client).
2. Send a message from Discord — the bot receives the message and calls `airiClient.send({ type: 'input:text', ... })`.
3. `server-runtime` parses the message with `JSON.parse`, gets `{ json: { type: 'input:text', ... }, meta: ... }`.
4. `switch (event.type)` sees `undefined`, falls through to the broadcast block, which then also reads wrong data.
5. No event ever reaches the renderer — no LLM response is generated, Discord gets no reply.

## Test plan

- [x] Start `pnpm dev:tamagotchi` (includes server-runtime)
- [x] Start `pnpm -F @proj-airi/discord-bot start` with valid `DISCORD_TOKEN`
- [x] Send a message in Discord — bot now correctly receives and replies
- [x] Desktop app chat (direct) continues to work as before

### DIFF
--- a/packages/stage-ui/src/stores/mods/api/context-bridge.ts
+++ b/packages/stage-ui/src/stores/mods/api/context-bridge.ts
@@ -88,9 +88,6 @@ export const useContextBridgeStore = defineStore('mods:api:context-bridge', () =
           }
         }
 
-        if (!activeProvider.value || !activeModel.value)
-          return
-
         if (activeProvider.value && activeModel.value) {
           let chatProvider: ChatProvider
           try {


packages/stage-ui/src/stores/mods/api/context-bridge.ts
```diff
@@ -88,8 +88,18 @@ export const useContextBridgeStore = defineStore('mods:api:context-bridge', () =
           }
         }
 
+        if (!activeProvider.value || !activeModel.value)
+          return
+
```
nekomeowww:
```suggestion
```

Issue #1173, opened 2026-03-06
refactor/fix(stage-ui-three): VRM performance issue


Lilia-Chen:
### Describe the bug

This is the issue for tracking all the related bugs and issues:
- VRM performance, overloading GPU and CPU
- Memory leak, incorrect dispose or none-dispose
- Memory overloading
- etc...

### System Info

```Shell
desktop & web
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [x] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

Lilia-Chen:
<img width="899" height="119" alt="Image" src="https://github.com/user-attachments/assets/54da9648-cd59-42ef-833e-fabaa700d271" />

Need to investigate the lip-sync & expression pipeline. Seems like it's broken.

Lilia-Chen:
https://github.com/moeru-ai/airi/pull/1194

Cleared some memory leaks and lifecycle issues

PR #1194, opened 2026-03-08
refactor: vrm performance tracing and better lifecycle management

Lilia-Chen:
### This PR redesigns the stage-three runtime so VRM state is owned by the scene runtime, not by incidental Vue component lifetime.

- The new lifecycle model separates **load**, **commit**, and **destroy**. A replacement VRM is now prepared off-scene, validated, and only committed after it is ready. This gives the runtime a clearer contract:
  - one committed VRM per scene
  - explicit model replacement
  - routine remounts do not automatically imply full destruction
- Cleanup is now reason-aware instead of unconditional. A real model switch still goes through a full destroy path. Ordinary unmount and hot-reload paths no longer eagerly deep-dispose the active VRM graph.
- To support reuse without introducing scene corruption, this PR adds a detached-instance cache for VRM reuse. That cache is intentionally local in scope: it is keyed by page context and reused instances are validated before they are recommitted. This matters because the main stage and the settings preview may render the same `modelSrc` while still belonging to different route and scene lifecycles. Reuse is therefore defensive and local, not global and optimistic.

### This PR also folds runtime diagnostics back into `@proj-airi/stage-ui-three`.
- Its responsibilities now live in a local `trace` submodule inside `stage-ui-three`, which keeps trace ownership next to the runtime that emits it. That trace layer now covers:
  - renderer info snapshots
  - hit-test readback diagnostics
  - VRM frame update breakdowns
  - VRM load / dispose lifecycle events
  - renderer memory and scene summary snapshots
- Desktop integration now builds on top of that local trace contract instead of depending on a separate runtime package. New renderer-side diagnostics stores and Eventa bridges in `stage-tamagotchi` expose a more coherent path for lag visualization, lifecycle inspection, and cross-window runtime diagnostics.

### Runtime coordination around renderer switching is also tightened.
- Live2D and VRM invalidation now use separate store channels instead of sharing a broadcast path. Window lifecycle handling is now modeled explicitly, so pause behavior can be derived from visibility and minimization state instead of leaking through renderer teardown.

## Linked Issues

https://github.com/moeru-ai/airi/issues/1173

## Additional Context

N/A


packages/stage-ui-three/src/components/Model/vrm-instance-cache.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export function stashManagedVrmInstance(scopeKey: string, instance: ManagedVrmInstance) {
+  const previous = managedVrmCacheState.detachedByScope[scopeKey]
+  managedVrmCacheState.detachedByScope[scopeKey] = instance
+
+  if (previous === instance)
+    return undefined
+
+  return previous ?? undefined
```
sumimakito: Because `previous` is `ManagedVrmInstance | undefined`, this can be simplified as:

```suggestion
  return previous === instance ? undefined : previous
```

packages/stage-ui-three/src/components/Model/vrm-instance-cache.ts
```diff
@@ -0,0 +1,49 @@
...
+export function takeManagedVrmInstance(scopeKey: string, modelSrc: string) {
+  const cached = managedVrmCacheState.detachedByScope[scopeKey]
+  if (!cached || cached.modelSrc !== modelSrc)
+    return undefined
+
+  delete managedVrmCacheState.detachedByScope[scopeKey]
+  return cached
+}
+
+export function stashManagedVrmInstance(scopeKey: string, instance: ManagedVrmInstance) {
```
sumimakito: If we didn't intend to update the `scopeKey` of a `ManagedVrmInstance`, the `scopeKey` parameter might be redundant here, as `ManagedVrmInstance` already contains one:
 
```suggestion
export function stashManagedVrmInstance(instance: ManagedVrmInstance) {
  const { scopeKey } = instance
```

If we intended to update the `scopeKey` (i.e., "move" the scope from one to another), we should also update the `scopeKey` in `ManagedVrmInstance`. Otherwise, we may end up with having `detachedByScope[scopeKey]` pointing to a `ManagedVrmInstance` with a different `scopeKey`.

packages/stage-ui-three/src/components/Model/VRMModel.vue
```diff
@@ -174,47 +195,286 @@ const vrmLipSync = useVRMLipSync(currentAudioSource)
...
+  if (activeInstance)
+    detachVrmGroup(activeInstance.group)
+
+  if (shouldDestroyResources) {
+    destroyManagedVrmInstance(activeInstance)
+    destroyManagedVrmInstance(clearManagedVrmInstance(getManagedVrmScopeKey()))
   }
-  // clear IBL probe
+  else if (shouldStashVrmResources(reason)) {
+    destroyManagedVrmInstance(activeInstance ? stashManagedVrmInstance(getManagedVrmScopeKey(), activeInstance) : undefined)
```
sumimakito: If we accept the simplification, the `getManagedVrmScopeKey()` call here can be removed:

```suggestion
    destroyManagedVrmInstance(activeInstance ? stashManagedVrmInstance(activeInstance) : undefined)
```

packages/stage-ui-three/src/components/Model/VRMModel.vue
```diff
@@ -253,203 +513,287 @@ function defaultTookAt(eyeHeight: number): Vec3 {
...
+      else {
+        if (!isLoadRequestCurrent(requestId)) {
+          destroyManagedVrmInstance(stashManagedVrmInstance(getManagedVrmScopeKey(), reusableInstance))
+          return
+        }
+
+        if (!airiIblProbe && scene.value)
+          airiIblProbe = createIblProbeController(scene.value)
+
+        commitManagedVrmInstance(reusableInstance)
```
sumimakito, line 616: Do we need to `didCommitLoad = true` in this "reusable" code path?

packages/stage-ui/src/composables/use-data-maintenance.ts
```diff
@@ -39,6 +39,14 @@ export function useDataMaintenance() {
...
 
   async function deleteAllModels() {
+    // if (import.meta.env.DEV) {
+    //   console.warn('[StageModel][set from data-maintenance-delete-models]', {
+    //     currentHref: typeof window !== 'undefined' ? window.location.href : 'unknown',
+    //     nextModelId: 'preset-live2d-1',
+    //     stack: new Error('[StageModel][data-maintenance-delete-models]').stack,
+    //   })
+    // }
+
```
nekomeowww: remove

packages/stage-ui/src/stores/settings/stage-model.ts
```diff
@@ -1,30 +1,96 @@
...
+//     href: event.href,
+//     instanceId: event.instanceId,
+//     nextValue: event.nextValue,
+//     previousValue: event.previousValue,
+//     sentAt: event.sentAt,
+//     stack: event.stack,
+//     type: event.type,
+//   }
+// }
+
```
nekomeowww: remove

packages/stage-ui/src/stores/settings/stage-model.ts
```diff
@@ -68,11 +139,51 @@ export const useSettingsStageModel = defineStore('settings-stage-model', () => {
...
+  //     currentHref: typeof window !== 'undefined' ? window.location.href : 'unknown',
+  //     senderHref: event.href,
+  //     senderInstanceId: event.instanceId,
+  //     nextValue: event.nextValue,
+  //     previousValue: event.previousValue,
+  //     sentAt: event.sentAt,
+  //   })
+  //   if (event.stack)
+  //     console.warn(`[StageModel][set][recv][stack]\n${event.stack}`)
+  // })
```
nekomeowww: remove

packages/stage-ui/src/stores/settings/stage-model.ts
```diff
@@ -68,11 +139,51 @@ export const useSettingsStageModel = defineStore('settings-stage-model', () => {
...
+
+  watch(stageModelSelectedState, (_newValue, _oldValue) => {
+    // if (import.meta.env.DEV) {
+    //   console.warn('[StageModel][selected]', {
+    //     currentHref: typeof window !== 'undefined' ? window.location.href : 'unknown',
+    //     newValue: _newValue,
+    //     oldValue: _oldValue,
+    //   })
+    // }
```
nekomeowww: remove

packages/stage-ui/src/components/scenarios/settings/model-settings/index.vue
```diff
@@ -46,21 +46,43 @@ const {
...
+
+async function handleModelPick(selectedModel: DisplayModel | undefined) {
+  if (!modelSelectorOpen.value) {
+    // if (import.meta.env.DEV) {
+    //   console.warn('[StageModel][ignore pick while dialog closed]', {
+    //     currentHref: typeof window !== 'undefined' ? window.location.href : 'unknown',
+    //     nextModelFormat: selectedModel?.format,
+    //     nextModelId: selectedModel?.id ?? '',
+    //   })
+    // }
```
nekomeowww: remove

packages/stage-ui/src/components/scenarios/settings/model-settings/index.vue
```diff
@@ -46,21 +46,43 @@ const {
...
+  }
+
+  // if (import.meta.env.DEV) {
+  //   console.warn('[StageModel][set from model-settings]', {
+  //     currentHref: typeof window !== 'undefined' ? window.location.href : 'unknown',
+  //     nextModelFormat: selectedModel?.format,
+  //     nextModelId: selectedModel?.id ?? '',
+  //     stack: new Error('[StageModel][model-settings]').stack,
+  //   })
+  // }
```
nekomeowww: remove

packages/stage-ui-three/src/stores/model-store.ts
```diff
@@ -70,19 +79,39 @@ export const useModelStore = defineStore('modelStore', () => {
...
   watch(data, (event) => {
-    if (event.type === 'should-update-view') {
+    if (event?.type === 'vrm-should-update-view') {
+      // if (import.meta.env.DEV) {
+      //   console.warn('[VRM shouldUpdateView][recv]', {
+      //     currentHref: typeof window !== 'undefined' ? window.location.href : 'unknown',
+      //     event,
+      //   })
+      // }
+
```
nekomeowww: remove

.vscode/settings.json
Lilia-Chen, line 1: Restore this change since it has nothing to do with this PR.

.vscode/settings.json
```diff
@@ -66,7 +66,7 @@
     "editor.defaultFormatter": "rust-lang.rust-analyzer",
     "editor.formatOnSave": true
   },
-  "rust-analyzer.checkOnSave": true,
+  "rust-analyzer.checkOnSave": false,
```
Lilia-Chen:
```suggestion
  "rust-analyzer.checkOnSave": true,
```

apps/stage-tamagotchi/src/main/services/electron/window.ts
```diff
@@ -1,15 +1,37 @@
...
 import { createRendererLoop } from '@proj-airi/electron-vueuse/main'
 
-import { electron, electronWindowClose } from '../../../shared/eventa'
+import {
+  electron,
+  electronGetWindowLifecycleState,
+  electronWindowClose,
+  electronWindowLifecycleChanged,
+
+} from '../../../shared/eventa'
```
Lilia-Chen, line 16:
```suggestion
import {
  electron,
  electronGetWindowLifecycleState,
  electronWindowClose,
  electronWindowLifecycleChanged,
} from '../../../shared/eventa'
```

apps/stage-tamagotchi/src/renderer/bridges/stage-three-runtime-trace.ts
```diff
@@ -0,0 +1,143 @@
...
+      continue
+
+    context.emit(stageThreeRuntimeTraceForwardedEvent, {
+      envelope,
+      origin: instanceId,
+    })
+  }
+}
+
+function forwardTraceEvent<T>(eventa: Eventa<T>, createEnvelope: (payload: T) => StageThreeRuntimeTraceEnvelope) {
```
Lilia-Chen:
```suggestion
function subscribeTraceEvent<T>(eventa: Eventa<T>, createEnvelope: (payload: T) => StageThreeRuntimeTraceEnvelope) {
```

apps/stage-tamagotchi/src/renderer/bridges/stage-three-runtime-trace.ts
```diff
@@ -0,0 +1,143 @@
...
+  })
+
+  forwardTraceEvent(stageThreeTraceRenderInfoEvent, payload => ({ type: 'three-render-info', payload }))
+  forwardTraceEvent(stageThreeTraceHitTestReadEvent, payload => ({ type: 'three-hit-test-read', payload }))
+  forwardTraceEvent(stageThreeTraceVrmUpdateFrameEvent, payload => ({ type: 'vrm-update-frame', payload }))
+  forwardTraceEvent(stageThreeTraceVrmLoadStartEvent, payload => ({ type: 'vrm-load-start', payload }))
+  forwardTraceEvent(stageThreeTraceVrmLoadEndEvent, payload => ({ type: 'vrm-load-end', payload }))
+  forwardTraceEvent(stageThreeTraceVrmLoadErrorEvent, payload => ({ type: 'vrm-load-error', payload }))
+  forwardTraceEvent(stageThreeTraceVrmDisposeStartEvent, payload => ({ type: 'vrm-dispose-start', payload }))
+  forwardTraceEvent(stageThreeTraceVrmDisposeEndEvent, payload => ({ type: 'vrm-dispose-end', payload }))
```
Lilia-Chen:
```suggestion
  subscribeTraceEvent(stageThreeTraceRenderInfoEvent, payload => ({ type: 'three-render-info', payload }))
  subscribeTraceEvent(stageThreeTraceHitTestReadEvent, payload => ({ type: 'three-hit-test-read', payload }))
  subscribeTraceEvent(stageThreeTraceVrmUpdateFrameEvent, payload => ({ type: 'vrm-update-frame', payload }))
  subscribeTraceEvent(stageThreeTraceVrmLoadStartEvent, payload => ({ type: 'vrm-load-start', payload }))
  subscribeTraceEvent(stageThreeTraceVrmLoadEndEvent, payload => ({ type: 'vrm-load-end', payload }))
  subscribeTraceEvent(stageThreeTraceVrmLoadErrorEvent, payload => ({ type: 'vrm-load-error', payload }))
  subscribeTraceEvent(stageThreeTraceVrmDisposeStartEvent, payload => ({ type: 'vrm-dispose-start', payload }))
  subscribeTraceEvent(stageThreeTraceVrmDisposeEndEvent, payload => ({ type: 'vrm-dispose-end', payload }))
```

apps/stage-tamagotchi/src/renderer/pages/settings/system/developer.vue
```diff
@@ -109,10 +111,24 @@ const openMarkdownStressWindow = useElectronEventaInvoke(electronOpenDevtoolsWin
...
+    v-motion
+    mb-2
+    icon="i-solar:chart-square-bold-duotone"
+    :text="t('tamagotchi.settings.devtools.pages.lag-visualizer.title')"
+    :initial="{ opacity: 0, y: 10 }"
+    :enter="{ opacity: 1, y: 0 }"
+    :duration="250 + (19 * 10)"
+    :delay="3 * 50"
+    transition="all ease-in-out duration-250"
+    @click="() => router.push('/devtools/lag-visualizer')"
```
Lilia-Chen:
```suggestion
    @click="() => router.push('/devtools/performance-visualizer')"
```

packages/stage-ui-live2d/src/stores/live2d.ts
```diff
@@ -36,7 +38,7 @@ export const defaultModelParameters = {
 }
 
 export const useLive2d = defineStore('live2d', () => {
-  const { post, data } = useBroadcastChannel<BroadcastChannelEvents, BroadcastChannelEvents>({ name: 'airi-stores-live2d' })
+  const { post, data } = useBroadcastChannel<BroadcastChannelEvents, BroadcastChannelEvents>({ name: live2dViewUpdateChannelName })
```
Lilia-Chen:
```suggestion
  const { post, data } = useBroadcastChannel<BroadcastChannelEvents, BroadcastChannelEvents>({ name: 'airi-stores-stage-ui-live2d' })
```

packages/stage-ui-live2d/src/stores/live2d.ts
```diff
@@ -7,9 +7,11 @@ type BroadcastChannelEvents
   = | BroadcastChannelEventShouldUpdateView
 
 interface BroadcastChannelEventShouldUpdateView {
-  type: 'should-update-view'
+  type: 'live2d-should-update-view'
 }
 
+const live2dViewUpdateChannelName = 'airi-stores-stage-ui-live2d'
+
```
Lilia-Chen:
```suggestion
```

packages/stage-ui-three/src/components/Model/vrm-instance-cache.ts
Lilia-Chen: Thanks, it's simplified a lot.

packages/stage-ui-three/src/components/Model/vrm-instance-cache.ts
Lilia-Chen: Thanks, by design, the scopeKey is not intended to be changed.

packages/stage-ui-three/src/components/Model/VRMModel.vue
Lilia-Chen: Yeah, thanks!

packages/stage-ui-three/src/components/Model/VRMModel.vue
```diff
@@ -174,47 +195,286 @@ const vrmLipSync = useVRMLipSync(currentAudioSource)
...
+  //   return
+  //
+  // console.warn(`[VRMModel][load:${_phase}]`, {
+  //   href: typeof window !== 'undefined' ? window.location.href : 'unknown',
+  //   loadSequence,
+  //   modelSrc: modelSrc.value,
+  //   ..._payload,
+  // })
+}
+
```
Lilia-Chen:
```suggestion
```

packages/stage-ui-three/src/components/Model/VRMModel.vue
```diff
@@ -253,203 +498,287 @@ function defaultTookAt(eyeHeight: number): Vec3 {
...
-          x: vrmModelSize.x,
-          y: vrmModelSize.y,
-          z: vrmModelSize.z,
-        })
-      }
+    logVrmLoadDebug('start', {
+      isFirstLoad,
+      reason: loadReason,
+      requestId,
+    })
```
Lilia-Chen:
```suggestion
```

packages/stage-ui-three/src/components/Model/VRMModel.vue
```diff
@@ -253,203 +498,281 @@ function defaultTookAt(eyeHeight: number): Vec3 {
...
-      const animation = await loadVRMAnimation(idleAnimation.value)
-      const clip = await clipFromVRMAnimation(_vrm, animation)
-      if (!clip) {
-        console.warn('No VRM animation loaded')
+        logVrmLoadDebug('end', {
+          isFirstLoad,
+          reason: loadReason,
+          requestId,
+          reused: true,
+        })
```
Lilia-Chen:
```suggestion
```

packages/stage-ui-three/src/components/Model/VRMModel.vue
Lilia-Chen: Indeed, we need to add it.

packages/stage-ui-three/src/components/Model/VRMModel.vue
```diff
@@ -253,203 +498,275 @@ function defaultTookAt(eyeHeight: number): Vec3 {
...
-      // Set model facing direction
-      // Lilia: I brought forward the rotation to the core.ts, so that any ad-hoc rotation will not impact the model centre position.
-      if (isFirstLoad) {
-        // Reset model rotation Y
-        emit('modelRotationY', 0)
-      }
+        if (!airiIblProbe && scene.value)
+          airiIblProbe = createIblProbeController(scene.value)
+
+        commitManagedVrmInstance(reusableInstance)
```
Lilia-Chen, line 616:
```suggestion
        commitManagedVrmInstance(reusableInstance)
        didCommitLoad = true
```

packages/stage-ui-three/src/components/Model/VRMModel.vue
Lilia-Chen: added, thanks

packages/stage-ui-three/src/stores/model-store.ts
```diff
@@ -56,11 +56,20 @@ type BroadcastChannelEvents
...
+  stack?: string
 }
 
+const vrmViewUpdateChannelName = 'airi-stores-stage-ui-three-vrm'
+const vrmViewUpdateRuntimeInstanceId = Math.random().toString(36).slice(2, 10)
+let vrmViewUpdateMessageSequence = 0
+
 export const useModelStore = defineStore('modelStore', () => {
-  const { post, data } = useBroadcastChannel<BroadcastChannelEvents, BroadcastChannelEvents>({ name: 'airi-stores-live2d' })
+  const { post, data } = useBroadcastChannel<BroadcastChannelEvents, BroadcastChannelEvents>({ name: vrmViewUpdateChannelName })
```
Lilia-Chen:
```suggestion
  const { post, data } = useBroadcastChannel<BroadcastChannelEvents, BroadcastChannelEvents>({ name: 'airi-stores-stage-ui-three-vrm' })
```

packages/stage-ui-three/src/stores/model-store.ts
```diff
@@ -56,11 +56,20 @@ type BroadcastChannelEvents
...
-  type: 'should-update-view'
+  type: 'vrm-should-update-view'
+  href: string
+  instanceId: string
+  reason: string
+  sentAt: number
+  stack?: string
 }
 
+const vrmViewUpdateChannelName = 'airi-stores-stage-ui-three-vrm'
```
Lilia-Chen:
```suggestion
```

packages/stage-ui-three/src/components/ThreeScene.vue
Lilia-Chen:
```suggestion
function onVRMModelLoadStart() {
 modelLoaded.value = false
 controlEnable.value = false
 componentState.value = 'loading'
}
```

Lilia-Chen:
```suggestion
  else if (componentState.value !== 'pending') {
```

packages/stage-ui-three/src/stores/model-store.ts
Lilia-Chen: thanks removed.

packages/stage-ui-three/src/stores/model-store.ts
```diff
@@ -70,19 +78,39 @@ export const useModelStore = defineStore('modelStore', () => {
...
   watch(data, (event) => {
-    if (event.type === 'should-update-view') {
+    if (event?.type === 'vrm-should-update-view') {
+      // if (import.meta.env.DEV) {
+      //   console.warn('[VRM shouldUpdateView][recv]', {
+      //     currentHref: typeof window !== 'undefined' ? window.location.href : 'unknown',
+      //     event,
+      //   })
+      // }
+
```
Lilia-Chen:
```suggestion
```

packages/stage-ui-three/src/stores/model-store.ts
```diff
@@ -70,19 +78,32 @@ export const useModelStore = defineStore('modelStore', () => {
...
   watch(data, (event) => {
-    if (event.type === 'should-update-view') {
+    if (event?.type === 'vrm-should-update-view') {
       shouldUpdateViewHooks.value.forEach(hook => hook())
     }
   })
 
   const scale = useLocalStorage('settings/stage-ui-three/scale', 1)
-  const lastModelSrc = useLocalStorage('settings/stage-ui-three/lastModelSrc', '')
+  const lastModelSrc = ref('')
```
Lilia-Chen:
```suggestion
  const lastModelSrc = useLocalStorage('settings/stage-ui-three/lastModelSrc', '')
```

packages/stage-ui/src/composables/use-data-maintenance.ts
Lilia-Chen: removed, thanks

Lilia-Chen:
```suggestion
```

packages/stage-ui/src/components/scenarios/dialogs/model-selector/model-selector-dialog.vue
```diff
@@ -4,44 +4,78 @@ import type { DisplayModel } from '../../../../stores/display-models'
...
+
+const PICK_GUARD_WINDOW_MS = 500
+function handlePick(value: DisplayModel | undefined) {
+  // lilia: Guard against the dialog-opening pointer event being reused as an immediate pick on dialog content.
+  const hasRecentDialogPointerDown = performance.now() - lastDialogPointerDownAt.value < PICK_GUARD_WINDOW_MS
+    return
+  }
+
+  emits('pick', value)
+}
```
Lilia-Chen:
```suggestion
function handlePick(value: DisplayModel | undefined) {
  // Lilia: Guard against the dialog-opening pointer event being reused as an immediate pick on dialog content.
  const hasRecentDialogPointerDown = performance.now() - lastDialogPointerDownAt.value < PICK_GUARD_WINDOW_MS

  if (!showDialog.value || !canAcceptPick.value || !hasRecentDialogPointerDown)
    return

  emits('pick', value)
}
```

packages/stage-ui/src/components/scenarios/settings/model-settings/index.vue
Lilia-Chen: removed, thanks

Lilia-Chen:
```suggestion
```

packages/stage-ui/src/components/scenarios/settings/model-settings/index.vue
Lilia-Chen: removed, thanks

packages/stage-ui/src/components/scenarios/settings/model-settings/index.vue
```diff
@@ -46,21 +46,36 @@ const {
...
```
Lilia-Chen:
```suggestion
```

packages/stage-ui/src/stores/settings/stage-model.ts
Lilia-Chen: removed, thanks

Lilia-Chen:
```suggestion
```

packages/stage-ui/src/stores/settings/stage-model.ts
Lilia-Chen: removed, thanks

packages/stage-ui/src/stores/settings/stage-model.ts
```diff
@@ -68,11 +124,51 @@ export const useSettingsStageModel = defineStore('settings-stage-model', () => {
...
+
```
Lilia-Chen:
```suggestion
```

packages/stage-ui/src/stores/settings/stage-model.ts
Lilia-Chen: removed, thanks

packages/stage-ui/src/stores/settings/stage-model.ts
```diff
@@ -68,11 +124,43 @@ export const useSettingsStageModel = defineStore('settings-stage-model', () => {
...
+  //     currentHref: typeof window !== 'undefined' ? window.location.href : 'unknown',
+  //     senderHref: event.href,
+  //     senderInstanceId: event.instanceId,
+  //     nextValue: event.nextValue,
+  //     previousValue: event.previousValue,
+  //     sentAt: event.sentAt,
+  //   })
+  //   if (event.stack)
+  //     console.warn(`[StageModel][set][recv][stack]\n${event.stack}`)
+  // })
```
Lilia-Chen:
```suggestion
```

packages/stage-ui/src/components/scenes/Stage.vue
```diff
@@ -95,22 +92,6 @@ type PresentEvent
     | { type: 'assistant-append', text: string }
 const { post: postPresent } = useBroadcastChannel<PresentEvent, PresentEvent>({ name: 'airi-chat-present' })
```
Lilia-Chen, line 96:
```suggestion

viewUpdateCleanups.push(live2dStore.onShouldUpdateView(async () => {
  showStage.value = false
  await settingsStore.updateStageModel()
  setTimeout(() => {
    showStage.value = true
  }, 100)
}))

```

packages/stage-ui/src/components/scenarios/settings/model-settings/index.vue
```diff
@@ -46,21 +46,43 @@ const {
...
         useLive2d().shouldUpdateView()
         break
       case DisplayModelFormat.VRM:
-        useModelStore().shouldUpdateView()
+        useModelStore().shouldUpdateView('settings-model-change')
         break
     }
   }
-}, { deep: true })
+}
```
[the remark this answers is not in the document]
Lilia-Chen: no change since Live2D still needed

packages/stage-ui/src/components/scenes/Stage.vue
```diff
@@ -77,11 +78,7 @@ const chatHookCleanups: Array<() => void> = []
 //             cross-window broadcast wiring.
 
 const providersStore = useProvidersStore()
```
Lilia-Chen, line 80:
```suggestion
const providersStore = useProvidersStore()
const live2dStore = useLive2d()
```

packages/stage-ui/src/components/scenes/Stage.vue
```diff
@@ -77,11 +78,7 @@ const chatHookCleanups: Array<() => void> = []
...
-const live2dStore = useLive2d()
-const vrmStore = useModelStore()
-
 const showStage = ref(true)
```
Lilia-Chen, line 82:
```suggestion
const showStage = ref(true)
const viewUpdateCleanups: Array<() => void> = []
```

Lilia-Chen:
I'm still investigating it ... don't merge

Lilia-Chen:
9 March 2026, for today's debug:

## Fix: ThreeScene stuck in `loading` state on main window (tamagotchi)

### Background

This fix originated from an investigation into ThreeScene lifecycle instability in dev mode. In development, `@tresjs/core` registers a `vite:afterUpdate` HMR listener on `import.meta.hot`. This event fires not only on `.vue`/`.ts` module updates, but also on UnoCSS style regeneration (`/__uno.css`) — meaning any UnoCSS class change triggers TresCanvas's internal `handleHMR → mountCustomRenderer(...)`, which remounts the entire Three.js subtree without ThreeScene itself unmounting or `modelSrc` changing.

The existing state machine in ThreeScene did not account for this HMR-driven remount pattern, causing it to reliably get stuck in `loading` after any CSS hot update during development.

While diagnosing this, a more fundamental deadlock was discovered that made the stuck-loading state completely unrecoverable without a full page reload.

### Root Cause: `v-show` deadlock

#### How TresCanvas measures its dimensions

`@tresjs/core` measures canvas dimensions via its internal `useSizes` composable:

```js
// from @tresjs/core internals
function useSizes(windowSize, canvas, debounceMs = 10) {
  const reactiveSize = toValue(windowSize)
    ? useWindowSize()
    : useElementSize(computed(() => toValue(canvas).parentElement))

  return {
    width:  readonly(refDebounced(reactiveSize.width,  debounceMs)),
    height: readonly(refDebounced(reactiveSize.height, debounceMs)),
  }
}
```

By default (no `window-size` prop), TresCanvas measures **its own parent element** via `useElementSize(canvas.parentElement)`. The `@ready` event only fires once all of the following are true:

1. The WebGL renderer has finished initializing
2. `sizes.width` and `sizes.height` are both non-zero
3. `renderer.domElement.width` and `renderer.domElement.height` are both non-zero

If the parent element has `display: none`, `useElementSize` reports `0×0`, the conditions are never satisfied, and `@ready` never fires. Critically, `hasTriggeredReady` is a one-shot flag — if it doesn't fire during the initial mount window, it never will, even if dimensions later become available (unless the entire TresCanvas is unmounted and remounted).

#### The circular dependency

In `apps/stage-tamagotchi/src/renderer/pages/index.vue`, `WidgetStage` (which contains ThreeScene and TresCanvas) was wrapped in a `v-show="!isLoading"` container:

```
componentState initializes as 'pending'
  → isLoading = true
  → v-show="!isLoading" sets display:none on the WidgetStage container
  → useElementSize(canvas.parentElement) reports 0×0
  → @ready never fires
  → componentState never transitions to 'mounted'
  → isLoading stays true forever
  → display:none is never lifted
  permanent deadlock
```

The unlock condition (`@ready`) depended on the lock being released (`display:none` removed), and the lock release depended on the unlock condition. Neither could make progress.

This explains why the main window always showed a perpetual loading overlay, while the settings page ThreeScene worked correctly — the settings page renders ThreeScene directly into the DOM with no `v-show` wrapper, so TresCanvas always has a measurable parent and `@ready` fires normally.

> **General rule:** Any ancestor of TresCanvas must not use `v-show` for conditional visibility if the condition's resolution depends on TresCanvas internal state (e.g., `@ready`, component load phase). Using `v-if` is safe because it fully unmounts and remounts the subtree with a clean DOM.

### Fix

Removed `WidgetStage` from the `v-show` wrapper so it is always present in the DOM. The loading overlay is now an independent `position: absolute; z-index: 99` element rendered on top of the stage, with its own `v-show="isLoading"` that has no circular dependency on TresCanvas internals. **No impact to Live2D.**

### Files Changed

- `apps/stage-tamagotchi/src/renderer/pages/index.vue` — decouple loading overlay from WidgetStage visibility; stage is always mounted, overlay sits on top

Lilia-Chen:
Still investigating other bugs, so do not merge.

Lilia-Chen:
## Debug Notes

---

### Model URL Lifecycle

`stageModelSelectedUrl` used to be derived directly from settings in a way that could go stale — the URL would briefly become `undefined` during a model switch, triggering an unnecessary teardown-and-reload cycle in the renderer. We now treat the URL as a stable ref that only updates once the new model source is confirmed, which eliminates spurious re-mounts.

---

### VRM Instance Cache: Detach-Stash-Reattach

Rather than destroying the VRM instance on every route change, we introduced a scope-keyed cache (`scopeKey = window.location.href`). When a `ThreeScene` unmounts, it *stashes* its `VRM + AnimationMixer + Group` under the current URL key. When the same route mounts again, it *takes* the cached instance and reattaches it to the new TresJS context — skipping the full GLB parse and shader compile. Cache misses fall back to a normal load.

Trace events (`take/stash/clear × hit/miss/stored/evicted/empty`) make it possible to verify the cache is actually working in devtools without guessing.

---

### Scene Lifecycle State Machine

`ThreeScene` now has an explicit phase model: `idle → binding → mounted`. The `sceneMutationLocked` computed blocks any store writes while a binding is in progress or a transaction is open (`sceneTransactionDepth > 0`). This means a settings-preview scene racing against the main stage scene will stall its own writes rather than corrupting the other's state.

The `binding:start` trace cause was added so you can observe the full binding window — not just `binding:complete` — and confirm that the lock actually held during the race.

---

### VRM Preview Resource Cleanup

The preview composable was allocating a `WebGLRenderer` per-call without a guaranteed teardown path. We moved cleanup into the component's `onUnmounted` hook and made the renderer reference explicit, so preview renders don't accumulate GPU contexts across model-selector opens.

Issue #1159, opened 2026-03-06
Feature: always on top


vochant:
### Clear and concise description of the problem

Hope to add the feature to always keep the model window on top.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

LemonNekoGH:
Thank you for submitting this issue, but, what OS and AIRI version did you use?

vochant:
> Thank you for submitting this issue, but, what OS and AIRI version did you use?

Ubuntu 24.04 (Wayland), v0.9.0-alpha.2

LemonNekoGH:
> Ubuntu 24.04 (Wayland), v0.9.0-alpha.2

Well, I don't have similar environment, let me ask other teamates.

lietblue:
I've added a toggle button for 'Always on Top' here https://github.com/moeru-ai/airi/pull/1183 , but I noticed that Airi should be always on top by default.
Could you share more details about your setup?

Which desktop environment and version are you using?

vochant:
> I've added a toggle button for 'Always on Top' here https://github.com/moeru-ai/airi/pull/1183 , but I noticed that Airi should be always on top by default.
> Could you share more details about your setup?
> 
> Which desktop environment and version are you using?

I'm using GNOME 44 and installed the .deb package for amd64 from releases using apt.

0xSelenicDove:
Try the latest version. Does this issue still persist?

PR #1201, opened 2026-03-08
feat(stage-tamagotchi): add always-on-top toggle IPC API

JasonOA888:
Implements #1159

## Changes
Added IPC API for toggling always-on-top mode:

### Main Process
- `window:set-always-on-top(enabled)` - Toggle always-on-top
- `window:get-always-on-top()` - Get current state

### Preload
- `window.alwaysOnTop.set(enabled)` - Set mode
- `window.alwaysOnTop.get()` - Get mode

## Default
Enabled (current behavior preserved)

## Usage
```typescript
// Renderer process
const enabled = await window.alwaysOnTop.get()
await window.alwaysOnTop.set(false) // Disable always-on-top
```

Note: UI toggle button can be added separately.


nekomeowww:
https://github.com/moeru-ai/airi/pull/1183 already implemented.

PR #1213, opened 2026-03-09
fix(minecraft): add stop method to Ticker to prevent resource leak

kbhujbal:
The Ticker class starts an infinite while(true) loop but has no way to stop it. Once created, the loop runs forever even after Mineflayer.stop() is called, leaking memory and CPU. This adds a stop() method that breaks the loop, clears the initial timer, and removes all listeners. The Mineflayer.stop() method now calls ticker.stop() during shutdown.

## Description

The `Ticker` class starts a `while(true)` loop but has no way to break out of it. Once created, the loop keeps running even after `Mineflayer.stop()` is called  the bot shuts down but the ticker keeps ticking in the background, leaking memory and CPU.

Added a `stop()` method to `Ticker` and wired it into `Mineflayer.stop()` alongside the existing `connectionSupervisor.stop()` call.


shinohara-rin:
Thanks for the PR. We're planning to deprecate the current mineflayer-based minecraft bot and migrate to fabric/neoforge soon btw

PR #1222, opened 2026-03-09
fix(llm): flatten content array for OpenAI-compatible providers

Reisenbug:
## Description
use deepseek deepseek-chat
I chatted casually for a while, then I ask "天气怎么样"(What's the weather like)
then
{
  "error": {
    "message": "Failed to deserialize the JSON body into the target type: messages[19]: invalid type: sequence, expected a string at line 1 column 9837",
    "type": "invalid_request_error",
    "param": null,
    "code": "invalid_request_error"
  }
}

chat.ts injects a context message with content as an array ([{ type: 'text', text: '...' }]).


Garfield550:
All model messages are processed by `packages/stage-ui/src/stores/llm.ts`. If the issue you encounter only occurs with **DeepSeek**, it is recommended to create/modify a `ChatProvider` to handle it.

nekomeowww:
Did you installed VSCode plugin? Or experimenting WebSocket? How can I reproduce this bug?

Reisenbug:
> All model messages are processed by `packages/stage-ui/src/stores/llm.ts`. If the issue you encounter only occurs with **DeepSeek**, it is recommended to create/modify a `ChatProvider` to handle it.

I searched online, finding that other projects like litellm and vscode-ai-toolkit fixed this the same way – in the shared layer rather than per-provider, since some strict OpenAI-compatible provider has the same issue

Reisenbug:
> Did you installed VSCode plugin? Or experimenting WebSocket? How can I reproduce this bug?

No VSCode plugin or WebSocket involved
The issue occurs in the standard web UI chat flow with DeepSeek as the provider.I added a temporary log to output the message, and it shows this:
{
    "role": "user",
    "content": [{ "type": "text", "text": "These are the contextual information..." }]
}
other outputs show this:
    {                                                
      "role": "user",                                
      "content": "天气怎么样"                
    }

DeepSeek strictly requires content to be a string.

Unfortunately I wasn't able to stably reproduce the 400 error. But the problematic message format is confirmed present in every request via the debug log above.

packages/stage-ui/src/stores/llm.ts
```diff
@@ -34,6 +34,12 @@ function sanitizeMessages(messages: unknown[]): Message[] {
         content: `User encountered error: ${String(m.content ?? '')}`,
       } as Message
     }
+    if (m && Array.isArray(m.content)) {
+      const contentParts = m.content as { type?: string, text?: string }[]
+      if (!contentParts.some(p => p?.type === 'image_url')) {
+        return { ...m, content: contentParts.map(p => p?.text ?? '').join('') } as Message
+      }
+    }
```
nekomeowww, line 46: /gemini help me generate a suggestion that contains `// NOTICE:` for explaining why this matters, and include essential failure case in PR description and PR comments. So that we sure this part will not be removed as backward compatibility for the critical path of LLM request rendering

PR #1224, opened 2026-03-09
refactor(stage-ui): steps of onboarding dialog

LemonNekoGH:
Use a array to describe steps for better extensible. I will add permission request step later.

### DIFF
--- a/packages/stage-ui/src/components/scenarios/dialogs/onboarding/onboarding.vue
+++ b/packages/stage-ui/src/components/scenarios/dialogs/onboarding/onboarding.vue
@@ -24,13 +24,12 @@ interface Emits {
   (e: 'skipped'): void
 }
 
-const props = defineProps<{
+const { extraSteps = [] } = defineProps<{
   extraSteps?: OnboardingStep[]
 }>()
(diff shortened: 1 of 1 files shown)


packages/stage-ui/src/components/scenarios/dialogs/onboarding/onboarding.vue
```diff
@@ -9,17 +18,20 @@ import StepWelcome from './step-welcome.vue'
...
 import { useConsciousnessStore } from '../../../../stores/modules/consciousness'
 import { useProvidersStore } from '../../../../stores/providers'
-import { OnboardingContextKey } from './utils'
 
 interface Emits {
   (e: 'configured'): void
   (e: 'skipped'): void
 }
 
+const props = defineProps<{
```
luoling8192: Is it better to use 'withDefaults' instead of computed?

LemonNekoGH: ~~You are abusolutely right.~~

I will fix it.

luoling8192 (approved):

PR #1231, opened 2026-03-10
feat(stage-tamagotchi): add top drag area to move window

fordelkon:
## Description

Added a draggable area at the top of the window to allow users to move the entire application window by dragging.

## Key Changes

- Added a **48px height draggable area** at the top of the window
- Shows a **flashing border hint** on mouse hover
- Supports **Windows/macOS** (via `startDraggingWindow` API)
- Supports **Linux** (via `drag-region` CSS class)

## User Experience

- Border hint appears when hovering over the top area
- Flash animation matches the existing window border hint style
- Dragging the area moves the entire application window

## Linked Issues
When resizing the window from the edges, the drag button and setting button may occasionally be shifted outside the visible screen area.
issue1

The window can be moved back into the visible screen area by dragging it from the top draggable area.

## Additional Context

The implementation follows the existing **window border hint styling** to maintain UI consistency.  
Linux systems use the `drag-region` CSS class as an alternative due to Electron API limitations.

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/pages/index.vue
+++ b/apps/stage-tamagotchi/src/renderer/pages/index.vue
@@ -333,24 +333,6 @@ watch([stream, () => vadLoaded.value], async ([s, loaded]) => {
     relative z-2 h-full overflow-hidden rounded-xl
     transition="opacity duration-500 ease-in-out"
   >
-    <!-- Drag area at top to move window -->
-    <Transition
-      enter-active-class="transition-opacity duration-250 ease-in-out"
-      enter-from-class="opacity-50"
-      enter-to-class="opacity-100"
-      leave-active-class="transition-opacity duration-250 ease-in-out"
-      leave-from-class="opacity-100"
-      leave-to-class="opacity-50"
-    >
-      <div v-if="showDragHint && !isLoading" class="pointer-events-none absolute left-0 top-0 z-10 h-12 w-full">
-        <div
-          :class="[
-            'b-primary/50',
-            'h-full w-full animate-flash animate-duration-3s animate-count-infinite b-t-4 rounded-t-2xl',
-          ]"
-        />
-      </div>
-    </Transition>
     <div
       v-show="!isLoading"
       absolute left-0 top-0 z-10 h-12 w-full


fordelkon:
Addressed in latest commit.  Added import { isLinux } from 'std-env' - a simple boolean constant. Now isLinux is just a boolean from std-env that's determined at build time, so no async detection needed. The drag functionality will work correctly on all platforms.

nekomeowww:
Would you mind include some screen recording for me to understand the case and the version you implemented?

fordelkon:
Sure, I'd be happy to.

In the previous version, there was an issue where the Transition and Settings buttons would go off-screen when scaling down from a large canvas to a smaller one.

https://github.com/user-attachments/assets/5054b014-f815-4ec7-bf0f-3f6c4675ef07

To solve this issue, I've implemented a feature that allows users to pan the canvas from its top edge.

https://github.com/user-attachments/assets/ca50ad9a-fa03-4be0-96c7-9d5049c7f80e

apps/stage-tamagotchi/src/renderer/pages/index.vue
```diff
@@ -322,6 +333,34 @@ watch([stream, () => vadLoaded.value], async ([s, loaded]) => {
...
+    >
+      <div v-if="showDragHint && !isLoading" class="pointer-events-none absolute left-0 top-0 z-10 h-12 w-full">
+        <div
+          :class="[
+            'b-primary/50',
+            'h-full w-full animate-flash animate-duration-3s animate-count-infinite b-t-4 rounded-t-2xl',
+          ]"
+        />
+      </div>
+    </Transition>
```
nekomeowww: Let's remove this for now, I will merge your request since it works perfectly and the feature request was reasonable, however i think there should be other ways to implement the UI of the top bar hint more elegant and easy to understand.

nekomeowww:
```suggestion
```

nekomeowww: I removed it, and perhaps you could raise another Pull Request with better design of the top hint area, or I could implement this later.

nekomeowww:
Not quite implement it correctly, I will offer more comments tomorrow, meanwhile, I will revert this Pull Request.

PR #1233, opened 2026-03-10
fix(stage-ui-live2d,stage-tamagotchi): use temp file to prevent config loss and handle more Live2D model load failures

### LINKED ISSUE #1207: AIRI desktop entered a broken state around 2026-03-08 15:00: built-in/custom models fail to load, config/state is lost, and previous saved data cannot restore a working state
### Describe the bug

[airi_application_events_1410_1430.txt](https://github.com/user-attachments/files/25824857/airi_application_events_1410_1430.txt)

[Report.wer.txt](https://github.com/user-attachments/files/25824859/Report.wer.txt)

[Report.wer.txt](https://github.com/user-attachments/files/25824861/Report.wer.txt)

## Summary

AIRI desktop on Windows entered a system-wide broken state around 2026-03-08 14:20 local time.

Before that time, my custom Live2D model worked normally.
After that point, AIRI itself could still launch, but the whole model/state system became abnormal:

- custom models could no longer be loaded/read
- built-in models also failed to load/read
- config/state became lost/reset
- previously saved AIRI app-data folders still contained important data, but copying them back no longer restored a usable working state

Because built-in models were also affected, and because config/state recovery also failed, this does not look like a single-model issue.
It looks like AIRI entered a broader broken state affecting model loading and state persistence/recovery together.

---

## Environment

- OS: Windows 11 (replace with exact version if desired)
- AIRI versions tested after the issue started:
  - v0.8.4
  - v0.9.0-alpha,1
  - v0.9.0-alpha.2
- Installed from official GitHub Releases

## What I tested

I tested all of the following:

- deleted `%AppData%\Roaming\ai.moeru.airi`
- deleted `%LocalAppData%\ai.moeru.airi`
- deleted `%LocalAppData%\ai.moeru.airi-updater`
- reinstalled AIRI
- tested v0.8.4 ,v0.9.0-alpha.1, v0.9.0-alpha.2
- tested built-in models
- tested with fresh config/state
- tested restoring previous saved AIRI data folders
- tested copying old saved folders into a fresh AIRI state

Results:
- built-in models still failed
- custom model still failed
- restoring old saved folders did not recover a usable working state

## Windows event / WER evidence

The earliest Windows-level failure I could confirm was around 2026-03-08 14:19:28 (UTC+8).

After that, AIRI repeatedly triggered the same failure path within the next several minutes (around 14:19, 14:21, and 14:26), with stable exception signatures such as:

0xc0000005

0xc000041d

and the same fault offset:

0x00000000003a4440

The app itself can still launch but all of the configs are missing.

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (32) x64 13th Gen Intel(R) Core(TM) i9-13980HX
    Memory: 18.49 GB / 31.63 GB
  Binaries:
    Node: 22.13.1 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 145.0.7632.160
    Edge: Chromium (143.0.3650.96)
    Internet Explorer: 11.0.26100.7309
```

### Validations

### Contributions

pratyush618:
## Summary

Fixes #1207 — AIRI desktop enters a broken state where models fail to load and config/state is lost after a crash.

- **Model loading error handling**: Added `catch` block to `loadModel()` in `Model.vue` — errors are now logged and emitted via an `error` event instead of propagating uncaught
- **Atomic config writes**: Config saves now write to a `.tmp` file first, then atomically rename to the target path, preventing truncated/corrupted configs from mid-write crashes
- **Backup before auto-heal**: Before overwriting a corrupted config with defaults, the old file is copied to `{path}.bak` so user data can be manually recovered
- **OPFS cache clear**: Added `OPFSCache.clearAll()` as a recovery mechanism to wipe cached model data and allow fresh re-caching

## Test plan

- [ ] Provide an invalid model URL and verify error is logged and `error` event is emitted without a silent blank screen
- [ ] Corrupt a config file manually, start the app, verify `.bak` file is created before healing
- [ ] Kill the app process during a config save cycle, verify config file isn't truncated on next startup
- [ ] Call `OPFSCache.clearAll()` from dev console, verify models can be re-cached fresh

Closes #1207


packages/stage-ui-live2d/src/utils/opfs-loader.ts
```diff
@@ -12,6 +12,18 @@ declare global {
 }
 
 export class OPFSCache {
+  static async clearAll(): Promise<void> {
```
nekomeowww, line 15: Where will use this `clearAll(...)` function?

pratyush618: It's used as a recovery mechanism — wired up to a "Clear Model Cache" button in the Live2D settings panel (packages/stage-ui/src/components/scenarios/settings/model-settings/live2d.vue).   
 Users can use it to clear corrupted OPFS cache when models fail to load. See the latest commit: [89a8601](https://github.com/moeru-ai/airi/pull/1233/commits/89a8601cd88fb5bf848e0866ea6f8ed30b63f75a)

Issue #1249, opened 2026-03-10
bug: Windows x64 version fails to launch - Missing package 'rou3' in Alpha 12


some111111:
### Describe the bug

A JavaScript error occurred in the main process

Uncaught Exception:
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'rou3' imported from C:\Program
Files\airi\resources\app.asar\node_modules\h3\dist\h3-Dol7UbDx.mjs
at Object.getPackageJSONURL (node:internal/modules/package_json_reader:301:9)
at packageResolve (node:internal/modules/esm/resolve:780:81)
at moduleResolve (node:internal/modules/esm/resolve:871:18)
at defaultResolve (node:internal/modules/esm/resolve:1003:11)
at #cachedDefaultResolve (node:internal/modules/esm/loader:719:20)
at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:736:38)
at ModuleLoader.resolveSync (node:internal/modules/esm/loader:765:52)
at #resolve (node:internal/modules/esm/loader:701:17)
at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:621:35)
at ModuleJob.syncLink (node:internal/modules/esm/module_job:160:33)

### System Info

```Shell
OS: Windows 10/11 x64

App Version: 0.9.0-alpha.12

Error Message: Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'rou3'
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

xianyu1124:
我也遇到了一样的情况，从alpha.7更新到12，更新完成后启动提示报错

woleigedouer:
+1 

<img width="711" height="494" alt="Image" src="https://github.com/user-attachments/assets/a7122b22-927a-491c-b012-7ee780817ff6" />

[...]

PR #1253, opened 2026-03-11
fix(stage-tamagotchi): patch startup and onboarding regressions

Ver-zhzh:
## Summary

Fixes [#1249](https://github.com/moeru-ai/airi/issues/1249) and also addresses two related onboarding regressions in `stage-tamagotchi`.

Changes included:
- include `h3` and `rou3` explicitly in the packaged desktop app so Windows release builds no longer fail at startup with `ERR_MODULE_NOT_FOUND`
- keep onboarding model selection usable when the model list is expanded by preventing the sticky action button from being pushed out of view
- make post-onboarding tool compatibility discovery best-effort so a failed probe does not take down the renderer right after setup

### Desktop startup packaging
Recent Windows release builds can fail during startup with an error like:

`Cannot find package 'rou3' imported from ...\\node_modules\\h3\\dist\\*.mjs`

This happens because `h3` resolves `rou3` at runtime, but the packaged app did not reliably include that dependency under the current Electron Builder + pnpm setup.

This PR explicitly includes both `h3` and `rou3` in the desktop package.

### Onboarding model selection layout
When the onboarding model selector is expanded, the `Save and Continue` button can be pushed out of view.

This PR keeps the list area shrinkable/scrollable and pins the action button in a sticky bottom section so the flow stays usable.

### Renderer stability after onboarding
After selecting a model, the renderer may immediately run a compatibility discovery probe.

This PR wraps that probe in error handling so provider/model-specific probe failures do not crash the stage renderer.

## Testing

Tested locally with:
- `pnpm lint:fix`
- `pnpm typecheck`
- desktop packaging with `electron-builder --dir`
- launching the unpacked Windows app and confirming it stays alive
- manually verifying the onboarding model selection layout

## Notes

`pnpm lint:fix` and `pnpm typecheck` still report unrelated existing issues in the current repository state, including:
- `packages/server-runtime`
- `services/minecraft`

These failures were not introduced by this PR.

### DIFF
--- a/apps/stage-tamagotchi/electron-builder.config.ts
+++ b/apps/stage-tamagotchi/electron-builder.config.ts
@@ -75,6 +75,13 @@ export default {
     '!**/node_modules/**/{CHANGELOG.md,README.md,README,readme.md,readme}',
     '!**/node_modules/**/{.turbo,test,src,__tests__,tests,example,examples}',
     '**/node_modules/debug/**/*',
+    // NOTICE: h3@2.x resolves rou3 at runtime from its distributed ESM entry.
+    // Electron Builder misses this transitive dependency in the packaged app under pnpm,
+    // which makes released Windows builds fail at startup with:
+    // `Cannot find package 'rou3' imported from ...\\node_modules\\h3\\dist\\*.mjs`.
+    // Keep h3/rou3 explicitly included until upstream packaging reliably preserves them.
+    '**/node_modules/h3/**/*',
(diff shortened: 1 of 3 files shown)


nekomeowww:
This should not be fixed like that.

Garfield550:
Thank you for your efforts. We are refactoring the dependency build system to fix the missing rou3 dependency issue on Windows x86_64.

For modifications to the onboarding UI, please open a new PR. Thank you

PR #1256, opened 2026-03-11
feat(providers): add Amazon Bedrock provider

chaosreload:
## Summary

Add support for Amazon Bedrock as an LLM provider using the native **Converse API** with API Key authentication.

## Motivation

AWS Bedrock provides access to frontier models (Claude, Amazon Nova, Llama, DeepSeek etc.) with enterprise-grade security and compliance. Many teams running AI workloads on AWS infrastructure want to use Bedrock directly, keeping all traffic within the AWS network.

## Implementation

- Uses Amazon Bedrock's native **Converse API** (`bedrock-runtime.{region}.amazonaws.com/model/{modelId}/converse`)
- Authentication via **Amazon Bedrock API Key** (`Authorization: Bearer <key>`) — no SigV4 signing, no extra dependencies
- Config: `apiKey` (Bedrock API key) + `region` (AWS region, default: `us-east-1`)
- All Bedrock-specific config is declared via the generic `onboardingFields` mechanism on the provider definition — no `isAmazonBedrock` branching, no `ProviderConfigData` type changes
- Dynamic model listing via `ListFoundationModels` + `ListInferenceProfiles` APIs, with fallback to a static list
- Streaming: calls `/converse` (standard JSON response), then re-emits the full text as `text/event-stream` character-by-character — bearer-token auth does not support the binary AWS Event Stream protocol required by `/converse-stream`

## Supported Models

**Anthropic Claude (via Bedrock)**
- Claude Opus/Sonnet/Haiku 4.x
- Claude Sonnet 3.7 (hybrid reasoning)
- Claude Sonnet 3.5 v2

**Amazon Nova**
- Nova Pro (multimodal)
- Nova Lite (multimodal, low cost)
- Nova Micro (text-only, lowest cost)

**Meta / Others**
- Llama 3.3 70B Instruct
- DeepSeek, Moonshot, Minimax models available via inference profiles

## Authentication

Generate a Bedrock API key in: **AWS Console → Amazon Bedrock → API Keys**

> Note: Long-term API keys have a 30-day expiry. AWS recommends them for development/exploration use.

## Notes

⚠️ **CORS in browser**: Like other providers (Anthropic, etc.), direct browser calls to Bedrock are subject to CORS restrictions. Works out of the box in Tauri desktop app and Node.js server environments. For browser use, a CORS proxy is needed.

## Testing

```
pnpm vitest run packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.test.ts
```


packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
```diff
@@ -0,0 +1,405 @@
...
+        const body = JSON.parse((init?.body as string) || '{}') as any
+        const messages: any[] = body.messages || []
+        const modelId: string = body.model || model
+
+        // Separate system messages
+        const systemMessages = messages.filter(m => m.role === 'system')
+        const chatMessages = messages.filter(m => m.role !== 'system')
+
+        // Convert to Converse messages format
+        const converseMessages = mergeConsecutiveRoles(
```
shinohara-rin, line 89: > Uses Amazon Bedrock's OpenAI-compatible endpoint 

well, it doesn't look like openai compatible here...

packages/stage-ui/package.json
```diff
@@ -91,6 +91,7 @@
     "@xsai/tool": "catalog:",
     "@xsai/utils-chat": "catalog:",
     "animejs": "^4.3.6",
+    "aws4fetch": "catalog:",
```
shinohara-rin: can we do this swiftly without introducing new deps?

or maybe import dynamically?

shinohara-rin (changes requested):
this pr is dragging more than a comfortable amount of bedrock specific code into the shared space

the isAmazonBedrock onboarding branching and ProviderConfigData type changes should be replaced with a generic customFields mechanism

chaosreload:
Thanks for the review @shinohara-rin!

Addressed in commit ba1556f3: replaced the `isAmazonBedrock` branching and `ProviderConfigData` type changes with a generic `onboardingFields` mechanism.

**Changes made:**
- Added `ProviderOnboardingField` interface and `onboardingFields` field to `ProviderDefinition` — providers can declare their custom auth fields
- `ProviderConfigData` no longer has Bedrock-specific fields; now uses `customFields?: Record<string, string>` instead
- `step-provider-configuration.vue` removed `isAmazonBedrock` computed and static refs; custom fields are now rendered dynamically via `v-for` over `provider.onboardingFields`
- `onboarding.vue` spreads `customFields` into the config generically
- Amazon Bedrock’s `accessKeyId`, `secretAccessKey`, and `region` fields are now declared in the provider definition itself via `onboardingFields`

This approach also makes it easy for future providers with non-standard auth (e.g. OAuth tokens, multi-field credentials) to declare their own onboarding fields without touching the shared components.

packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
```diff
@@ -0,0 +1,160 @@
...
+  },
+
+  validationRequiredWhen(config) {
+    return !!config.accessKeyId?.trim() && !!config.secretAccessKey?.trim()
+  },
+
+  validators: {
+    validateConfig: [],
+    validateProvider: [],
+  },
```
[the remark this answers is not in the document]
chaosreload: Fixed in commit `903be205` — added `validators.validateProvider` that makes a real AWS API call to `bedrock.amazonaws.com/foundation-models` and returns `{ valid: false }` on 401/403.

packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
```diff
@@ -0,0 +1,160 @@
...
+    .string()
+    .default('us-east-1'),
+  accessKeyId: z
+    .string(),
+  secretAccessKey: z
+    .string(),
+  sessionToken: z
+    .string()
+    .optional(),
+})
```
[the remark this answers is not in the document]
chaosreload: Fixed in commit `903be205` — added `.regex(/^[a-z]{2,3}-[a-z]+-\d+$/, ...)` validation on the region field in the Zod schema.

packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
```diff
@@ -0,0 +1,160 @@
...
+
+type AmazonBedrockConfig = z.input<typeof amazonBedrockConfigSchema>
+
+function createAmazonBedrock(config: AmazonBedrockConfig) {
+  const region = config.region || 'us-east-1'
```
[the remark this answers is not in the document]
chaosreload: Fixed in commit `cf725ce0` — changed `AmazonBedrockConfig` from `z.input<...>` to `z.infer<...>` so `region` is typed as `string` (default guaranteed by schema). Removed the `|| 'us-east-1'` fallback and the unnecessary type casts in `listModels`.

packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
```diff
@@ -0,0 +1,160 @@
...
+      section: 'advanced',
+    }),
+  }),
+
+  createProvider(config) {
+    return createAmazonBedrock(config)
+  },
+
+  extraMethods: {
+    listModels: async () => ([
```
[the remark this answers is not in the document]
chaosreload: Fixed in commit `903be205` — updated signature to `async (config, _provider)` to match `ProviderExtraMethods<TConfig>`.

packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
chaosreload: You're right, that comment is misleading. The implementation actually intercepts the xsai request and translates it to Bedrocks Converse API format — not OpenAI-compatible at all. Will clean up the comments to accurately describe what's happening.

packages/stage-ui/package.json
chaosreload: Good point. `aws4fetch` is used for AWS SigV4 request signing. The alternatives are:
1. Remove the dep and implement SigV4 using the Web Crypto API (available in browsers, Node.js 18+, and edge runtimes) — more code but zero new deps
2. Keep `aws4fetch` — it's ~3KB gzipped and purpose-built for browser/edge SigV4

Happy to go with option 1 if you prefer no new dependencies. Would a native Web Crypto implementation be acceptable, or is there a preferred signing utility already in the project?

packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
chaosreload: Fixed in commit `0902bbc0` — updated both misleading comments:
- `// Parse incoming OpenAI format request` → `// Parse xsai chat request body (messages array + model)`
- `// Convert to OpenAI SSE stream format` → `// Wrap Converse response into xsai-compatible SSE stream (chat.completion.chunk format)`

packages/stage-ui/package.json
chaosreload: Fixed in commit `64d1a9d3` — switched to Amazon Bedrocks native API key authentication.

AWS recently added native API key support to Bedrock (generated in AWS Console → Bedrock → API Keys, used as `Authorization: Bearer <key>`). This completely removes the need for `aws4fetch`:

- Config simplified to `apiKey` + `region` (same pattern as other providers)
- Auth is just a Bearer token header — no SigV4 signing
- `aws4fetch` removed from `package.json`
- Converse API format translation unchanged

Note: Bedrock API keys expire in 30 days (long-term) per AWS recommendations for dev/exploration use.

chaosreload: Fixed in commits `64d1a9d3`–`3f6146c3` — switched from SigV4 to Amazon Bedrock's native API key authentication, which completely eliminates the `aws4fetch` dependency.

AWS added native API key support to Bedrock (available in AWS Console → Bedrock → API Keys). The key is passed as a standard `Authorization: Bearer <key>` header — no signing required. This brings the auth pattern in line with all other providers in this project.

Changes:
- Config simplified to `apiKey` + `region` (same pattern as all other providers)
- All SigV4 signing code removed
- `aws4fetch` removed from both `package.json` and `pnpm-lock.yaml`
- Zero new dependencies introduced

Locally tested: provider validation, model listing, and chat all work correctly with the API key.

chaosreload:
Updated in the latest push (bf7d3969):

**Authentication simplified**: Removed SigV4 / aws4fetch entirely. Now uses Amazon Bedrock native **API Key** auth () — no extra dependencies, no browser-incompatible crypto.

**onboardingFields already in place** (from ba1556f3): All Bedrock-specific config is declared via the generic  mechanism on the provider definition. No  branching, no  type changes — the shared space is clean.

**Streaming**: Switched from  (binary AWS Event Stream protocol, unsupported with bearer-token auth) to  + SSE re-emission. The full response JSON is received, then re-emitted character-by-character as a standard , so the rest of the xsai pipeline sees a normal streaming response.

**Validator**:  guards the validator so it only fires when  is non-empty. Validation hits the lightweight  endpoint (GET, no model invocation needed).

chaosreload:
@shinohara-rin Thanks again for the review! All feedback has been addressed in the latest push. Would appreciate a re-review when you have time 🙏

nekomeowww (approved):

chaosreload:
Hi @nekomeowww, the previous approval got auto-dismissed after I rebased onto upstream/main today (only `pnpm-lock.yaml` was updated to reflect upstream's new dependencies — no code changes). Could you re-approve when you get a chance? Thanks!

chaosreload:
Hey @shinohara-rin 👋 — could you take a look at this when you have a moment? The conflict has been resolved and CI is passing. Would love to get a maintainer review. Thanks!

packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
```diff
@@ -0,0 +1,400 @@
...
+              choices: [{ delta: { role: 'assistant' }, index: 0, finish_reason: null }],
+            })
+
+            for (const char of fullText) {
+              enqueue({
+                id,
+                object: 'chat.completion.chunk',
+                choices: [{ delta: { content: char }, index: 0, finish_reason: null }],
+              })
+              await new Promise(r => setTimeout(r, 10))
```
shinohara-rin:
```suggestion

```
is this artifical latency necessary?

we could just stream the fullText as a single large chunk if that doesn't break stuff.

chaosreload:
Fixed in commit `5c38b1ad` — removed the character-by-character loop and the `setTimeout(10ms)` artificial delay. The full response is now emitted as a single SSE content chunk, which is simpler and doesn't introduce unnecessary latency.

packages/stage-ui/src/libs/providers/providers/amazon-bedrock/index.ts
chaosreload: @shinohara-rin Done! Removed the char-by-char loop and the `setTimeout(10ms)` artificial delay. The provider now emits `fullText` as a single SSE content chunk. Tests passing, pushed in commit 5c38b1ad.

chaosreload:
Hi @nekomeowww @shinohara-rin — CI is passing and the two outstanding issues have been fixed:

- **Generic `onboardingFields`** — addressed since Mar 14
- **Streaming simplified** — removed char-by-char loop and 10ms delay in `5c38b1ad` (Apr 13)

Could one of you re-approve so this can be merged? The previous approval was auto-dismissed after the Mar 31 rebase. Thanks!

shinohara-rin (approved):

shinohara-rin:
Sorry for the delay! merged

PR #1288, opened 2026-03-11
fix(stage-tamagotchi): provide the missing core components to createCoreContext

sumimakito:
Register the messageCompiler, messageResolver, and localeFallbacker. It could be due to **@intlify/core** being too minimalistic, so that we need to provide more core components imported from **@intlify/core** back to `createCoreContext`.

While **vue-i18n** registers them globally:

```
registerMessageCompiler(compile)
registerMessageResolver(resolveValue)
registerLocaleFallbacker(fallbackWithLocaleChain)
```

See: https://github.com/intlify/vue-i18n/blob/884060b63122ccca3088e4af283ef223a1e2bd29/packages/vue-i18n/src/runtime.ts#L15-L22

... but we will provide them to `createCoreContext` here for better scoping.

Screenshot 2026-03-12 at 0 28 38

### DIFF
--- a/apps/stage-tamagotchi/src/main/libs/i18n/index.ts
+++ b/apps/stage-tamagotchi/src/main/libs/i18n/index.ts
@@ -160,7 +160,7 @@ export function createI18n<Schema extends Record<string, any> = Record<string, a
     // ```
     // See: https://github.com/intlify/vue-i18n/blob/884060b63122ccca3088e4af283ef223a1e2bd29/packages/vue-i18n/src/runtime.ts#L15-L22
     //
-    // ... but we will provide them via `createCoreContext` for better scoping.
+    // ... but we will provide them to `createCoreContext` here for better scoping.
     messageCompiler: compile,
     messageResolver: resolveValue,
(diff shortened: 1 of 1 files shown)


apps/stage-tamagotchi/src/main/libs/i18n/index.ts
```diff
@@ -149,6 +149,21 @@ export function createI18n<Schema extends Record<string, any> = Record<string, a
...
+    // While vue-i18n registers them globally:
+    //
+    // ```
+    // registerMessageCompiler(compile)
+    // registerMessageResolver(resolveValue)
+    // registerLocaleFallbacker(fallbackWithLocaleChain)
+    // ```
+    // See: https://github.com/intlify/vue-i18n/blob/884060b63122ccca3088e4af283ef223a1e2bd29/packages/vue-i18n/src/runtime.ts#L15-L22
+    //
+    // ... but we will provide them via `createCoreContext` for better scoping.
```
sumimakito:
```suggestion
    // ... but we will provide them to `createCoreContext` here for better scoping.
```

PR #1289, opened 2026-03-11
fix(tray): auto-restore window position from snapshot on startup

dasilva333:
This PR fixes the issue where the main window loads in an incorrect or random position on startup. It ensures the window automatically restores its position from a saved snapshot if one exists in the configuration.


nekomeowww:
Shouldn't it auto-restore for now? I cannot see the difference.

dasilva333:
Updated with refinements (Sett/Restore Home labels, comments, synchronization with main).

PR #1295, opened 2026-03-11
feat(speech): pipeline stability and audio quality fixes

dasilva333:
- resolves VAD crackling by properly connecting source nodes
- fixes transcription network hangs by skipping empty buffers
- syncs special token timings with playback
- resolves tool call mapping logic in voice chat

### DIFF
--- a/apps/stage-tamagotchi/src/main/windows/onboarding/index.ts
+++ b/apps/stage-tamagotchi/src/main/windows/onboarding/index.ts
@@ -66,9 +66,10 @@ export function setupOnboardingWindowManager(params: {
 
     const { context } = createContext(ipcMain, newWindow)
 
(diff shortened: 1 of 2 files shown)


shinohara-rin:
We'd like to see some demos if that's possible

dasilva333:
Closing this PR for now. While the diff contains reasonable improvements to the speech pipeline, the user is experiencing persistent 'crackly' audio quality in their local environment that prevents a clean demonstration of these gains.

Key Findings:
- Raw bytes verified correct at both download and server intake.
- Problem appears isolated to the user's specific setup.
- Hardware interference from a 'microphone array' or 2-device setup is suspected.

We'll revisit these improvements once we can demonstrate tangible results in a clean environment.

PR #1302, opened 2026-03-12
feat(stage-pages,stage-ui,i18n): add OpenRouter as a speech (TTS) provider

monolithic827:
## Summary

- Add OpenRouter as a built-in speech provider, using the `/chat/completions` API with audio modality (`modalities: ["text", "audio"]`) since OpenRouter does not support the `/audio/speech` endpoint
- Fetches available audio models dynamically from `/models?output_modality=audio`
- Fix a race condition in the speech store where provider selection was reset on startup before async validation completed (affects all speech providers, not just OpenRouter)

## Details

OpenRouter supports audio output through its chat completions API with streaming. The provider bridges this into the existing `SpeechProvider` interface using a custom `fetch` that:

1. Sends a streaming chat completion request with `modalities: ["text", "audio"]` and `format: "pcm16"`
2. Parses SSE chunks, collecting base64-encoded `delta.audio.data`
3. Wraps the decoded PCM16 data in a WAV header so the browser can play it

### Files changed

- `packages/stage-ui/src/stores/providers.ts` - new `openrouter-audio-speech` provider metadata entry; expose `providerRuntimeState`
- `packages/stage-pages/src/pages/settings/providers/speech/openrouter-audio-speech.vue` - settings page with model selector and speech playground
- `packages/i18n/src/locales/en/settings.yaml` - i18n entries
- `packages/stage-ui/src/stores/modules/speech.ts` - fix startup race condition that cleared persisted speech provider selection before validation ran

### DIFF
--- a/packages/i18n/src/locales/vi/tamagotchi/electron/tray.yaml
+++ b/packages/i18n/src/locales/vi/tamagotchi/electron/tray.yaml
@@ -21,7 +21,7 @@ menu:
       close_caption: Close Caption...
       caption_overlay: Lớp phủ phụ đề
       follow_window: Theo cửa sổ
-      reset_position: Đặt lại vị trí 
+      reset_position: Đặt lại vị trí
       devtools: Công cụ phát triển
       troubleshoot_beatsync: Sửa lỗi BeatSync...
       quit: Thoát ra
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1827,7 +1827,7 @@ importers:
         version: 5.0.0(vue@3.5.29(typescript@5.9.3))
       '@intlify/unplugin-vue-i18n':
         specifier: ^11.0.7
-        version: 11.0.7(@vue/compiler-dom@3.5.29)(eslint@9.39.3(jiti@2.6.1))(rollup@4.59.0)(typescript@5.9.3)(vite@8.0.0-beta.15(@types/node@24.10.14)(esbuild@0.27.2)(jiti@2.6.1)(less@4.5.1)(terser@5.46.0)(tsx@4.21.0)(yaml@2.8.2))(vue-i18n@11.2.8(vue@3.5.29(typescript@5.9.3)))(vue@3.5.29(typescript@5.9.3))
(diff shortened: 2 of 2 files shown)


packages/stage-ui/src/stores/providers.ts
```diff
@@ -1350,6 +1350,212 @@ export const useProvidersStore = defineStore('providers', () => {
...
+              if (!init?.body || typeof init.body !== 'string')
+                throw new Error('Invalid request body')
+
+              const body = JSON.parse(init.body)
+              const text = body.input
+              const voice = body.voice
+
+              const sseResponse = await globalThis.fetch(new URL('chat/completions', baseUrl), {
+                method: 'POST',
+                headers: {
```
shinohara-rin: need to extract this to somewhere else, this is too messy

monolithic827:
I've extracted the new OpenRouter TTS logic into its own module in order to de-clutter `providers.ts`.
Also addressed the two other comments by Gemini.

/gemini review

monolithic827:
Hi @shinohara-rin, sorry for the ping - wasn't sure you saw, I've addressed the comments and moved the logic out from `providers.ts` per your request. Would love to get this in 🙂

packages/stage-ui/src/stores/providers/openrouter/audio-speech.ts
```diff
@@ -0,0 +1,264 @@
...
+function decodeBase64PCM(chunks: string[]): Uint8Array {
+  const binaryString = atob(chunks.join(''))
+  const bytes = new Uint8Array(binaryString.length)
+  for (let i = 0; i < binaryString.length; i++)
+    bytes[i] = binaryString.charCodeAt(i)
+  return bytes
+}
+
+/** Wraps raw PCM16 mono data in a minimal WAV container. */
+function wrapPCM16InWAV(pcmBytes: Uint8Array, sampleRate = 24000): Uint8Array {
```
shinohara-rin, line 96: @nekomeowww do we want to extract this?

monolithic827:
> Would you mind declaring the coding agents used for this PR (if any) for our statistical purposes? Just reply to this.

Sure. I used Cursor with Opus 4.6 to boilerplate the changes, then adjusted the code until it was minimal and no longer contained AI slop. The smaller changes later were applied by hand.

PR #1304, opened 2026-03-12
fix(stage-ui): skip optimistic updates before apply

Gujiassh:
## Summary

- move the `skipActionIf()` guard ahead of `apply()` so skipped optimistic mutations do not leave phantom UI state behind
- add a focused regression in `packages/stage-ui/src/composables/use-optimistic.test.ts` that proves skipped mutations do not call `apply()` or `action()`
- keep the change localized to the shared stage-ui composable layer, away from the previously shipped Telegram message fixes

## Validation

- `pnpm -F @proj-airi/stage-ui exec vitest run src/composables/use-optimistic.test.ts`
- pre-commit hook ran `moeru-lint --fix` on the staged files successfully during commit

## Notes

- A root-level `pnpm lint:fix` attempt still reports unrelated pre-existing lint issues elsewhere in the monorepo, but the touched composable files passed the staged-file lint hook and the targeted test/typecheck flow

### DIFF
--- a/packages/i18n/src/locales/vi/tamagotchi/electron/tray.yaml
+++ b/packages/i18n/src/locales/vi/tamagotchi/electron/tray.yaml
@@ -20,7 +20,7 @@ menu:
       open_caption: Mở phụ đề...
       caption_overlay: Lớp phủ phụ đề
       follow_window: Theo cửa sổ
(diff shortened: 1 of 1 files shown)


Garfield550:
No place uses the useOptimisticMutation hook, but LGTM

PR #1312, opened 2026-03-12
fix(stage-ui): use stable chat history keys in the stage UI

### LINKED ISSUE #1267: bug: alpha.14 AIRI chat content disappears each time the software restarts 聊天内容每次重新启动软件后都会消失
### Describe the bug

(Continuation of https://github.com/moeru-ai/airi/issues/1206)

The issue still persists after I installed version alpha6...
在安装alpha6后还是有这个问题。。。

### System Info

```Shell
Windows 11 25H2 26200.7840
v0.9.0-alpha.6@03188f9855
2026-03-10T09:43:56.160Z
```

### Validations

### Contributions

stablegenius49:
Fixes #1267

## What changed
- stop keying rendered chat history rows by `createdAt`
- prefer each message's stable `id` instead
- fall back to a `role + createdAt + index` composite only when an id is missing
- add a tiny regression test for the new key helper

## Why
The chat history list was keyed by timestamp alone. When a user message and the assistant reply landed in the same millisecond, Vue could treat them as the same row and reuse/replace DOM incorrectly. That makes messages look like they disappeared even though they were still in session state / persisted storage.

## Verification
- `node --experimental-strip-types --test packages/stage-ui/src/components/scenarios/chat/message-key.test.ts`
- `git diff --check`

### DIFF
--- a/packages/stage-ui/src/components/scenarios/chat/message-key.test.ts
+++ b/packages/stage-ui/src/components/scenarios/chat/message-key.test.ts
@@ -1,6 +1,6 @@
-import assert from 'node:assert/strict'
-
 import type { ChatHistoryItem } from '../../../types/chat'
+
+import assert from 'node:assert/strict'
 
 import { describe, it } from 'vitest'
 
--- a/packages/stage-ui/src/stores/chat/session-message-merge.test.ts
+++ b/packages/stage-ui/src/stores/chat/session-message-merge.test.ts
@@ -1,6 +1,6 @@
-import assert from 'node:assert/strict'
-
 import type { ChatHistoryItem } from '../../types/chat'
+
+import assert from 'node:assert/strict'
 
 import { describe, it } from 'vitest'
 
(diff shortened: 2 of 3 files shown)


Garfield550:
Still not working; it looks like IndexedDB didn't save the message with `role: user`

Screenshot 2026-03-12 at 8 07 53 PM
Screenshot 2026-03-12 at 8 09 25 PM

stablegenius49:
Thanks — you were right, the keying change alone did not fix the underlying loss.

I pushed a follow-up that preserves in-flight local messages if IndexedDB finishes loading the session after a new send has already started. That avoids the stale stored snapshot clobbering the newly typed user message before the assistant reply is persisted.

Also added a regression test for the late-load merge path and refreshed the key helper tests.

Validation run:
- `pnpm exec vitest run packages/stage-ui/src/components/scenarios/chat/message-key.test.ts packages/stage-ui/src/stores/chat/session-message-merge.test.ts`

PR #1341, opened 2026-03-13
fix(stage-tamagotchi): fix model flickering and jumping on window resize

Reisenbug:
## Description
Fixed the Live2D model flickering/jumping when resizing the window, especially noticeable during slow vertical drags.
First, the Screen component had two competing watchers both updating canvasHeight. One using getBoundingClientRect(), the other useElementBounding. During resize they'd fire alternately with different values, making the canvas height oscillate. Replaced both watchers with a single computed property sourced from useElementBounding.
Second, Model.vue auto-repositioned and rescaled the model on every resize with a 100ms debounce. The Pixi renderer resizes immediately but the model position/scale lagged 100ms behind. Since users can already tweak position, scale, and offset in Settings, the auto-repositioning was unnecessary. Removed it.

## Additional Context
Is it more convenience to add a button to resize the model so user don't have to dive into expand->Open-Settings->Models?

### DIFF
--- a/packages/ui/src/components/layouts/screen.vue
+++ b/packages/ui/src/components/layouts/screen.vue
@@ -5,7 +5,7 @@ import { computed, onMounted, ref } from 'vue'
 const containerRef = ref<HTMLDivElement>()
 
 const breakpoints = useBreakpoints(breakpointsTailwind)
-const { width, height } = useWindowSize()
+const { width } = useWindowSize()
 const containerElementBounding = useElementBounding(containerRef, { immediate: true, windowResize: true, reset: true })
 
 const canvasWidth = computed(() => {


LemonNekoGH (commented):
Why?


https://github.com/user-attachments/assets/222db548-ac2e-43a7-9ee0-e47679c9cb7d

shinohara-rin:
Is #1345 related to this?

Reisenbug:
> Why?
> 
>  CleanShot.2026-03-14.at.03.09.02.mp4

What why, could you explain it more clearly
You can change the position and size of the model in expand->Open-Settings->Models to perfectly fit the window. However it is not very convenient...
Maybe we can add a button to turn on/off automatically change the model size or add a button to conveniently change the size and position

shinohara-rin:
> > Why?
> > CleanShot.2026-03-14.at.03.09.02.mp4
> 
> What why, could you explain it more clearly You can change the position and size of the model in expand->Open-Settings->Models to perfectly fit the window. However it is not very convenient... Maybe we can add a button to turn on/off automatically change the model size or add a button to conveniently change the size and position

I believe it makes more sense for the canvas coordinates to be centred to the window

Reisenbug:
> Is #1345 related to this?

I'm a macOS user and I can't reproduce this exact issue. However, when resizing the window by dragging from the lower-left, upper-left, or upper-right corner, the bottom and right edges vibrate slightly

Reisenbug:
> > > Why?
> > > CleanShot.2026-03-14.at.03.09.02.mp4
> > 
> > 
> > What why, could you explain it more clearly You can change the position and size of the model in expand->Open-Settings->Models to perfectly fit the window. However it is not very convenient... Maybe we can add a button to turn on/off automatically change the model size or add a button to conveniently change the size and position
> 
> I believe it makes more sense for the canvas coordinates to be centred to the window

It makes sense. I'll change this later

Reisenbug:
https://github.com/user-attachments/assets/a7835d93-a229-4215-a023-a3a3848ef638

What about now

LemonNekoGH (approved):

LemonNekoGH:
Thanks it looks more concise now, let me merge it.

LemonNekoGH:
CI check failed, should I wait for you to fix it before merge?

Reisenbug:
> CI check failed, should I wait for you to fix it before merge?

Fixed

nekomeowww:
I cannot agree with the auto-repositioning was unnecessary argument.

The implementation seems fixed the issue of "non-lerp-ed transition" for position, we could fix this by using animejs to help to animate this, but still the PR description and the way to fix this wasn't even aligned. Why would the change to https://github.com/moeru-ai/airi/pull/1341/changes#diff-c347dabdb53fcd3344f7b587ea5222bf59bf71be7e5a7c36bc053ee4625f4cd6L123-L140 have to do with the auto-repositioning and jumppy update?

Rewrite to `Screen.vue` to remove the un-implemented branch for tablet, desktop sizes were reasonable, but still the Live2D impl with initial scale should remain unchanged. 

Before the impl, the model could be scaled automatically to seems-to-be-like-VTuber size:

image

With the implementation here later, the model will have this scale:

image

And the behavior change wasn't mentioned.

nekomeowww:
I reverted this PR, we could find better version of impl to fix this.

Reisenbug:
> I reverted this PR, we could find better version of impl to fix this.

Simplify Screen.vue canvas dimension calculation to a single computed
Fix jumping using animejs to lerp model position/scale on resize
Is this ok

PR #1378, opened 2026-03-15
feat(providers): add Z.ai provider

AdairLi2504:
## Description

Add support for [Z.ai](z.ai) as an LLM provider.


nix/assets-hash.txt
shinohara-rin, line 1: I think you accidentally deleted this...

PR #1395, opened 2026-03-16
fix(stage-ui-live2d): animate model position and scale on resize with animejs

Reisenbug:
## Description
Related to #1341 
using animejs to fix the issue of "non-lerp-ed transition" for position


nekomeowww (approved):

nekomeowww:
Thanks, this version perfectly fixed this.

PR #1401, opened 2026-03-16
Integrate with QQ chat software so that it can chat on QQ

aibijiaop-commits:
Currently provides QQ message sending and receiving. When the voice module is configured, you can choose whether to send voice messages. The model context is shared with AIRI's built-in chat (shared memory).

Currently providing official QQBot methods and custom WS connection methods, more functions are under development.


shinohara-rin:
image

image

please clean up, or i'm going to sue you for emotional damage/j

aibijiaop-commits:
Packed the extra files, will pay attention next time, hahaha

PR #1408, opened 2026-03-17
Integrate with QQ chat software

aibijiaop-commits:
This is a new method to dock with QQ chat software, 
providing both the official method and a custom WS address method. It's also a matter of taking time to re-upload a copy (ahaha, 
currently implementing message sending and receiving memory synchronized with AIRI, when configuring voice you can choose to send voice,


shinohara-rin:
first this: 
image

then this:
image

sorry to be mean but bro you should really learn some basic git 101, this is way too hilarious i'm not even mad

PR #1423, opened 2026-03-18
feat(plugin-sdk): much better refactor, and new permission model

nekomeowww:
## Description

Follow up of #255

### DIFF
--- a/apps/stage-tamagotchi/src/main/services/airi/plugins/examples/devtools-sample-plugin/README.md
+++ b/apps/stage-tamagotchi/src/main/services/airi/plugins/examples/devtools-sample-plugin/README.md
@@ -7,6 +7,8 @@ This sample plugin is for validating plugin host behavior in the **Plugin Host I
 - `devtools-sample-plugin.json`: plugin manifest (`ManifestV1`)
 - `devtools-sample-plugin.mjs`: plugin implementation
 
+The manifest declares the protocol permissions required by `apis.providers.listProviders()`: invoke `capabilities:wait`, invoke `resources:providers:list-providers`, read the provider resource, and wait for the provider-list capability.
+
 ## How to use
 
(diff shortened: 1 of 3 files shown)


packages/plugin-sdk/src/plugin-host/shared/types.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export type ModulePermissionDeclaration = ProtocolModulePermissionDeclaration
+
+export type ModulePermissionGrant = ProtocolModulePermissionGrant
+
+export interface ManifestV1 {
+  apiVersion: 'v1'
+  kind: 'manifest.plugin.airi.moeru.ai'
+  name: string
+  permissions: ModulePermissionDeclaration
```
sumimakito: Just a nit: We can keep this aligned with the schema.

```suggestion
  permissions?: ModulePermissionDeclaration
```

packages/plugin-sdk/src/plugin-host/runtimes/node/loaders/fs.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (!entrypoint) {
+      throw new Error(''
+        + `Plugin entrypoint is required for runtime \`${runtime}\`. `
+        + 'Define one of `entrypoints.<runtime>`, `entrypoints.default`, '
+        + 'or `entrypoints.electron` in the plugin manifest.',
+      )
+    }
+
+    return join(root, entrypoint)
```
sumimakito: Maybe we want to check for absolute paths as well?

```suggestion
    return isAbsolute(entrypoint) ? entrypoint : join(root, entrypoint)
```

**NOTE:** We need to import `isAbsolute`.

sumimakito (approved):

PR #1445, opened 2026-03-21
fix(server-runtime): use constant-time comparison for auth token

### LINKED ISSUE #1444: Security: Auth token comparison vulnerable to timing attack (CWE-208)
## Summary

The WebSocket authentication handler in `packages/server-runtime/src/index.ts` uses direct string comparison (`!==`) to validate authentication tokens:

```typescript
if (authToken && event.data.token !== authToken) {
```

This is vulnerable to timing-based side-channel attacks ([CWE-208](https://cwe.mitre.org/data/definitions/208.html)). The `!==` operator exits early on the first mismatched character, meaning the response time leaks information about how many leading characters of the token are correct. An attacker can exploit this to guess the token one byte at a time.

## Recommendation

Replace with Node.js `crypto.timingSafeEqual()`, which compares all bytes in constant time regardless of content:

```typescript
import { timingSafeEqual } from 'node:crypto';

const bufA = Buffer.from(clientToken);
const bufB = Buffer.from(authToken);
if (bufA.length !== bufB.length || !timingSafeEqual(bufA, bufB)) {
  // reject
}
```

## Impact

- **Severity**: Medium
- **Attack vector**: Network (attacker sends crafted tokens and measures response timing)
- **Prerequisite**: `AUTHENTICATION_TOKEN` must be configured (auth enabled)

## References

- [CWE-208: Observable Timing Discrepancy](https://cwe.mitre.org/data/definitions/208.html)
- [Node.js crypto.timingSafeEqual](https://nodejs.org/api/crypto.html#cryptotimingsafeequala-b)

spidershield-contrib:
## Summary

- Replace direct string comparison (`!==`) with `crypto.timingSafeEqual()` in the WebSocket authentication handler to prevent timing-based side-channel attacks ([CWE-208](https://cwe.mitre.org/data/definitions/208.html))
- Add a `timingSafeCompare` helper that handles variable-length strings safely (compares against itself on length mismatch to avoid leaking length information)
- Add `import { timingSafeEqual } from 'node:crypto'`

Fixes #1444

## Details

The `event.data.token !== authToken` comparison in `packages/server-runtime/src/index.ts` (line 342) exits early on the first mismatched byte. An attacker observing response latency can incrementally guess the token character by character.

`crypto.timingSafeEqual()` always compares all bytes in constant time, eliminating the timing oracle.

## Test plan

- [ ] Verify WebSocket authentication still works with correct token
- [ ] Verify authentication is rejected with incorrect token
- [ ] Verify server works without auth token configured (no-auth mode)
---
*Found by [SpiderShield](https://github.com/teehooai/spidershield) security scanner*

### DIFF
--- a/packages/server-runtime/src/index.ts
+++ b/packages/server-runtime/src/index.ts
@@ -30,15 +30,25 @@ import {
   matchesDestinations,
 } from './middlewares'
 
-/** Constant-time string comparison that prevents timing attacks (CWE-208). */
+/**
+ * Constant-time string comparison that prevents timing attacks (CWE-208).
+ *
+ * @param {string} a - the first string to compare
+ * @param {string} b - the expected value (e.g., the real secret)
+ * @returns {boolean} `true` if the strings are equal, `false` otherwise
+ */
 function timingSafeCompare(a: string, b: string): boolean {
   const bufA = Buffer.from(a)
   const bufB = Buffer.from(b)
   if (bufA.length !== bufB.length) {
     // Compare against itself to keep constant time, then return false
     timingSafeEqual(bufA, bufA)
+    // To prevent leaking length information, we perform a dummy comparison on the
+    // expected value, making the execution time dependent on its length.
+    timingSafeEqual(bufB, bufB)
     return false
   }
+
   return timingSafeEqual(bufA, bufB)
 }
 


packages/server-runtime/src/index.ts
```diff
@@ -1,4 +1,17 @@
 import type { MetadataEventSource, WebSocketEvent } from '@proj-airi/server-shared/types'
+import { timingSafeEqual } from 'node:crypto'
+
+/** Constant-time string comparison that prevents timing attacks (CWE-208). */
+function timingSafeCompare(a: string, b: string): boolean {
+  const bufA = Buffer.from(a)
+  const bufB = Buffer.from(b)
+  if (bufA.length !== bufB.length) {
+    // Compare against itself to keep constant time, then return false
+    timingSafeEqual(bufA, bufA)
```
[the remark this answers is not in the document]
Zazzik1: agree with gemini review here, if we compared the expected value (`bufB`) against itself we would have constant execution time regardless what parameters we pass to the `timingSafeCompare` fn (if we assume that b is the secret)

packages/server-runtime/src/index.ts
```diff
@@ -1,4 +1,17 @@
 import type { MetadataEventSource, WebSocketEvent } from '@proj-airi/server-shared/types'
+import { timingSafeEqual } from 'node:crypto'
+
+/** Constant-time string comparison that prevents timing attacks (CWE-208). */
```
Zazzik1: It might be helpful to note in this comment that the b parameter represents the expected value (e.g., the real secret), so developers and future reviewers know how to use it safely. What do you think?


```suggestion
/**
 * Constant-time string comparison that prevents timing attacks (CWE-208).
 *
 * @param {string} a - the first string to compare
 * @param {string} b - the expected value (e.g., the real secret)
 * @returns {boolean} `true` if the strings are equal, `false` otherwise
 */
```

packages/server-runtime/src/index.ts
```diff
@@ -28,6 +30,18 @@ import {
   matchesDestinations,
 } from './middlewares'
 
+/** Constant-time string comparison that prevents timing attacks (CWE-208). */
+function timingSafeCompare(a: string, b: string): boolean {
+  const bufA = Buffer.from(a)
+  const bufB = Buffer.from(b)
+  if (bufA.length !== bufB.length) {
+    // Compare against itself to keep constant time, then return false
+    timingSafeEqual(bufA, bufA)
```
nekomeowww, line 45:
```suggestion
    // Compare against itself to keep constant time, then return false
    timingSafeEqual(bufA, bufA)
    // To prevent leaking length information, we perform a dummy comparison on the
    // expected value, making the execution time dependent on its length.
    timingSafeEqual(bufB, bufB)
```

packages/server-runtime/src/index.ts
```diff
@@ -28,6 +30,18 @@ import {
   matchesDestinations,
 } from './middlewares'
 
+/** Constant-time string comparison that prevents timing attacks (CWE-208). */
```
nekomeowww:
```suggestion
/**
 * Constant-time string comparison that prevents timing attacks (CWE-208).
 *
 * @param {string} a - the first string to compare
 * @param {string} b - the expected value (e.g., the real secret)
 * @returns {boolean} `true` if the strings are equal, `false` otherwise
 */
```

packages/server-runtime/src/index.ts
```diff
@@ -28,6 +30,18 @@ import {
...
+function timingSafeCompare(a: string, b: string): boolean {
+  const bufA = Buffer.from(a)
+  const bufB = Buffer.from(b)
+  if (bufA.length !== bufB.length) {
+    // Compare against itself to keep constant time, then return false
+    timingSafeEqual(bufA, bufA)
+    return false
+  }
```
nekomeowww, line 50:
```suggestion
  }

```

PR #1468, opened 2026-03-24
fix(stage-ui): tools not available when using remote API

BeanDz:


packages/stage-ui/src/stores/llm.test.ts
```diff
@@ -1,37 +1,49 @@
...
-import { attemptForToolsCompatibilityDiscovery } from './llm'
-
-function doesHaveOpenRouterApiKey() {
-  const apiKey = env.LLM_API_OPENROUTER_API_KEY
-  if (!apiKey) {
-    console.warn('Skipping llm store tests, because LLM_API_OPENROUTER_API_KEY is not set')
+import { isToolRelatedError } from './llm'
+
+describe('isToolRelatedError', () => {
+  const positives: [provider: string, msg: string][] = [
```
nekomeowww, line 6: Epic!

PR #1492, opened 2026-03-27
fix(stage-pages): XSS vulnerability in provider definition

shinohara-rin:
🛡️ Sentinel: [HIGH] Remove unnecessary v-html in provider definition

**Vulnerability:**
The `packages/stage-pages` module was using the `v-html` directive to inject `providerDefinition` content for troubleshooting into the DOM. Since this content is actually plain text defined in the i18n configurations, using `v-html` exposes the application to unnecessary XSS vectors if the definitions dynamically include untrusted elements in the future.

**Fix:**
Replaced the `v-html` usage entirely with standard Vue template text interpolation (`{{ }}`). This ensures Vue strictly treats the content as plain text, eliminating the XSS vector by design.

**Sentinel Journal:**
Logged a learning in `.jules/sentinel.md` emphasizing that we should actively question if `v-html` is strictly necessary and strongly prefer text interpolation (`{{ }}`) by default for rendering dynamic configuration and translations.

---
*PR created automatically by Jules for task [11374405078978691146](https://jules.google.com/task/11374405078978691146) started by @shinohara-rin*


shinohara-rin:
This might break rendering if we're putting html in i18n

although i don't see why we'd need one to begin with

packages/stage-pages/src/pages/v2/settings/providers/edit/[providerId]/index.vue
```diff
@@ -3,6 +3,8 @@ import type { ProviderValidationStep } from '@proj-airi/stage-ui/libs'
 import type { ZodType } from 'zod'
 import type { $ZodType } from 'zod/v4/core'
 
+import DOMPurify from 'dompurify'
```
Comment (nekomeowww) [inline packages/stage-pages/src/pages/v2/settings/providers/edit/[providerId]/index.vue:7]:
```suggestion
// TODO: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API
import DOMPurify from 'dompurify'
```

PR #1521, opened 2026-03-30
fix(server-runtime): preserve explicit empty route destinations

Gujiassh:
## Summary
- treat an explicit empty `route.destinations` array as an intentional override instead of falling back to `data.destinations`
- keep the server-runtime route middleware aligned with the caller's explicit routing decision
- add focused middleware coverage for the empty-destinations override case

## End-user benefit
When a caller explicitly sets `route.destinations: []`, the runtime now preserves that choice instead of unexpectedly reusing destinations from the event payload. That prevents messages from being routed to stale targets when the sender is intentionally clearing the destination list.

## Testing
- `pnpm -F @proj-airi/server-runtime exec vitest run src/middlewares/route.test.ts`
- `pnpm -F @proj-airi/server-runtime typecheck`

### DIFF
--- a/packages/server-runtime/src/middlewares/route.ts
+++ b/packages/server-runtime/src/middlewares/route.ts
@@ -79,7 +79,7 @@ export function createPolicyMiddleware(policy: RoutingPolicy): RouteMiddleware {
 }
 
 export function collectDestinations(event: WebSocketEvent | (Omit<WebSocketEvent, 'metadata'> & Partial<Pick<WebSocketEvent, 'metadata'>>)) {
-  if (event.route?.destinations !== undefined) {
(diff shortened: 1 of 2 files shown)


Gujiassh:
Follow-up note: this branch was rechecked locally against current main. The exact two-file diff in `packages/server-runtime/src/middlewares/route.ts` and `route.test.ts` still reproduces the intended fix, and the focused validation set passed locally: `pnpm -F @proj-airi/server-runtime exec vitest run src/middlewares/route.test.ts` and `pnpm -F @proj-airi/server-runtime typecheck`.

nekomeowww (approved):

PR #1527, opened 2026-03-30
QoL/Fixes to updater

jensenhuangfan:
This PR stabilizes the Airi auto-updater system by resolving metadata fetch errors and implementing a native cleanup mechanism to reclaim disk space.

Key Fixes and Improvements:

Switched the update provider from github to generic in app-update.yml.
Pointed the update URL directly to https://github.com/moeru-ai/airi/releases/latest/download/. This allows the updater to resolve metadata for pre-release tags (e.g., v0.9.0-alpha.36) without hitting the aggressive GitHub API tag filter that was causing update failures.

Automated Updater Cache Purge:

Implemented a startup cleanup routine in auto-updater.ts that automatically deletes the ai.moeru.airi-updater directory.
This resolves the issue where the application was accumulating hundreds of megabytes of redundant installer files (.exe, .dmg, .AppImage) in the user's Local AppData/Cache folder.
Cross-Platform Resolution Support:

Refactored the cache-path resolution to be fully platform-agnostic.
The cleanup script now correctly identifies the cache directory for Windows (%LOCALAPPDATA%), macOS (~/Library/Caches), and Linux (~/.cache), ensuring consistency for all users.
Removed Conflicting Runtime Overrides:

Removed codebase-level autoUpdater.channel hardcodes that were causing double platform suffixes (e.g., latest-x64-x64.yml), allowing the native electron-updater build-time configuration to handle channel expansion correctly.

Linked Issues
#1530

Additional Context
Reviewers should focus on the getAppCacheDir() helper in auto-updater.ts, which I've implemented using native Node APIs (node:os, node:process) to maintain high performance and broad compatibility without adding external dependencies.

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/pages/settings/data/components/desktop-folder-section.vue
+++ b/apps/stage-tamagotchi/src/renderer/pages/settings/data/components/desktop-folder-section.vue
@@ -31,13 +31,13 @@ async function triggerOpenDesktopUserDataFolder() {
 </script>
 
 <template>
-  <div :class="['border-2 border-sky-200/80 rounded-xl bg-sky-50/80 p-4 shadow-sm', 'dark:border-sky-500/40 dark:bg-sky-500/10']">
+  <div :class="['border-2 border-neutral-200/50 rounded-xl bg-white/70 p-4 shadow-sm', 'dark:border-neutral-800/60 dark:bg-neutral-900/60']">
     <div :class="['grid grid-cols-1 items-start gap-3 md:grid-cols-[minmax(0,1fr)_auto]']">
       <div :class="['flex flex-col gap-1 md:max-w-[560px]']">
-        <div :class="['text-lg text-sky-700 font-medium dark:text-sky-200']">
+        <div :class="['text-lg font-medium']">
           {{ t('settings.pages.data.sections.desktop-folder.title') }}
         </div>
-        <p :class="['text-sm text-sky-700/80 dark:text-sky-200/80']">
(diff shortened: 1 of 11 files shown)


jensenhuangfan:
the patch I made in 34 was incomplete because I thought it couldn't get updates because it had no more updates at the time but the updater was just broken. after testing (on windows) It seems like I fixed it and made it so it should erase the update files after a successful update (hopefully)

I got it to update in the background after clicking the update button and it was smooth and seemless but after some work it started to detect the code below

nsis: {
    oneClick: false, // forces the visual setup wizard
}
is the updater supposed to do this or first time setup. I can fix this part when I know but the other fixes to the updater should work now

jensenhuangfan:
If there's any issues with my work, please let me know so I can learn from them

nekomeowww:
Incorrect implementation. 

Let's focus your previous PR here: https://github.com/moeru-ai/airi/commit/ee538769f976123193992219f29a85b370cdc127 the channel was switched to latest-arm64 or latest-x64, this will by default break the auto upgrading mechanism.

jensenhuangfan:
> Incorrect implementation.
> 
> Let's focus your previous PR here: [ee53876](https://github.com/moeru-ai/airi/commit/ee538769f976123193992219f29a85b370cdc127) the channel was switched to latest-arm64 or latest-x64, this will by default break the auto upgrading mechanism.

None of my issues fixed in the build, just the one on my pc which was weird so I'm gonna work on a clean slate

PR #1533, opened 2026-03-31
chore(tauri-*,crates): now maintaining tauri and rust packages in another repo

nekomeowww:


nekomeowww:
> LGTM, do we need a script to bump versions in the separate repo?

No. We will maintain another CI/CD versioning for https://github.com/moeru-ai/tauri-plugins-airi

Issue #1547, opened 2026-04-02
Updater not working.


jensenhuangfan:
### Describe the bug

A lot of details of my fix in the PR.

<img width="615" height="635" alt="Image" src="https://github.com/user-attachments/assets/4878ef34-192d-4405-b6ce-8340a9534775" />
the screenshot is the updater working on the alpha before my first/unsuccessful patch was distributed (but i did hotfix patch [8567b702] so i could test with it)

<img width="732" height="434" alt="Image" src="https://github.com/user-attachments/assets/55787af6-b754-4b00-8685-9e384ec0c5c0" />

this screenshot is it working in the newest version that was built from commit [631a6ba] (I redacted my name from the user directoy of my PC.)

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (20) x64 Intel(R) Core(TM) Ultra 7 265K
    Memory: 39.11 GB / 63.47 GB
  Binaries:
    Node: 24.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 11.9.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.32.1 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Chrome: 146.0.7680.165
    Edge: Chromium (127.0.2651.74)
    Firefox: 149.0 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.8115
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [x] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

nekomeowww:
Should be fixed.

Issue #1530, opened 2026-03-31
Updater Error: No published versions on GitHub


d1tokirb:
### Describe the bug

<img width="1350" height="877" alt="Image" src="https://github.com/user-attachments/assets/4a6c5e62-de2e-446d-86b2-642232f5cce0" /> My best friend fixed this issue in his PR but I just wanted to make this issue to show that the issue exists for other people.

### System Info

```Shell
System:
    OS: macOS 26.4
    CPU: (10) arm64 Apple M4
    Memory: 396.45 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.14.0 - /usr/local/bin/node
    npm: 11.9.0 - /usr/local/bin/npm
    pnpm: 10.28.2 - /opt/homebrew/bin/pnpm
    Watchman: 2025.11.10.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 146.0.7680.165
    Safari: 26.4
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

nekomeowww:
Intended. Manual upgrade is needed.

Issue #1428, opened 2026-03-19
Bug: auto updater on mac do not pick the right cpu architecture


jmaitrehenry:
### Describe the bug

I'm on Mac M1 and the updater install the Intel version of the project instead of the ARM one.
I need to redownload and reinstall the correct version everytime.

### System Info

```Shell
MacbookPro M1 (ARM)
Airi 0.9+
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

JasonOA888:
I analyzed this issue. The problem appears to be in how electron-updater selects the correct architecture from GitHub releases.

**Release artifacts observed:**
```
AIRI-0.9.0-alpha.18-arm64-mac.zip
AIRI-0.9.0-alpha.18-darwin-arm64.dmg
AIRI-0.9.0-alpha.18-darwin-x64.dmg
```

**Possible causes:**

1. **electron-updater URL pattern mismatch** - The auto-updater may not be correctly detecting `process.arch` (arm64) when fetching updates

2. **Missing `autoUpdateArch` in latest.yml** - The `latest-mac.yml` file on GitHub releases should include architecture info

3. **NSIS/DMG naming inconsistency** - Different naming patterns between `arm64-mac.zip` and `darwin-arm64.dmg`

**Debugging steps:**

1. Check what `autoUpdater.getFeedURL()` returns on M1:
   ```js
   console.log(autoUpdater.getFeedURL())
   ```

2. Check if `latest-mac.yml` exists and contains:
   ```yaml
   - url: AIRI-0.9.0-alpha.18-arm64-mac.zip
     sha512: ...
   ```

3. Try forcing architecture in electron-builder.config.ts:
   ```ts
   publish: {
     provider: 'github',
     owner: 'moeru-ai',
     repo: 'airi',
     // Add explicit channel for architecture
   }
   ```

[...]

PR #1548, opened 2026-04-02
fix(stage-tamagotchi): harden cross-platform auto-updater flow, diagnostics, logs, and cache cleanup

jensenhuangfan:
## Description

This PR improves the desktop updater end-to-end for Windows, macOS, and Linux, with a focus on architecture-correct update resolution, better reliability, and better debugging visibility.

## Linked Issues
#1530 #1547 #1428 (validated on Windows; macOS/Linux architecture selection logic implemented, but native runtime verification is still needed).

Ensures updater channel/feed selection is architecture- and platform-aware (including macOS and Linux arm64/x64 split)..

Improves update install behavior:
Windows keeps silent install + relaunch behavior.
macOS/Linux use safe default non-Windows updater install flow.

Cleans updater cache artifacts more thoroughly:
pending update directory cleanup
root installer artifact cleanup (including installer executables where present to save a lot of storage after installs)

Adds richer diagnostics from main process into updater state/UI:
platform/arch
channel
feed URL
cache/pending paths
exact updater log path
executable/uninstall path diagnostics

Improves About page update UX:
explicit “Up to date” state in green
allows manual re-check even when up to date
platform-specific install messaging (Windows silent wording vs macOS/Linux restart-install wording since at least to my knowledge silent updates with autorestart for this is mainly for windows. but if I missed something I'm happy to learn from my mistakes)

auto-updater.ts electron-builder.config.ts about.vue and index.ts and some of the new QoL features like logging.

Validation notes:

Verified build/runtime behavior in Windows environment.
macOS/Linux logic is implemented and platform-aware; HOWEVER final runtime confirmation requires native macOS/Linux test runs. (any feedback will be useful)


jensenhuangfan:
im looking into the stuff that the AIs are suggesting but some of the suggestions break things like pre release updates for example

apps/stage-tamagotchi/src/main/services/electron/auto-updater.ts
```diff
@@ -55,35 +85,161 @@ export interface AutoUpdater {
...
+    if (!response.ok)
+      throw new Error(`Failed to query releases API (${response.status})`)
+
+    const releases = await response.json() as Array<{ tag_name?: string, draft?: boolean, prerelease?: boolean }>
+    const latestRelease = releases.find(release => !!release.tag_name && !release.draft && !!release.prerelease === isPrereleaseBuild)
+      ?? releases.find(release => !!release.tag_name && !release.draft)
+    if (!latestRelease?.tag_name)
+      throw new Error('No published versions on GitHub')
+
+    const feedUrl = `https://github.com/moeru-ai/airi/releases/download/${latestRelease.tag_name}`
```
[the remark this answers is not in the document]
jensenhuangfan: i would assume that the tradeoff would be that it won't work for pre-release builds. i could be wrong but changing something to do with how it connected to github allowed all versions to work. if someone corrects me on this/finds a workaround then great but IDK if this is possible.

jensenhuangfan:
If their is any issue with this PR that needs fixed before it gets merged please let me know if possible

apps/stage-tamagotchi/electron-builder.config.ts
```diff
@@ -33,11 +33,7 @@ if (!useIconFormattedMacAppIcon) {
   console.warn('[electron-builder/config] Warning: Xcode version is below 26. Using .icns format for macOS app icon.')
 }
 else {
-  // NOTICE: This success-path message intentionally uses stderr via `console.warn`.
-  // The artifact metadata CLI imports this config and is used in GitHub Actions
-  // command substitution for `GITHUB_ENV`; writing this log to stdout would break
-  // machine-readable output such as `BUNDLE_NAME=$(...)`.
-  console.warn('[electron-builder/config] Xcode version is 26 or above. Using .icon format for macOS app icon.')
+  console.info('[electron-builder/config] Xcode version is 26 or above. Using .icon format for macOS app icon.')
```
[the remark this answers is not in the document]
jensenhuangfan: should be reverted

apps/stage-tamagotchi/src/main/services/electron/auto-updater.ts
```diff
@@ -55,35 +85,161 @@ export interface AutoUpdater {
...
+      channel: autoUpdater.channel || getReleaseChannelName(),
+      feedUrl: resolvedFeedUrl,
+      logFilePath: UPDATER_LOG_FILE,
+      updaterCacheDir: OFFICIAL_UPDATER_CACHE_DIR,
+      pendingDir: OFFICIAL_UPDATER_PENDING_DIR,
+      executablePath,
+      uninstallPath,
+      uninstallExists: uninstallPath ? fs.existsSync(uninstallPath) : undefined,
+    }
+  }
```
nekomeowww, line 118: Do we need this?

apps/stage-tamagotchi/src/main/services/electron/auto-updater.ts
```diff
@@ -55,35 +85,161 @@ export interface AutoUpdater {
...
+      return
+
+    const archChannel = getReleaseChannelName()
+    const releasesApi = 'https://api.github.com/repos/moeru-ai/airi/releases?per_page=20'
+    const response = await fetch(releasesApi, {
+      headers: {
+        'accept': 'application/vnd.github+json',
+        'User-Agent': 'airi-updater',
+      },
+    })
```
nekomeowww, line 183: Must we re-implement this by ourselves?

apps/stage-tamagotchi/electron-builder.config.ts
```diff
@@ -151,72 +165,5 @@ export default {
     artifactName: '${productName}-${version}-linux-${arch}.${ext}',
   },
   npmRebuild: false,
-  publish: {
-    provider: 'github',
```
nekomeowww, line 155: Why many comments were deleted

jensenhuangfan: I'm so sorry about that. I was grinding this update until 2 AM in order to fix this, so without knowing, I might've erased some comments, but I went back and looked at what was gone and put them back in my pull request.

apps/stage-tamagotchi/src/main/services/electron/auto-updater.ts
jensenhuangfan: No. We still rely on electron-updater for check/download/install. This patch only adds deterministic release selection (stable vs prerelease) and explicit feed/channel resolution so electron-updater reads the correct metadata every time.

apps/stage-tamagotchi/src/main/services/electron/auto-updater.ts
jensenhuangfan: Yes. Without this, channel/metadata resolution can drift across track or arch and produce incorrect update lookups. With arch-aware channel names plus explicit feed pinning, update resolution is predictable across platforms and much easier to debug in production if anything did fail later.

apps/stage-tamagotchi/src/main/services/electron/auto-updater.ts
```diff
@@ -22,12 +24,40 @@ import {
 } from '../../../shared/eventa'
 import { MockAutoUpdater } from './mock-auto-updater'
 
+function getCacheRoot() {
+  switch (process.platform) {
+    case 'win32':
+      return process.env.LOCALAPPDATA || join(process.env.USERPROFILE || '', 'AppData', 'Local')
```
nekomeowww, line 30: May we switch to use `app.getPath(name)` with `appData`, https://www.electronjs.org/docs/latest/api/app

apps/stage-tamagotchi/src/main/services/electron/auto-updater.ts
```diff
@@ -22,12 +24,40 @@ import {
...
+    case 'darwin':
+      return join(process.env.HOME || '', 'Library', 'Caches')
+    default:
+      return process.env.XDG_CACHE_HOME || join(process.env.HOME || '', '.cache')
```
nekomeowww, line 34: Here could be switched to use the `app.getPath(name)` with `appData`, https://www.electronjs.org/docs/latest/api/app

PR #1593, opened 2026-04-07
fix(stage-ui): improve LLM KV-cache compatibility for datetime context injection

### LINKED ISSUE #1539: airi:system:datetime plugin injection breaks LLM KV Cache (Severe TTFT degradation)
### Describe the bug

I’m opening this issue to report a severe performance bottleneck affecting multi-turn, long-context conversations, especially for users running local backends. The current implementation of the datetime plugin is unintentionally destroying the LLM's KV Cache, forcing the backend to recompute tens of thousands of tokens on every single turn.

### 📌 Environment & Symptoms

* **Backend:** `llama.cpp` + `Qwen3.5-27B` (Local deployment on my Macbook M5 Pro 48GB)
* **Context Size:** ~14k+ tokens of roleplay history.
* **Issue:** The Time To First Token (TTFT) is extremely slow. Looking at the `llama.cpp` logs, the engine constantly throws: 
  `forcing full prompt re-processing due to lack of cache data`
  The `n_past` variable only hits around ~1,800 tokens (exactly the length of the System Prompt). The remaining 10,000+ tokens of static chat history are discarded and fully recomputed every single time.

### 🔍 Root Cause Analysis

To debug this, I exported the raw prompt payloads from two consecutive turns and wrote a Python script to tokenize and compare them token-by-token. 

The divergence point happens exactly where the `airi:system:datetime` plugin injects its block, right after the System Prompt:

**Turn 1:**
`[{"id":"dcj1QnDKqs_BiOsuszhWe","contextId":"system:datetime","strategy":"replace-self","text":"Current datetime...`

**Turn 2:**
`[{"id":"S1WuLr2ZVIti7s_lw1QiP","contextId":"system:datetime","strategy":"replace-self","text":"Current datetime...`

LLM KV caching relies strictly on **Longest Common Prefix matching**. Because Airi injects a newly generated random UUID (`id`) and a millisecond-precision timestamp (`createdAt`) near the very beginning of the prompt, the prefix match breaks immediately.

Since the prefix breaks at token ~1,200, the LLM assumes the rest of the prompt is entirely new and flushes the cache for the remaining 10k+ tokens of unchanged chat history. This causes a massive and unnecessary computational overhead.

### 💡 Proposed Solutions

To make Airi more cache-friendly for local deployments and API users alike, I suggest one of the following approaches:

1. **Move the Injection Point:** Instead of injecting the datetime block right after the System Prompt, append it to the *very end* of the message array (just before the newest user message). This allows 95% of the static background and chat history to perfectly hit the KV cache, turning a minutes-long wait into a seconds-long response.
2. **Add a Configuration Toggle / Time Quantization**: Provide an option to completely disable datetime injection for users who do not require their agents to be strictly aware of real-world time. Alternatively, consider implementing Time Quantization (e.g., updating the timestamp only once per hour or day) to preserve essential temporal context while drastically reducing the frequency of cache invalidation.

Implementing this would be a massive quality-of-life upgrade for anyone doing heavy roleplay or long-context tasks. Thanks for the great work on Airi!

### System Info

```Shell
System:
    OS: macOS 26.4
    CPU: (15) arm64 Apple M5 Pro
    Memory: 667.61 MB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.8.2 - /opt/homebrew/bin/node
    npm: 11.11.1 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 146.0.7680.165
    Safari: 26.4
```

### Validations

### Contributions

luoling8192:
## Summary

Fixes #1539 — the `airi:system:datetime` plugin was breaking LLM KV-cache prefix matching on every turn, causing severe TTFT degradation for local backends.

**Root cause**: A random `nanoid()` ID and millisecond-precision `createdAt` timestamp were serialized into the prompt right after the system message, invalidating the entire KV-cache prefix on every request.

**Changes**:
- **Move context injection to end of messages** (`chat.ts`) — context block is now inserted just before the latest user message instead of after the system prompt, so static chat history stays prefix-matched
- **Whitelist serialized fields** (`context-prompt.ts`) — only `contextId`, `strategy`, and `text` are included in the prompt; volatile `id`, `createdAt`, and `metadata` are excluded
- **Quantize datetime to the minute** (`datetime.ts`) — seconds and milliseconds are zeroed, so calls within the same minute produce identical output
- **Use fixed deterministic ID** (`datetime.ts`) — replace `nanoid()` with a stable singleton ID

## Test plan

- [x] Added unit tests for `createDatetimeContext()` — fixed ID, minute quantization, same-minute stability, cross-minute difference
- [x] Added unit tests for `formatContextPromptText()` — volatile field exclusion, whitelist-only serialization, identical output regardless of volatile fields
- [x] All 10 tests passing via `pnpm -F @proj-airi/stage-ui exec vitest run`

### DIFF
--- a/packages/stage-ui/src/stores/chat.ts
+++ b/packages/stage-ui/src/stores/chat.ts
@@ -13,7 +13,7 @@ import { ref, toRaw } from 'vue'
 import { useAnalytics } from '../composables'
 import { useLlmmarkerParser } from '../composables/llm-marker-parser'
 import { categorizeResponse, createStreamingCategorizer } from '../composables/response-categoriser'
-import { buildContextPromptMessage } from './chat/context-prompt'
+import { formatContextPromptText } from './chat/context-prompt'
 import { createDatetimeContext, createMinecraftContext } from './chat/context-providers'
 import { useChatContextStore } from './chat/context-store'
 import { createChatHooks } from './chat/hooks'
@@ -279,7 +279,7 @@ export const useChatOrchestratorStore = defineStore('chat-orchestrator', () => {
         ],
       })
 
-      let newMessages = sessionMessagesForSend.map((msg) => {
+      const newMessages = sessionMessagesForSend.map((msg) => {
         const { context: _context, id: _id, createdAt: _createdAt, ...withoutContext } = msg
         const rawMessage = toRaw(withoutContext)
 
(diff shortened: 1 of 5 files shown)


Vimime108:
Hello Team project AIRI !

I've been using AIRI locally for about 2 months now, and I just want to start by saying that this project is absolutely amazing. The way it integrates local TTS, STT, and LLMs is incredibly well done. Keep up the fantastic work!

I wanted to chime in on this PR. I've been experiencing the timestamp injection issue firsthand. Some models, particularly Dolphin-Llama3 (8B), tend to mirror the injected millisecond timestamps into their responses, which breaks the roleplay and chat experience.

This fix looks perfect, especially with the KV-cache optimization. I'm currently on the latest Windows installer build and can confirm the issue is still present there, so I'm really looking forward to seeing this merged into the next stable release.

Thank you for your hard work on this!

PR #1621, opened 2026-04-10
fix(server-runtime): preserve explicit empty route destinations

Gujiassh:
## Summary
- keep explicitly empty route destination arrays instead of falling back to the full destination list
- align the middleware behavior with the configured route record semantics
- add regression coverage for routes that intentionally disable all destinations

## Testing
- pnpm exec vitest run packages/server-runtime/src/middlewares/route.test.ts


nekomeowww:
Would you mind rebase this?

Gujiassh:
Rebased as requested. During rebase, the functional commit () was detected as already applied on current , and the remaining autofix lockfile commit was dropped.\n\nCurrent branch status vs : zero-diff ().\n\nSo this PR no longer carries a unique patch to merge.

Gujiassh:
Rebased as requested. During rebase, the functional fix commit was detected as already applied on current main, and the remaining autofix lockfile commit was dropped as redundant.

Current branch status vs main is zero-diff (0 ahead / 0 behind), so this PR no longer carries a unique patch to merge.

PR #1635, opened 2026-04-11
fix(server-runtime): preserve explicit empty data destinations

Gujiassh:
## Summary
- treat event.data.destinations the same way as event.route.destinations when the property is explicitly present
- preserve explicit empty destination arrays ([]) instead of collapsing them to undefined
- add a regression test for the explicit-empty-data override case

## Validation
- pnpm exec vitest run packages/server-runtime/src/middlewares/route.test.ts

This is a follow-up to the routing edge case discussed on #1621.

### DIFF
--- a/packages/server-runtime/src/middlewares/route.test.ts
+++ b/packages/server-runtime/src/middlewares/route.test.ts
@@ -99,22 +99,6 @@ describe('route middleware', () => {
 
     expect(collectDestinations(event)).toEqual(['label:env=prod'])
   })
-  it('respects explicit empty destinations as an override', () => {
-    const event = createSparkNotifyEvent({
-      data: {
-        id: 'evt-3',
-        eventId: 'spark-3',
-        kind: 'ping',
-        urgency: 'soon',
-        headline: 'hello',
-        destinations: ['module:character'],
-      },
-      route: { destinations: [] },
-    })
-
-    expect(collectDestinations(event)).toEqual([])
-  })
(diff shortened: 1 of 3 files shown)


nekomeowww:
@copilot resolve the merge conflicts in this pull request

Gujiassh:
Rebased this branch onto current `main` to resolve the merge conflict, and I also addressed the remaining runtime guard concern from review.\n\nUpdated head: `d3677d04`\n- rebased the branch onto latest `main`\n- kept explicit empty `data.destinations = []` as the override\n- guarded `collectDestinations` against truthy primitive `event.data` values before using the `in` operator\n- added a regression test for primitive payloads\n\nValidation:\n- `pnpm exec vitest run packages/server-runtime/src/middlewares/route.test.ts`

Gujiassh:
Follow-up update on the current branch head `c4be8a3e`.

I narrowed the destination override typing so `collectDestinations()` only returns array-shaped overrides, and removed the duplicate empty-route regression block from the route middleware tests.

Focused verification on the refreshed head:
- `pnpm exec vitest run packages/server-runtime/src/middlewares/route.test.ts`
- `pnpm run typecheck` (the workspace still has unrelated package-level dependency resolution noise, but `packages/server-runtime` now completes as `Done`, which clears the three `collectDestinations` type errors from the previous CI failure)

nekomeowww (approved):

PR #1638, opened 2026-04-11
fix(stage-ui): prevent Kokoro fp32-webgpu hang and fix STT mic device enumeration

ENTWOPY:
## Summary

- **Kokoro TTS no longer hangs on first visit** — `getDefaultKokoroModel` now always returns `q4f16` (WASM, ~320 MB) instead of `fp32-webgpu` (~700 MB). The WebGPU model caused the settings page to become unresponsive on first load while the worker silently attempted a massive background download. Users who want the full-precision WebGPU model can still select it manually.
- **Existing `fp32-webgpu` saves are migrated** — `kokoro-local.vue` now also migrates any already-saved `fp32-webgpu` config to `q4f16` on mount, so returning users are unblocked without having to manually reset their settings.
- **STT mic device dropdown no longer appears empty** — `hearing.vue` now calls `askPermission()` in `onMounted` so the browser permission dialog fires immediately when the user opens the Hearing page. Previously the dropdown stayed empty until the user manually triggered device enumeration, making STT look completely broken.
- **Auto-selected device is now persisted** — `audio-device.ts` adds a reverse watcher: when the `useAudioDevice` composable auto-selects a default mic after permission is granted, that selection is written back to `localStorage`. Without this, the dropdown and the audio stream could fall out of sync across reloads.

## Test plan

- [ ] Open Settings → Providers → Speech → Kokoro TTS (Local) on a fresh profile (no saved model) — page should load without hanging and `q4f16` should be selected automatically
- [ ] Open the same page with `fp32-webgpu` previously saved in `localStorage` — page should migrate to `q4f16` on mount instead of hanging
- [ ] Open Settings → Modules → Hearing — browser should prompt for microphone permission immediately; after granting, the audio input dropdown should be populated
- [ ] Reload the page after granting mic permission — the previously auto-selected device should still be selected in the dropdown

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/stage-pages/src/pages/settings/modules/hearing.vue
+++ b/packages/stage-pages/src/pages/settings/modules/hearing.vue
@@ -35,7 +35,7 @@ const providersStore = useProvidersStore()
 const { configuredTranscriptionProvidersMetadata } = storeToRefs(providersStore)
 
 const { trackProviderClick } = useAnalytics()
-const { stopStream, startStream } = useSettingsAudioDevice()
+const { stopStream, startStream, askPermission } = useSettingsAudioDevice()
 const { audioInputs, selectedAudioInput, stream } = storeToRefs(useSettingsAudioDevice())
 const { startRecord, stopRecord, onStopRecord } = useAudioRecorder(stream)
 const { startAnalyzer, stopAnalyzer, onAnalyzerUpdate, volumeLevel } = useAudioAnalyzer()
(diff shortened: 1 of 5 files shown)


nekomeowww (approved):

nekomeowww:
Rebase is needed, it conflicted.

PR #1687, opened 2026-04-19
fix(stage-tamagotchi): hide chat input scrollbar when content fits

### LINKED ISSUE #1567: Visible vertical scrollbar on the right side of the chat input bar
### Describe the bug

Description:
In the chat interface, a thin vertical line appears on the right side of the input bar. It is the scrollbar of the text area inside the input field.

Even when the input is empty or contains only a small amount of text, this scrollbar remains visible, which makes the UI look visually broken.

Expected behavior:
The chat input should not display a scrollbar in the default state. A scrollbar should only appear when the content exceeds the maximum height of the input field.

Actual behavior:
A thin vertical scrollbar is always visible on the right edge of the input bar, even when the input is empty or contains only one line of text.

Possible fix:
Hide the default scrollbar style of the textarea
Enable internal scrolling only after the input reaches its maximum height

Screenshot:
Image

### System Info

```Shell
Environment:

AIRI Desktop
Version: 0.9.0-beta.2
OS: macOS
```

### Validations

### Contributions

octo-patch:
Fixes #1567

## Problem

The chat input textarea in the desktop app (stage-tamagotchi) uses `overflow-y-scroll`, which forces the browser to always render a vertical scrollbar, even when the input is empty or contains only a small amount of text. This makes the UI look visually broken.

## Solution

Changed `overflow-y-scroll` to `overflow-y-auto` in `apps/stage-tamagotchi/src/renderer/components/InteractiveArea.vue`.

With `overflow-y-auto`:
- No scrollbar is shown when content fits within the textarea's height (auto-resize keeps height == scrollHeight)
- Scrollbar appears only when content exceeds the `max-h` constraint (`10lh`)

This matches the expected behavior described in the issue.

## Testing

- Open the desktop app (stage-tamagotchi)
- Verify the chat input shows no scrollbar when empty or with a short message
- Type a long multi-line message that exceeds the max height and verify a scrollbar appears

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/components/InteractiveArea.vue
+++ b/apps/stage-tamagotchi/src/renderer/components/InteractiveArea.vue
@@ -252,12 +252,12 @@ async function handleDeleteMessage(index: number) {
       v-model="messageInput"
       :submit-on-enter="false"
       :placeholder="t('stage.message')"
-      class="ph-no-capture"
+      class="ph-no-capture [scrollbar-gutter:stable]"
       text="primary-600 dark:primary-100  placeholder:primary-500 dark:placeholder:primary-200"
       border="solid 2 primary-200/20 dark:primary-400/20"
       bg="primary-100/50 dark:primary-900/70"
       max-h="[10lh]" min-h="[1lh]"
-      [scrollbar-gutter:stable] w-full shrink-0 resize-none overflow-y-auto rounded-xl p-2 font-medium outline-none
+      w-full shrink-0 resize-none overflow-y-auto rounded-xl p-2 font-medium outline-none
       transition="all duration-250 ease-in-out placeholder:all placeholder:duration-250 placeholder:ease-in-out"
       @compositionstart="isComposing = true"
       @compositionend="isComposing = false"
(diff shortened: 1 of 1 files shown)


octo-patch:
Thanks for the catch! Good point about the layout shift — the scrollbar appearing as content grows would shove the input width around. Added `[scrollbar-gutter:stable]` (UnoCSS arbitrary value) so the gutter is reserved up-front and the input width stays constant regardless of scrollbar visibility. (ddcf2407)

octo-patch:
Quick follow-up: the previous commit placed `[scrollbar-gutter:stable]` as a bare attributify token, but bracketed UnoCSS shorthand isn't a valid HTML attribute name in Vue templates and `vue-tsc` rejected it (`TS1005: ']' expected`). Pushed [a8f538f](https://github.com/octo-patch/airi/commit/a8f538ff) which moves it inside the existing `class` attribute alongside `ph-no-capture`. The runtime behavior (reserving the scrollbar gutter so the input width does not jump) is unchanged. CI should be green now.

nekomeowww (approved):

PR #1702, opened 2026-04-21
feat(pipelines-audio): Universal Speech Transformer port with CJK & keep-text support

0xSelenicDove:
## Description

As discussed with @dasilva333 on Discord, this PR ports the Universal Speech Transformer (UST) logic from the `stage-ui` frontend down to the `pipelines-audio` core (`tts-chunker.ts`). This architectural move ensures all backend TTS providers natively benefit from narrative stripping before the text even hits the API.

**Key Enhancements:**
1. **True Universal CJK Support**: Upgraded the regex to properly handle full-width CJK brackets like `（ ）` and `【 】`.
2. **Wishlist Fulfilled (`keepNarrativeText`)**: Added a new `keepNarrativeText` option in `TtsInputChunkOptions`. Using regex capture groups, the pipeline now supports two modes:
   - `keepNarrativeText: false` (Default): `[laughs]` -> `""` (strips everything)
   - `keepNarrativeText: true`: `[laughs]` -> `"laughs"` (drops punctuation, keeps the inner word)

## Linked Issues

Related to the ongoing Fish Audio integration and UST improvements discussed in Discord.

## Additional Context

- The `pipelines-audio` workspace builds and passes typechecks perfectly locally.
- **Note for Reviewers:** The CI check might fail due to the existing `navigator.gpu` typecheck issue in `stage-ui/providers.ts` that is currently present on `main`. This PR only touches the audio pipeline and does not introduce any UI type errors.

PR #1708, opened 2026-04-22
fix(pipelines-audio): resolve narrative stripping edge cases and add comprehensive tests

0xSelenicDove:
## Description

This PR provides a comprehensive overhaul of the streaming narrative stripping logic in `pipelines-audio`. It addresses the complex edge cases and regressions identified by Codex and Gemini-Code-Assist in #1702, ensuring that math operators, code literals, and Markdown syntax are preserved while narrative spans are accurately stripped.

### 1. Robust Heuristic for Angle Brackets (`< >`)
* **Bidirectional Detection:** Introduced `isProbablyAngleTag` as a dedicated helper. It uses both lookahead (checking for digits/spaces) and lookbehind (checking for preceding characters like `List<T>`) to distinguish between code generics/math and narrative tags.
* **Hyphenated & Alphanumeric Support:** Refined the heuristic to correctly identify and protect HTML-like tags (e.g., `<h2>`) and hyphenated component tags (e.g., `<my-tag>`) that might be split across streaming chunks.

### 2. Range-Based Parsing State Machine
* **Non-Destructive Scanning:** Shifted from a single-pass "strip-as-you-go" approach to an index-based range deletion method in `processNarrative`.
* **Leakage Prevention:** By identifying and removing only validated and closed ranges, we ensure that unmatched openers (e.g., `Version (beta` or `a * b`) no longer cause subsequent non-narrative text to be swallowed.

### 3. Dynamic Streaming Fallback
* **Unified Buffer Safeguard:** Implemented a dynamic `fallbackLimit`. While active narrative markers are being tracked, the buffer limit is increased to 800 characters to prevent premature flushing of descriptive spans, while maintaining a sensitive 200-character limit for normal text to minimize latency.

## Edge Cases Solved
* **Markdown vs. Narrative:** Properly distinguishes between Markdown bullets (`* item`) and narrative emphasis (`*sigh*`) using whitespace-sensitive state toggling.
* **CJK Bracket Support:** Restored full stripping support for full-width/Chinese brackets (`（ ）`, `【 】`) that were missing in previous iterations.
* **Literals Preservation:** Fixed a regression in `keepNarrativeText` mode where global regex previously corrupted literal bracket characters in math/code expressions.

## Tests Added
* **Bidirectional Math Detection:** Validates `x<y` (preserved) vs. `List<String>` (preserved) vs. `<sigh>` (stripped).
* **Split-Chunk Tags:** Ensures tags split across streaming boundaries (e.g., `<s ... mile>`) are correctly tracked as unclosed.
* **Nested Bracket Handling:** Verifies complex nesting like `[laughs (quietly)]`.
* **Markdown Ambiguity:** Confirms that trailing punctuation inside brackets (e.g., `[note.]`) and unclosed asterisk算式 are handled correctly.

### DIFF
--- a/packages/pipelines-audio/src/processors/tts-chunker.ts
+++ b/packages/pipelines-audio/src/processors/tts-chunker.ts
@@ -310,6 +310,11 @@ export function createTtsSegmentStream(
                   if (/^[a-z][^a-z\s>]/i.test(remainder)) {
                     continue
                   }
+
+                  const leftStr = pendingText.slice(0, i)
+                  if (/^[a-z]$/i.test(remainder) && /(^|[^a-z])[a-z]$/i.test(leftStr)) {
+                    continue
+                  }
                 }
                 stack.push(char)
               }
--- a/packages/pipelines-audio/src/processors/tts-chunker.ts
+++ b/packages/pipelines-audio/src/processors/tts-chunker.ts
@@ -303,6 +303,10 @@ export function createTtsSegmentStream(
                 if (char === '<') {
                   const remainder = pendingText.slice(i + 1)
 
+                  if (remainder.length > 0 && /[0-9\s=]/.test(remainder[0])) {
+                    continue
+                  }
+
                   if (/^[a-z]([^a-z>]|$)/i.test(remainder)) {
                     continue
                   }
--- a/packages/pipelines-audio/src/processors/tts-chunker.ts
+++ b/packages/pipelines-audio/src/processors/tts-chunker.ts
@@ -333,7 +333,7 @@ export function createTtsSegmentStream(
               && starMatch !== null && !starMatch[1].startsWith(' ')
             const hasUnclosed = bracketsUnclosed || starsUnclosed
             const isStrippingActive = options?.stripNarrative && hasUnclosed
-            const fallbackLimit = isStrippingActive ? 800 : 200
+            const fallbackLimit = (isStrippingActive && bracketsUnclosed) ? 800 : 200
 
             if (!hasUnclosed || pendingText.length > fallbackLimit) {
               const textToEmit = processNarrative(pendingText, options)
--- a/packages/pipelines-audio/src/processors/tts-chunker.ts
+++ b/packages/pipelines-audio/src/processors/tts-chunker.ts
@@ -307,7 +307,7 @@ export function createTtsSegmentStream(
                     continue
                   }
 
-                  if (/^[a-z]([^a-z>]|$)/i.test(remainder)) {
+                  if (/^[a-z][^a-z>]|\s/i.test(remainder)) {
                     continue
                   }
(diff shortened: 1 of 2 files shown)


0xSelenicDove:
@gemini-code-assist I've resolved all the Codex P1/P2 issues and updated the unit tests. Can you do a final sweep to see if I missed any documentation, potential type mismatches, or if the code could be cleaner?

0xSelenicDove:
@gemini-code-assist help me check the new fixed code

packages/pipelines-audio/src/processors/tts-chunker.ts
```diff
@@ -241,21 +241,147 @@ export async function chunkEmitter(
...
+
+  // Lookahead: if followed by num, space or equals, not a label
+  if (nextChar && /[0-9\s=]/.test(nextChar))
+    return false
+
+  if (prevChar && (isUnicodeLetter(prevChar) || /\d/.test(prevChar))) {
+    // fix: check whether remainder is piefix with any keywords, or contains the whole keyword
+    const isLikelyNarrative = NARRATIVE_KEYWORDS.some(kw =>
+      (remainder.length > 1 && kw.startsWith(remainder)) || remainder.startsWith(kw),
+    )
```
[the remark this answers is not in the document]
0xSelenicDove: Acknowledged. This is a deliberate trade-off to prioritize code/math literal preservation (e.g., List<T>) and minimize streaming latency. Given the infrequency of non-spaced custom tags, we're treating this as an acceptable edge case.

0xSelenicDove:
@nekomeowww I think I'm done here, pls check! If you have suggestions just leave a comment, thx!

packages/pipelines-audio/src/processors/tts-chunker.ts
```diff
@@ -241,21 +241,147 @@ export async function chunkEmitter(
...
+  'sigh',
+  'action',
+  'note',
+  'breath',
+  'giggle',
+  'whisper',
+  'cry',
+  'smile',
+  'thought',
+]
```
nekomeowww, line 267: Should this be configurable?

nekomeowww (approved):

packages/pipelines-audio/src/processors/tts-chunker.ts
0xSelenicDove: I thought about it, but decided to keep it as a hardcoded 'sane default' for now.

The primary goal of this PR is to stabilize the stripping logic and handle streaming edge cases. Making it configurable would increase the API surface area and might introduce new logical conflicts (e.g., users adding very short keywords that break code literal protection). If we see a strong demand for custom tags in the future, we can easily expose this as an option then.

nekomeowww (approved):

PR #1728, opened 2026-04-25
fix(stage-ui): guard Kokoro generation until ready

fuyua9:
## Summary
- reject Kokoro `generate()` calls unless the adapter has a worker and the model state is `ready`
- keep the not-loaded error as a usage/state error instead of routing it through worker restart handling
- add a focused regression test that checks generation before load does not move the adapter out of `idle`

## Why
`generate()` previously only checked whether a worker object existed. After lifecycle errors or restarts, a worker can exist while the Kokoro model is not loaded/ready, which lets `run-inference` be posted to a fresh worker without a loaded model. Guarding on `state === 'ready'` keeps the public generation path aligned with the model lifecycle.

## Validation
- `git diff --check -- packages/stage-ui/src/libs/inference/adapters/kokoro.ts packages/stage-ui/src/libs/inference/adapters/kokoro.test.ts`
- static assertion that the ready-state guard and regression test are present

### DIFF
--- a/packages/stage-ui/src/libs/inference/adapters/kokoro.test.ts
+++ b/packages/stage-ui/src/libs/inference/adapters/kokoro.test.ts
@@ -2,10 +2,30 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 // Mock Worker globally since it's not available in Node
 class MockWorker {
-  addEventListener = vi.fn()
-  removeEventListener = vi.fn()
+  static instances: MockWorker[] = []
+
+  listeners = new Map<string, Set<(event: any) => void>>()
+  addEventListener = vi.fn((type: string, listener: (event: any) => void) => {
+    if (!this.listeners.has(type))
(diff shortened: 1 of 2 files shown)


fuyua9:
Thanks for the review. I pushed a follow-up commit f6da2d63 to make the two guarantees explicit in tests:

- loadModel(..., { signal }) now has focused coverage that the signal is forwarded into the load queue and that aborting an in-flight load sends a worker cancel message before rejecting with AbortError.
- Worker DEVICE_LOST errors now have focused coverage that the adapter classifies the original error payload, increments deviceLossCount, and records the loss against the active Kokoro model before restart.

Validation run locally:

pnpm --filter @proj-airi/stage-ui test:run src/libs/inference/adapters/kokoro.test.ts
Test Files: 1 passed (1)
Tests: 10 passed (10)

I also ran git diff --check and a conflict-marker scan before pushing.

nekomeowww (approved):

PR #1757, opened 2026-04-29
feat(stage-tamagotchi): add local account avatar upload

fordelkon:
## Description

  This PR adds local account avatar upload support for the stage account settings page.

  Users can now choose an image from their local device, preview it immediately as their account avatar, and keep that avatar
  across app restarts. The local avatar is stored per user ID, so switching accounts does not leak avatars between users.

  The avatar display logic is shared through a stage-ui composable, so both the account settings page and the stage-tamagotchi
  controls island auth button show the same locally customized avatar. The implementation avoids mutating the server-backed auth
  user object, preventing fetchSession() from resetting the local avatar after restarting pnpm dev:tamagotchi or reopening the
  app.

  The upload flow includes MIME type validation, a 1 MiB size limit, localized success/error messages, and translated strings
  across the affected locales.

  ## Linked Issues

  N/A

  ## Additional Context

  Review focus:

  - Local avatar persistence is intentionally device-local and not synced to the server.
  - The local avatar overrides user.image only at display time.
  - authStore.user.image is not mutated, so session refreshes can safely replace server auth state.
  - controls-island-auth-button.vue now reads the same local avatar source as the settings page.


nekomeowww:
Rebase is needed.

PR #1759, opened 2026-04-30
fix: resolve provider validation and CORS issues in dev mode

haha396475360-byte:
- Only validate explicitly added providers to prevent unnecessary connection attempts
- Start periodic validation only for added providers
- Watch for newly added providers and start their validation
- Disable webSecurity in development mode to fix CORS errors with external APIs

Fixes connection refused errors for Ollama/LM Studio when not running locally. Fixes CORS errors when using OpenAI-compatible APIs in development mode.


nekomeowww:
Not quite correct approach for implementing this.

PR #1765, opened 2026-05-02
fix(chat): strip echoed datetime prefixes from streamed model output

felixtremblay:
## Summary

Weak local models echo back the `[YYYY-MM-DD HH:MM] ` prefix that we inject onto every user/assistant turn for KV-cache stability (`packages/stage-ui/src/stores/chat/datetime-prefix.ts`). The echoed prefix was leaking into both the chat transcript and the TTS audio. Two flavors observed:

1. Leading echo at the start of the reply: `[2026-05-01 20:08] ooh nice...`
2. Mid-stream echo after the model emits its own newline: `Cool, cool. What kind of edge cases?\n[2026-05-01 20:25] Like, is it the timing?...`

This patch adds a streaming-aware stripper at the single chokepoint feeding both surfaces (`onLiteral` in `chat.ts`), plus a one-shot helper used in `onEnd` so the persisted text does not re-send `[ts] [ts] ...` on subsequent turns and reinforce the pattern.

- New `stripLeadingTimestampPrefix(text)` and `createTimestampPrefixStripper()` in `packages/stage-ui/src/stores/chat/datetime-prefix.ts`. Stripping is line-anchored (start-of-stream or after `\n`), not just leading.
- The streaming version tracks the last character the model emitted across `consume()` calls so boundary detection stays correct after the stripper rewrites the output.
- Wired into `chat.ts` so neither chat UI (`buildingMessage.content`, `slices`) nor TTS (`hooks.emitTokenLiteralHooks`) ever sees the prefix.

## Test plan

- [x] 14 unit tests in `datetime-prefix.test.ts` covering: single-chunk strip, prefix split across many small chunks, trailing-space arriving in next chunk, no-prefix passthrough, `[note]` false positive, mid-line timestamp passthrough, prefix without trailing space, partial prefix at stream end, almost-prefix with non-digit, mid-stream `\n[ts]` echo, newline-then-bracket split across chunks, multiple newline-anchored prefixes, `\n[note]` not stripped.
- [x] `pnpm exec vitest run packages/stage-ui/src/stores/chat` (34/34 pass).
- [x] `pnpm -F @proj-airi/stage-ui typecheck` clean.
- [x] `pnpm lint` shows no new findings on the changed files.
- [ ] Manual: send a few turns with a weak local model and confirm the timestamp no longer appears in the rendered transcript or in TTS audio.

### DIFF
--- a/apps/stage-tamagotchi/package.json
+++ b/apps/stage-tamagotchi/package.json
@@ -18,6 +18,7 @@
     "app:build": "pnpm run build",
     "start": "electron-vite preview",
     "dev": "electron-vite dev",
+    "dev:xwayland": "electron-vite dev -- --ozone-platform=x11",
     "build": "electron-vite build",
     "postinstall": "electron-builder install-app-deps",
     "build:unpack": "pnpm run build && electron-builder --dir",
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "dev:pocket:android": "pnpm -rF @proj-airi/stage-pocket run dev:android",
     "dev:server": "pnpm -rF @proj-airi/server-runtime run dev",
     "dev:tamagotchi": "pnpm -rF @proj-airi/stage-tamagotchi run dev",
+    "dev:tamagotchi:xwayland": "pnpm -rF @proj-airi/stage-tamagotchi run dev:xwayland",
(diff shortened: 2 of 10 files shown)


felixtremblay:
meant to be in my fork. my bad

PR #1766, opened 2026-05-02
fix(server): count TTS billing characters by code point

fuyua9:
## Summary

- Count TTS billing units by Unicode code point instead of UTF-16 code units.
- Use the same computed character count for both affordability checks and debt-ledger accumulation.
- Add a regression test covering astral Unicode symbols so emoji are not billed as two characters.

## Why

The TTS billing configuration and docs describe `FLUX_PER_1K_CHARS_TTS` as a per-character rate, but JavaScript `string.length` counts UTF-16 code units. That means astral symbols such as emoji consume two units even though they are one Unicode code point.

## Tests

- `pnpm --filter @proj-airi/server exec vitest run src/routes/openai/v1/route.test.ts`
- `pnpm -F @proj-airi/server typecheck`


luoling8192:
It's unnecessary, but it will cause a performance issue

PR #1783, opened 2026-05-06
fix(stage-ui-live2d): prevent motions from overriding lip sync mouth values

felixtremblay:
## Context

Speech audio drives `mouthOpenSize` via the wlipsync analyser, but the previous wiring used a Vue `watch` inside `packages/stage-ui-live2d/src/components/scenes/live2d/Model.vue` to push the value into `ParamMouthOpenY` from outside the Pixi render loop. The motion update hook ran in the Pixi tick and overwrote the watch most frames, so the mouth barely opened or stayed completely open during speech.

## Changes

Lip sync now runs as a motion-update plugin in the `final` stage instead of a Vue `watch` outside the Pixi render loop, with a 200 ms smoothstep release fade so the mouth eases back to motion's value when speech ends. Also stops shadowing `useSpeakingStore.nowSpeaking` with a local ref in `Stage.vue` (side effect: `nowSpeakingAvatarBorderOpacity` now toggles during TTS). Small `normalizeTimeDeltaMs` helper extracted along the way.

## Tested

- `pnpm -F @proj-airi/stage-ui typecheck` clean
- `pnpm -F @proj-airi/stage-ui-live2d typecheck` clean
- `pnpm exec eslint` on the four changed files clean
- Local TTS playback, ElevenLabs tts playback.

## Note

This adds a smoothing at the end so that the mouth can go back to its current idle animation coordinate without skipping to it right away.\
I did not implement a smoothing transition at the beginning of the animation because I was worried that it would add an un-natural delay, but it does mean that the transition from "idle animation mouth" to "speaking mouth" can show a slight jitter. \
I'm open to opinions on the subject.

### DIFF
--- a/packages/stage-ui-live2d/src/composables/live2d/motion-manager.ts
+++ b/packages/stage-ui-live2d/src/composables/live2d/motion-manager.ts
@@ -129,6 +129,14 @@ export function useLive2DMotionManagerUpdate(options: UseLive2DMotionManagerUpda
 
 // -- Plugins ---------------------------------------------------------------
 
+// pixi-live2d-display passes timeDelta in ms during normal frames but can pass
+// seconds at frame boundaries; threshold of 5 separates the two ranges safely
+// (5 ms == ~200 fps lower bound, 5 s == 5000 ms upper bound).
+function normalizeTimeDeltaMs(timeDelta: number | undefined): number {
+  const raw = Math.max(timeDelta ?? 0, 0)
+  return raw < 5 ? raw * 1000 : raw
+}
+
 export function useMotionUpdatePluginBeatSync(beatSync: BeatSyncController): MotionManagerPlugin {
   return (ctx) => {
     beatSync.updateTargets(ctx.now)
@@ -330,9 +338,7 @@ export function useMotionUpdatePluginAutoEyeBlink(
 
       // Force ON or eyeBlink null: timer blink + markHandled.
       if (ctx.live2dForceAutoBlinkEnabled.value || !ctx.internalModel.eyeBlink) {
-        const rawDelta = Math.max(ctx.timeDelta ?? 0, 0)
-        const dt = rawDelta < 5 ? rawDelta * 1000 : rawDelta
-        const safeDt = dt || 16
+        const safeDt = normalizeTimeDeltaMs(ctx.timeDelta) || 16
         const { eyeLOpen, eyeROpen } = updateForcedBlink(safeDt, baseLeft, baseRight)
         ctx.model.setParameterValueById('ParamEyeLOpen', eyeLOpen)
         ctx.model.setParameterValueById('ParamEyeROpen', eyeROpen)
@@ -399,9 +405,7 @@ export function useMotionUpdatePluginAutoEyeBlink(
 
     // Advance blink timer.
     const wasActive = blinkState.phase !== 'idle'
-    const rawDelta = Math.max(ctx.timeDelta ?? 0, 0)
-    const dt = rawDelta < 5 ? rawDelta * 1000 : rawDelta
-    const safeDt = dt || 16
+    const safeDt = normalizeTimeDeltaMs(ctx.timeDelta) || 16
     const { eyeLOpen: blinkFactorL, eyeROpen: blinkFactorR } = updateForcedBlink(safeDt, 1.0, 1.0)
 
     // Blink cycle complete: restore exact pre-blink values.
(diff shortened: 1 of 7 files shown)


felixtremblay:
> ## Code Review
> This pull request enhances Live2D lip-syncing by adding a nowSpeaking prop and a useMotionUpdatePluginLipSync plugin, which implements a smooth cross-fade transition between speech-driven and motion-driven mouth states. It also introduces a utility for normalizing time deltas. The review feedback correctly identifies that the 5ms threshold used for time normalization is too high for high-refresh-rate monitors (e.g., 240Hz) and recommends lowering it to 1ms to prevent animation glitches.
> 
> I am having trouble creating individual review comments. Click here to see my feedback.
> ### packages/stage-ui-live2d/src/composables/live2d/motion-manager.ts (132-139)
> medium
> 
> The threshold of `5` ms for distinguishing seconds from milliseconds is problematic for high-refresh-rate displays. For example, a 240Hz monitor has a frame time of ~4.16ms, and a 360Hz monitor has ~2.77ms. On such screens, this logic will incorrectly multiply the millisecond delta by 1000, causing animations (like eye blinks and the new lip sync release) to behave erratically. Lowering the threshold to `1` is safer, as it supports up to 1000 FPS while still correctly identifying second-based deltas for anything down to 1 FPS.
> 
> ```
> // pixi-live2d-display passes timeDelta in seconds (e.g. 0.016 at 60 fps) on most
> // paths, but some callers in this codebase pass it already in ms. The < 1 threshold
> // distinguishes them: a 1 ms frame would imply 1000 fps, so any value below 1
> // must be seconds and gets multiplied to ms.
> function normalizeTimeDeltaMs(timeDelta: number | undefined): number {
>   const raw = Math.max(timeDelta ?? 0, 0)
>   return raw < 1 ? raw * 1000 : raw
> }
> ```

This value is based on existing code. I just extracted it into a helper.\
I'm not against changing it but it seems a bit out of the scope of this PR.

packages/stage-ui-live2d/src/composables/live2d/motion-manager.ts
```diff
@@ -129,6 +129,15 @@ export function useLive2DMotionManagerUpdate(options: UseLive2DMotionManagerUpda
 
 // -- Plugins ---------------------------------------------------------------
 
+// pixi-live2d-display passes timeDelta in seconds (e.g. 0.016 at 60 fps) on most
+// paths, but some callers in this codebase pass it already in ms. The < 5 threshold
+// distinguishes them: a 5 ms frame would imply ~200 fps which never happens, so
+// any value below 5 must be seconds and gets multiplied to ms.
+function normalizeTimeDeltaMs(timeDelta: number | undefined): number {
+  const raw = Math.max(timeDelta ?? 0, 0)
+  return raw < 5 ? raw * 1000 : raw
```
felixtremblay: The more I look at this logic, the less confident I become about it 😅.\
@nekomeowww : The git blame of the original logic points to you. Do you remember why this had to be done? Currently, I don't see any use cases in which it would be in ms 🤔

nekomeowww (approved):

packages/stage-ui-live2d/src/composables/live2d/motion-manager.ts
nekomeowww: I can't quite remember the reason. I used to log out the delta values and thought it was milliseconds, this is the far most i could recall.

nekomeowww: Would you mind helping me to check whether the timeDelta values are in millisecond unit?

felixtremblay: Did some digging.

Tracing it through `pixi-live2d-display@0.4.0`,`motionManager.update` only ever gets called from `Cubism4InternalModel.update`, and that method divides both `dt` and `now` by 1000 before calling us:

https://github.com/guansss/pixi-live2d-display/blob/v0.4.0/src/cubism4/Cubism4InternalModel.ts#L184-L195

```ts
public update(dt: DOMHighResTimeStamp, now: DOMHighResTimeStamp): void {
    super.update(dt, now);

    dt /= 1000;
    now /= 1000;
    ...
    const motionUpdated = this.motionManager.update(this.coreModel, now);
```

So by the time `now` reaches our hook, it is already in seconds, which means our locally computed `timeDelta = now - lastUpdateTime.value` is also always in seconds.

I also confirmed this with a temporary log. It's always in seconds.

**Note:**\
This is only true for `Cubism4`. `Cubism2` does not do the conversion to seconds.\
I don't see any kind of support for `Cubism2` in the Airi project so it's probably less confusing to remove the helper and assume that this will always be seconds (unless you have objections).

felixtremblay: Removed the helper and rebased on `main` :)

nekomeowww (approved):

PR #1801, opened 2026-05-09
fix(computer-use-mcp): terminate one-shot PTY commands

duyua9:
## Summary

- ensure one-shot PTY workflow commands end with a CR/LF before sending them through `pty_send_input`
- keep caller-provided command terminators unchanged
- add a workflow-engine regression test for the PTY `run_command` path

## Why

`executePtyCommand` drives `pty_send_input` directly for one-shot PTY steps. Without an explicit command terminator, the command can stay buffered at the shell prompt instead of being submitted.

## Tests

- `git diff --check`
- `bash /Users/ming/.openclaw/skills/code-change-delivery-contract/scripts/scan_delivery_blockers.sh /Users/ming/code/moeru-ai/airi-pty-terminator-duyua9/services/computer-use-mcp/src/workflows`

Not run locally: `pnpm -F @proj-airi/computer-use-mcp exec vitest run src/workflows/engine.test.ts` because this clean worktree had no installed `vitest`, and a scoped `pnpm install --frozen-lockfile --ignore-scripts --filter @proj-airi/computer-use-mcp...` was interrupted by repeated registry `ECONNRESET` errors through the configured proxy.

### DIFF
--- a/services/computer-use-mcp/src/workflows/engine.test.ts
+++ b/services/computer-use-mcp/src/workflows/engine.test.ts
@@ -819,6 +819,47 @@ describe('workflow engine', () => {
     expect(displayInfo?.displays).toHaveLength(2)
   })
 
+  it('terminates one-shot PTY commands before sending input', async () => {
+    const wf: WorkflowDefinition = {
+      id: 'test_pty_command_terminator',
+      name: 'PTY Command Terminator Test',
+      description: 'One-shot PTY commands should be submitted to the shell.',
+      maxRetries: 3,
+      steps: [
+        {
+          label: 'Run PTY command',
+          kind: 'run_command',
+          description: 'Run via PTY',
+          params: { command: 'echo hello' },
+          terminal: { mode: 'pty', interaction: 'one_shot' },
+        },
+      ],
+    }
+
+    const executeAction: ExecuteAction = vi.fn().mockResolvedValue(makeSuccessResult())
+    const executePrepTool = vi.fn().mockImplementation(async (toolName: string) => {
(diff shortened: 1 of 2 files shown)


duyua9:
Thanks for the review. I moved `ensurePtyCommandTerminator` above the existing `executePtyCommand` docstring so the comment is again attached to `executePtyCommand`.

Validation:
- `git diff --check`
- scoped test command attempted, but this checkout currently has no `node_modules` / `vitest` available locally

3361559784:
Thanks, that resolves the docstring placement issue for me.

I understand the local dependency issue around running the scoped vitest command. Before merge, I’d still prefer we have actual test evidence somewhere, either from CI or from a local rerun once dependencies are available:

pnpm -F @proj-airi/computer-use-mcp exec vitest run src/workflows/engine.test.ts

One more small coverage gap: since the PR summary says caller-provided command terminators are preserved, could you add a regression case showing that commands already ending in \n or \r are not double-terminated?

The implementation direction looks good to me; this is mostly about closing the validation gap.

3361559784:
> 感谢您的审阅。我已将注释`ensurePtyCommandTerminator`移至现有`executePtyCommand`文档字符串上方，因此注释再次附加到文档字符串中`executePtyCommand`。
> 
> 验证：
> 
> * `git diff --check`
> * 尝试执行范围测试命令，但此检出当前本地没有可用的`node_modules`/`vitest`

Please ping me if anything is unclear. Once the validation gap is closed, ping me again and I’ll do the final review / merge check.

PR #1780, opened 2026-05-06
test(computer-use-mcp): add desktop v3 smoke coverage

3361559784:
Broadens desktop v3 coverage across the overlay and Chrome grounding stack.

What it does:
- adds live-window smoke coverage and helper scripts under `apps/stage-tamagotchi`
- expands `services/computer-use-mcp` with `smoke-chrome-grounding`, Chrome session handling, desktop grounding updates, and supporting tests
- updates `support-matrix` and `docs/desktop-lane-status.md` to match the new coverage
- keeps the packaging and lockfile wiring in sync where needed

Scope:
- tamagotchi desktop overlay runtime and smoke scripts
- computer-use-mcp browser/desktop grounding, session lifecycle, and smoke runner
- docs and support-matrix status

Verification:
- `pnpm -C services/computer-use-mcp exec vitest run src/bin/smoke-chrome-grounding.test.ts src/support-matrix.test.ts`
- `pnpm -C services/computer-use-mcp exec tsc --noEmit --pretty false`
- `git diff --check`

PR #1804, opened 2026-05-09
feat(desktop): background scheduler for observe/click

3361559784:
desktop_observe no longer forces a foreground switch just to capture Chrome semantics. desktop_click_target only switches foreground when the browser-dom route is unavailable and the action falls back to OS input.

Validation:
- pnpm -C services/computer-use-mcp exec vitest run src/desktop-grounding.test.ts src/server/register-desktop-grounding-tools.test.ts src/server/register-desktop-grounding.test.ts src/server/action-executor.test.ts src/server/desktop-scheduler.test.ts
- pnpm -C services/computer-use-mcp exec tsc --noEmit --pretty false
- git diff --check

This is intentionally separate from PR #1780.


3361559784:
Superseded by #1805, which is the recut from latest origin/main with only the background scheduler slice.

PR #1808, opened 2026-05-09
fix(stage): refresh attachment preview URLs after failed send

juyua9:
## Summary
- recreate attachment preview object URLs after a failed send before restoring the attachment list
- revoke the optimistic-send object URLs on both success and failure
- revoke any remaining attachment preview URLs when the interactive area unmounts
- add focused tests for recreating and revoking attachment object URLs

## Tests
- `node ../../node_modules/vitest/dist/cli.js run src/renderer/utils/chat-attachments.test.ts --config vitest.config.ts` (from `apps/stage-tamagotchi`) — 2 passed

Note: I also attempted a full workspace `pnpm install --frozen-lockfile --ignore-scripts` to run the normal package commands, but the registry repeatedly returned `ECONNRESET` before the install completed.


juyua9:
Closing this after rerunning the submission value gate on the final PR shape. The patch is a small resource-lifecycle cleanup without an upstream issue or maintainer request, and the stricter gate classifies it as needs-human-override rather than an automatic submission. Sorry for the noise.

PR #1815, opened 2026-05-11
fix(stage-tamagotchi): prevent duplicate desktop instances

Neko-233:
## Summary

- add an Electron single-instance guard for the Tamagotchi desktop main process
- quit secondary launches before local runtime services start
- focus the existing AIRI window when Windows forwards a second launch to the primary instance

## Why

Windows can launch multiple AIRI desktop processes. Each process starts the server channel on the fixed loopback address `127.0.0.1:6121`, so the second process hits `EADDRINUSE: address already in use` after the first process has already bound the port.

Using Electron `requestSingleInstanceLock()` keeps one process responsible for the local runtime resources while preserving the expected user behavior: launching AIRI again brings the existing window forward.

## Validation

- `pnpm exec vitest run apps/stage-tamagotchi/src/main/app/single-instance.test.ts apps/stage-tamagotchi/src/main/services/airi/channel-server/config.test.ts packages/server-runtime/src/server.test.ts`
- `pnpm exec eslint --cache apps/stage-tamagotchi/src/main/index.ts apps/stage-tamagotchi/src/main/app/single-instance.ts apps/stage-tamagotchi/src/main/app/single-instance.test.ts`
- `pnpm -F @proj-airi/stage-tamagotchi typecheck`
- `git diff --check`

### DIFF
--- a/apps/stage-tamagotchi/src/main/app/single-instance.test.ts
+++ b/apps/stage-tamagotchi/src/main/app/single-instance.test.ts
@@ -1,0 +1,78 @@
+import type { App, BrowserWindow } from 'electron'
+
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const windowMock = vi.hoisted(() => ({
+  toggleWindowShow: vi.fn(),
+}))
+
+vi.mock('../windows/shared/window', () => ({
+  toggleWindowShow: windowMock.toggleWindowShow,
+}))
+
+function createMockApp(hasSingleInstanceLock: boolean): MockApp {
+  return {
+    on: vi.fn(),
+    quit: vi.fn(),
+    requestSingleInstanceLock: vi.fn(() => hasSingleInstanceLock),
+  } as unknown as MockApp
+}
+
+function createMockWindow() {
+  return {} as BrowserWindow
+}
+
+describe('installSingleInstanceGuard', async () => {
+  const { installSingleInstanceGuard } = await import('./single-instance')
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  /**
+   * @example
+   * const installed = installSingleInstanceGuard({ app, getWindow })
+   * expect(installed).toBe(false)
+   */
+  it('quits the secondary process when another AIRI instance already owns the lock', () => {
+    const app = createMockApp(false)
+
+    const installed = installSingleInstanceGuard({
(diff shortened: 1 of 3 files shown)


apps/stage-tamagotchi/src/main/app/single-instance.ts
```diff
@@ -0,0 +1,59 @@
...
+  const window = getAllWindows().find(window => !window.isDestroyed())
+  if (!window) {
+    return
+  }
+
+  if (window.isMinimized()) {
+    window.restore()
+  }
+  window.show()
+  window.focus()
```
nekomeowww: We should have existing utils to show and focus window?

apps/stage-tamagotchi/src/main/app/single-instance.ts
```diff
@@ -0,0 +1,59 @@
+import type { App, BrowserWindow } from 'electron'
+
+interface SingleInstanceGuardOptions {
+  app: Pick<App, 'on' | 'quit' | 'requestSingleInstanceLock'>
```
nekomeowww: Simply `App`

apps/stage-tamagotchi/src/main/app/single-instance.test.ts
```diff
@@ -0,0 +1,71 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+function createMockApp(hasSingleInstanceLock: boolean) {
+  return {
+    on: vi.fn(),
+    quit: vi.fn(),
+    requestSingleInstanceLock: vi.fn(() => hasSingleInstanceLock),
+  }
+}
```
nekomeowww: Mock could `as App`.

apps/stage-tamagotchi/src/main/index.ts
```diff
@@ -88,12 +91,21 @@ if (isLinux) {
 app.dock?.setIcon(icon)
 electronApp.setAppUserModelId('ai.moeru.airi')
 
-initScreenCaptureForMain()
+let userFacingMainWindow: BrowserWindow | undefined
+const shouldStartMainProcess = installSingleInstanceGuard({ app, getWindow: () => userFacingMainWindow })
+
+if (shouldStartMainProcess) {
```
nekomeowww, line 104: Add comment to explain this.

PR #1856, opened 2026-05-20
fix(computer-use-mcp): bound waitForElement frame timeouts

fuyua9:
## Summary\n- Pass the remaining waitForElement budget into each frame-level CU_ACTION send so unresponsive frames cannot consume the fixed 8s sendMessage timeout.\n- Use the remaining deadline for each poll and stop polling immediately when the budget is exhausted.\n- Reduce the bridge-side waitForElement grace from the legacy 9.5s buffer to a small transport grace.\n- Add a regression test covering the hanging-extension case.\n\n## Validation\n- pnpm -C services/computer-use-mcp exec vitest run --config ./vitest.config.ts src/browser-dom/extension-bridge.test.ts


3361559784:
Approved and merging this one since the core timeout-budget fix is scoped and correct.

Follow-up request for a separate PR: `waitForElement` still has a small existing edge case around `frameIds: 0`. Some call sites use `cmd.frameIds || null`, which treats numeric `0` as falsy and can accidentally fall back to all frames instead of targeting the root frame explicitly.

This is not introduced by this PR and does not block the timeout fix, but it would be good to clean up in a follow-up by preserving `0` as a valid frame id, e.g. using an explicit nullish/empty-array check instead of `|| null`.

PR #1858, opened 2026-05-20
fix(core-agent): stabilize legacy chat timestamps

luyua9:
## Summary
- Backfill `createdAt` when a legacy user message without a timestamp is first projected into provider messages.
- Keep subsequent provider prompt timestamp prefixes stable instead of recalculating them from the current clock.
- Add a runtime regression test that sends twice with different clocks and verifies the legacy message keeps the original prefix.

## Why
`ChatHistoryItem.createdAt` is optional, but provider prompt composition currently uses `formatTimePrefix(createdAt ?? nowTs)`. For imported or older chat history entries without `createdAt`, rebuilding provider messages later can produce a different timestamp prefix for the same historical user message. That makes the prompt less stable across sends and can break the cache-friendly timestamp semantics.

## Validation
- `git diff --check`

Targeted Vitest was attempted in this sparse checkout, but `vitest` is not installed. Installing dependencies was blocked by repeated npm registry `ECONNRESET` failures, so the test could not be executed locally here.

### DIFF
--- a/packages/core-agent/src/runtime/chat-orchestrator-runtime.test.ts
+++ b/packages/core-agent/src/runtime/chat-orchestrator-runtime.test.ts
@@ -228,6 +228,47 @@ describe('createChatOrchestratorRuntime', () => {
     expect(harness.promptProjections).toHaveLength(1)
   })
 
+  it('keeps timestamp prefixes stable for legacy user messages without createdAt', async () => {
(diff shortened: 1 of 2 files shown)


nekomeowww (approved):

nekomeowww:
@codex review

Issue #1626, opened 2026-04-11
Updater doesn't delete C:\Users\actualuser\AppData\Local\ai.moeru.airi-updater after a successful update (IDK about mac or linux's equivilant)


jensenhuangfan:
### Describe the bug

It leaves the folder which has 2 copies of the update wasting over 1gb+ of storage and could be worse with more features and bigger updates overtime.

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (20) x64 Intel(R) Core(TM) Ultra 7 265K
    Memory: 37.07 GB / 63.47 GB
    GPU: NVIDIA GeForce RTX 5080
  Binaries:
    Node: 24.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 11.9.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.33.0 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Chrome: 146.0.7680.178
    Edge: Chromium (127.0.2651.74)
    Firefox: 149.0.2 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.8115
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/moeru-ai/airi/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/moeru-ai/airi/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [x] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

jensenhuangfan:
Still doesn't now but im too busy right this second to fix it. I'll fix it again soon but rn is not the best time

NeonKatVT:
I will vouch for this as well. I cleaned my Windows Registry today and found an empty key registry even though I uninstalled this earlier this month.

```
moeru
|
+--- airi
     (no key contents)
```

It's found under the user program registry, in the Windows stuff. It may be under `CurrentVersion`, I just don't remember where I found it.
`Computer\HKEY_CURRENT_USER\Software\Microsoft\Windows\`

[...]

PR #1867, opened 2026-05-23
fix(stage-tamagotchi): updater path and exit

jensenhuangfan:
## Description

This PR resolves several critical issues across different platforms relating to our auto-updater and launcher behavior:

1. **macOS Updater Hang**: Changed the `app.exit(0)` call in `index.ts` to `app.quit()`. This allows Electron's `before-quit` hooks to execute properly, enabling `electron-updater` to safely swap the binary payload and finish the installation instead of doing nothing when "Install" is clicked.
2. **Windows Silent Install Auto-Launch**: Added `runAfterFinish: true` to the NSIS configuration in `electron-builder.config.ts`. This ensures the application automatically launches after a silent background update completes.
3. **Legacy Cache Pathing**: Restored platform-specific OS cache resolution logic in `getLegacyCacheRoot()` (`auto-updater.ts`) to fix pathing errors and ensure backwards compatibility for older cache layouts during cleanup.

## Linked Issues

#1626  

## Additional Context

- Tested the Windows executable locally to confirm the silent update auto-launch flag works.
I have a 2012 macbook pro with OCLP now and I can test release versions of airi but its too weak to build the versions on my device so i cant test at the moment on the mac.


jensenhuangfan:
devs if you have any questions or things needed to be clarified, as always please let me know.

PR #1871, opened 2026-05-24
fix(server): sync character engagement counters on deletion

vuyua9:
## Summary
- Keep `deleteAllForUser()` character deletion cleanup in one transaction.
- When a user's active likes/bookmarks are soft-deleted, decrement the affected characters' cached `likesCount` / `bookmarksCount` counters without going below zero.
- Add a user-deletion regression covering counter rollback and retry idempotency.

## Why
Account deletion already soft-deletes the user's `characterLikes` and `characterBookmarks` rows. Without adjusting the denormalized counters, characters that were liked or bookmarked by the deleted account can keep stale engagement totals.

## Validation
- changed-file conflict marker scan clean
- Not run: targeted server tests, because this worktree has no `node_modules`; `pnpm --version` attempted a Corepack download and failed with `ECONNRESET`.


nekomeowww:
cc @luoling8192

PR #1887, opened 2026-05-27
fix(i18n): restore localized retry labels

wuyua9:
## Summary
- restore localized `chat.actions.retry` labels in the stage UI for French, Japanese, Korean, Vietnamese, and Traditional Chinese
- keep the change limited to existing locale string values

## Why
Current `main` still shows the English `Retry` label in these non-English stage locale files, while neighboring locales and the settings translations already use localized retry labels.

## Validation
- local assertion that each changed locale has a localized `retry` value and no longer has `retry: Retry`

### DIFF
--- a/packages/i18n/src/locales/fr/stage.yaml
+++ b/packages/i18n/src/locales/fr/stage.yaml
@@ -1,6 +1,6 @@
 chat:
   actions:
-    retry: Retry
+    retry: Réessayer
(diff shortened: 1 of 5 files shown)


Garfield550:
Crowdin will automatically override i18n that differs from their platform, so for i18n updates, please use Crowdin.

PR #1888, opened 2026-05-27
fix(stage-ui): bound sign-out requests

wuyua9:
## Summary

- add an 8s AbortController timeout around the stage-ui server sign-out request
- keep the existing server-side-first logout behavior when the request responds
- clear local auth state in a `finally` block so a hung `/oauth2/end-session` or `/sign-out` request cannot leave the UI stuck forever
- add unit coverage for the hung OIDC request, the responsive server path, and the bearer fallback signal wiring

## Testing

- conflict-marker scan on changed files
- `corepack pnpm -F @proj-airi/stage-ui test:run -- src/libs/auth.test.ts` *(blocked locally: this clean worktree has no `node_modules`, so `vitest` was not available)*

### DIFF
--- a/packages/stage-ui/src/libs/auth.ts
+++ b/packages/stage-ui/src/libs/auth.ts
@@ -173,30 +173,29 @@ export async function signOut() {
   // persistence (applyOIDCTokens started saving id_token in this branch);
   // without it, those legacy sessions would skip server cleanup and hit
   // exactly the silent-re-login bug described above.
-  try {
-    if (idTokenHint && clientId) {
-      const url = new URL('/api/auth/oauth2/end-session', SERVER_URL)
-      url.searchParams.set('id_token_hint', idTokenHint)
-      url.searchParams.set('client_id', clientId)
-      await fetchWithTimeout(url.toString(), { method: 'GET' })
-    }
-    else if (bearerToken) {
-      const url = new URL('/api/auth/sign-out', SERVER_URL)
-      await fetchWithTimeout(url.toString(), {
-        method: 'POST',
-        headers: { Authorization: `Bearer ${bearerToken}` },
-      })
-    }
-  }
-  catch {
-    // Network failure or timeout: still clear local state below. Server-side
-    // rows will expire by TTL; the local refreshToken/idToken/clientId are
-    // about to be wiped, so the local user has no way to spend them in the
-    // meantime.
-  }
-  finally {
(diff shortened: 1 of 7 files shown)


wuyua9:
Addressed the automated review on the sign-out timeout path.

What changed:
- a timed-out or failed server sign-out no longer clears local auth state, so the login surface is not exposed as logged out while the authoritative server session may still be live
- local auth state is cleared only after the OIDC end-session or bearer sign-out request resolves, or when no server-side credential is available
- added regression coverage for both OIDC and bearer timeout paths

Validation:
- `git diff --check` passed
- conflict marker scan on touched files passed
- `corepack pnpm --filter @proj-airi/stage-ui test -- packages/stage-ui/src/libs/auth.test.ts` could not run in this checkout because `node_modules` is absent and `vitest` is not installed locally

packages/stage-ui/src/libs/auth.ts
```diff
@@ -119,6 +119,25 @@ export async function listSessions() {
...
+    return await fetch(input, {
+      ...init,
+      signal: controller.signal,
+    })
+  }
+  finally {
+    clearTimeout(timeoutId)
+  }
+}
+
```
nekomeowww: I'd better to suggest to use the native `signal` parameter in `fetch` instead of writing additional timout wrapper on our own.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/RequestInit

nekomeowww (changes requested):
Please take a look at the comments

wuyua9:
Maintained this PR against current `main` and addressed the outstanding reviewer feedback in the sign-out path.

Changes in `0571fca6cd`:
- keep local auth state when OIDC or bearer sign-out times out, rejects, or returns non-2xx
- pass the timeout through `fetch`'s native `signal` option, with an iOS 15 `AbortController` fallback
- keep the timeout constant private and document the rejection contract
- surface sign-out rejection to the web, tamagotchi, and header-avatar consumers
- add regression coverage for OIDC, bearer, successful sign-out, and the iOS 15 abort fallback

Local validation:
- `pnpm --filter @proj-airi/stage-ui exec vitest run src/stores/auth.test.ts` (8 passed)
- targeted `moeru-lint` (5 files) passed
- `git diff --check` passed

The branch was rebased onto current `main`; the existing review decision is stale and CI is running on the new head. Please re-review `0571fca6cd`.

@codex review

Issue #957, opened 2026-01-15
feat(stage-*): auto discover websocket server


LemonNekoGH:

sebastian-zm:
We could probably implement this via server-advertised mDNS for all platforms except web. Or did you plan a brute-force approach to discovery? I'd like to see if I can help, but I think I need more concrete requirements.

sebastian-zm:
I've been testing mDNS with node and got this working. It advertises an IP on a known address so that from the browser we can simply connect to ws://airi-websocket-server.local:${port}/ We could do something like this in the server to allow discovery.

```
const WebSocket = require('ws');
const mdns = require('multicast-dns')();
const ip = require('ip');
const localIP = ip.address();

// Create and start WebSocket server on port 8080
const wss = new WebSocket.Server({ port: 8080 });

wss.on('connection', (ws) => {
  console.log('Client connected');
  ws.send('Connected to airi-websocket-server');

  ws.on('message', (message) => {
    console.log('Received:', message);
    ws.send(`Echo: ${message}`);
  });

  ws.on('close', () => {
    console.log('Client disconnected');
  });
});

wss.on('listening', () => {
  console.log(`WebSocket server running at ws://airi-websocket-server.local:8080`);
});

// mDNS - periodically announce the airi service and hostname
function announceService() {
  const timestamp = new Date().toISOString();
  console.log(`[${timestamp}] Announcing airi service and hostname...`);
  mdns.respond({
    answers: [
      {
        name: '_airi._tcp.local',
        type: 'PTR',
        ttl: 120,
        data: 'airi-websocket-server._airi._tcp.local'
      },
      {
        name: 'airi-websocket-server._airi._tcp.local',
        type: 'SRV',
        ttl: 120,
        data: {
          priority: 0,
          weight: 0,
          port: 8080,
          target: 'airi-websocket-server.local.'
        }
      },
      {
        name: 'airi-websocket-server.local',
        type: 'A',
        ttl: 120,
        data: localIP
      }
    ]
  });
}

// Handle mDNS queries and respond to TXT queries for our service
mdns.on('query', (query, rinfo) => {
  console.log(`[QUERY] From ${rinfo.address}:${rinfo.port}`);
  query.questions.forEach((q) => {
    console.log(`  Question: ${q.name} (type: ${q.type})`);

    // Respond to TXT record queries for our service
    if (q.name === 'airi-websocket-server._airi._tcp.local' && q.type === 'TXT') {
[...]

PR #1893, opened 2026-05-27
feat(server-runtime): mdns advertiser

sebastian-zm:
﻿> Edited by Claude (AI Assistant)

## Description

Adds opt-in mDNS auto-discovery advertising to `@proj-airi/server-runtime` so native clients (stage-pocket, stage-tamagotchi) on the same LAN can locate the AIRI WebSocket server without manual IP/port configuration.

## What is here

- `packages/server-runtime/src/discovery/mdns.ts` _(new)_ — New `createMdnsAdvertiser` module. Advertises the server under `_airi._tcp.local` via DNS-SD with a PTR / SRV / TXT / A record bundle (TXT schema: `txtvers`, `path`, `proto`, `auth`, `id`). Implements RFC 6762 §8.1 hostname conflict probing (3 × 250 ms) with numeric-suffix fallback, periodic re-announcement (75 s, 120 s TTL), and TTL = 0 goodbye packets awaited before socket destroy. Unique RRSets (SRV/TXT/A/AAAA) carry `flush: true` per §11.3 so peers evict stale records on restart; the shared PTR does not. Goodbye (TTL = 0) answers are ignored during probing so an own/peer departure is not mistaken for a name conflict. The `MdnsTransport` interface is injectable so tests can assert on packets without touching the network.

- `packages/server-runtime/src/server/index.ts` _(modified)_ — Wires the advertiser into the server lifecycle. Opt-in via `ServerOptions.mdns.enabled` or `MDNS_ADVERTISE=true|1`; instance name via `ServerOptions.mdns.serviceName` or `MDNS_SERVICE_NAME`. Handles three bind paths: IPv4 wildcard (`0.0.0.0`, advertises IPv4 addresses only), IPv6 wildcard (`::`, warns/skips when no IPv4 since `udp6` isn't supported yet), and a specific hostname (resolved via `dns.lookup` with a stop/restart guard; loopback and IPv6 binds are skipped with a warning). Skips with a `warn` (never crashes) when no usable, loopback-only, or link-local addresses exist, and stops a failed advertiser before clearing its handle. Exports three address helpers reused by stage-tamagotchi: `isLoopbackAddress` (`127.0.0.0/8` + `::1`), `isLinkLocalAddress` (`fe80::/10`), `isIPv4LinkLocalAddress` (APIPA `169.254.0.0/16`). `instanceId` is generated once per process (`nanoid` or `SERVER_INSTANCE_ID`) and reused across `restart()` so clients don't see a restart as a second server.

- `packages/server-runtime/README.md` _(new)_ — Package docs: usage, config table, mDNS section (enabling, service type, TXT schema, conflict handling, per-platform discovery commands and caveats), and `createServer` API reference.

- `packages/server-runtime/src/bin/run.ts` _(modified)_ — Forwards `hostname: env.HOST` to `createServer` so the CLI honours `HOST` (previously only `PORT`). When `HOST` is unset the `127.0.0.1` default is preserved (`@moeru/std` `merge` skips `undefined`).

- `packages/server-runtime/package.json` _(modified)_ — Adds `multicast-dns` (runtime) and `@types/dns-packet` / `@types/multicast-dns` (dev). `pnpm-lock.yaml` updated accordingly.

- `apps/stage-tamagotchi/src/main/services/airi/channel-server/index.ts` _(modified)_ — Drops the local `LOOPBACK_HOSTS` constant and rewrites `isLoopbackHost` to delegate to the exported `isLoopbackAddress` (now covering the full `127.0.0.0/8` range, plus `'localhost'`); the QR-code host filter now also excludes IPv6 link-local and IPv4 APIPA addresses.

## Tests

- `src/discovery/mdns.test.ts` _(new)_ — Probing happy path, conflict fallback, TXT coverage, PTR response, goodbye TTL = 0, cache-flush bit, and regressions (loopback self-conflict, goodbye-before-destroy ordering, stop/start reset, stop-during-probe error).
- `src/server/mdns-wiring.test.ts` _(new)_ — Default-disabled, env activation, loopback/no-address skips, IPv4/IPv6 wildcard paths, hostname lookup concurrency guard, `restart()` instanceId reuse, `closeServer()` cleanup, and loopback-alias / APIPA filtering regressions.
- `src/server.test.ts` _(new)_ — The three address helpers.

## Open Questions for Reviewers — Naming & Service Registration

A few naming decisions in this PR are worth a closer look before merge, since
they become part of a wire-visible contract once clients start browsing:

- **Hostname.** The server currently advertises under `airi-websocket-server.local`
  (default `serviceName`, configurable via `ServerOptions.mdns.serviceName` /
  `MDNS_SERVICE_NAME`).

- **Service type.** The DNS-SD service type is `_airi._tcp` — i.e. a single
  custom type for the whole product. We could instead scope it per protocol,
  e.g. `_airi-ws._tcp` for the WebSocket server, leaving room for future types
  like `_airi-model._tcp` for a possible model runtime. Worth deciding now,
  because changing the type after clients ship is a breaking change.

If this PR is merged we would be advertising under `_airi._tcp.local` via
DNS-SD. The service name `airi` is **not** currently registered with IANA. A
search of the registry returns only fuzzy matches (`iRAPP`, various `*pairing`
services) — no real collision — so the name appears free to claim.

We could register the service name `airi` (the DNS-SD type) with IANA, with our
TXT keys (`txtvers`, `path`, `proto`, `auth`, `id`) documented alongside it. We
would request **no port number** — port assignment goes through Expert Review,
and IANA will (correctly) tell us to use a configurable port / the dynamic
range, which is already how this PR behaves.

## Linked Issues

[Contributes towards completing this.](https://github.com/moeru-ai/airi/issues/957)

## Additional Context

Provides the server-side advertising half of auto-discovery (both the server listener and the AIRI service are announced via mDNS DNS-SD); client-side browsing/discovery is left as follow-up. IPv6-only discovery (`udp6` / `ff02::fb`) is not yet supported and is skipped with a warning.


sebastian-zm:
@codex please go through each issue marked as resolved and check whether it’s actually fixed in the current code. For each one, tell me: resolved / not resolved / unclear, with the file and line as evidence. Flag anything that’s only partially addressed.

sebastian-zm:
Something reviewers should look at is the host and protocol names. Currently in this PR `airi-websocket-server.local` is advertised. Also the protocol is called `_airi._tcp` as a custom TCP protocol. We could change that to `_airi-ws._tcp` in case we have more protocols later e.g. `_airi-model._tcp` for a possible future model runtime or something.

sebastian-zm:
If this PR is merged we would advertise under `_airi._tcp.local` via DNS-SD. The service name `airi` is still **not** registered with IANA. A registry search returns only fuzzy matches (`iRAPP`, `*pairing` services), no real collision, so the name is free to claim.

We could register the service name `airi` (the DNS-SD type), with our TXT keys documented. No port number requested. Port assignment goes through Expert Review and IANA will (correctly) tell us to use a configurable port / the dynamic range.

If you agree, go to IANA's website and register `airi` there.

sebastian-zm:
@codex review

sebastian-zm:
I'm sorry, I got tired of keeping this PR up-to-date, specially when I just wanted to help, and I'm not even interested in the feature. I'm closing it.

PR #1899, opened 2026-05-29
fix(stage-ui): limit provider validation to listed providers only

lulu0119:
## Summary

- `updateConfigurationStatus` only runs `validateProvider` for providers where `shouldListProvider` is true (added in settings or credentials differ from defaults).
- `startPeriodicRuntimeValidation` uses the same filter so interval-based revalidation (e.g. LM Studio) does not hit localhost for providers the user never added.
- Removes the old TODO and the previous “validate all providers on startup” behavior that caused noise (e.g. Ollama/LM Studio on cold load).

We do not set `providerMetadata[id].configured` after validation because metadata is the static provider catalog, while probe results are per-browser runtime state. `providerRuntimeState.isConfigured` is already the source of truth; UI reads it via computed projection. Writing into metadata would duplicate that flag and drift from `isConfigured`.

## Test plan

- [ ] Cold-load stage-web: no HTTP to `11434` / `1234` unless LM Studio/Ollama (or similar) is listed.
- [ ] Add/list a local provider, configure base URL: startup or credential watch should probe; periodic revalidation only for that provider.
- [ ] Open LM Studio settings page only (not listed): requests may still come from `useProviderValidation` on that page; store-wide probe should not include `lm-studio` until listed.
- [ ] `pnpm -F @proj-airi/stage-ui typecheck`

### DIFF
--- a/packages/stage-ui/src/stores/providers.ts
+++ b/packages/stage-ui/src/stores/providers.ts
@@ -2436,11 +2436,15 @@ export const useProvidersStore = defineStore('providers', () => {
       }))
   }
 
-  // Call initially and watch for changes
-  watch(providerCredentials, updateConfigurationStatus, { deep: true, immediate: true })
-  startPeriodicRuntimeValidation()
+  async function refreshListedProviderValidation() {
+    await updateConfigurationStatus()
+    startPeriodicRuntimeValidation()
+  }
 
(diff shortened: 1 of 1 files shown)


lulu0119:
OpenAI-compatible generic validation: Connectivity and ModelList both call the same `/models` endpoint, so each validate round issues duplicate requests.

lulu0119:
@codex review

PR #1900, opened 2026-05-29
fix(stage-ui): onboarding repeat prompt when provider credentials exist

shojikumaru:
## Summary

Fix the first-time setup dialog reappearing after reload when provider credentials already exist in local storage.

Currently, onboarding visibility depends on `onboarding/completed` / `onboarding/skipped`, but it does not account for already-persisted essential provider configuration. If a user has provider credentials saved but the onboarding completion flag is false or missing, the setup dialog can be shown repeatedly.

This change makes `needsOnboarding` also check:

- `hasEssentialProviderConfigured`
- `hasEssentialProviderCredentialConfigured`

so an already configured essential provider prevents the setup dialog from being forced again.

## Validation

Validated locally on a Mac mini deployment:

- `pnpm --filter @proj-airi/stage-ui typecheck` passes
- UI health check returned `200`
- API `/readyz` returned `200`
- Browser reload test confirmed the setup dialog does not reappear when an essential provider credential is present even if `onboarding/completed=false`

## Notes

This is intentionally a minimal fix: it does not change provider storage or onboarding persistence format.


nekomeowww (approved):

shojikumaru:
Thanks for the review, and thanks to the Codex bot for both findings — they were real.

- P1: `saveProviderConfiguration()` persists the API key at the provider step, before model selection and before `configured` is emitted. With a reactive condition the `needsOnboarding` watcher would close the dialog mid-flow and `onboarding/completed` would never be set. Fixed by taking a one-time, non-reactive snapshot of the validated provider status at store creation instead.
- P2: raw credential presence is not validation. The condition now uses only the persisted `status === 'configured'` (restored synchronously from localStorage), so an azure-openai record with an apiKey but no baseUrl no longer suppresses first-time setup. `hasEssentialProviderCredentialConfigured` is left exported and unchanged.

Added three unit tests (startup snapshot suppresses; a provider becoming configured mid-flow does not close the dialog; azure-shaped credentials without validation still require onboarding). The two regression tests fail against the previous reactive version.

Known narrow edge, intentionally out of scope: a logged-out user whose only essential provider was auth-owned and who never completed onboarding gets one page load without the dialog (the snapshot is taken before `removeAuthenticationProviderConfiguration()` runs in `onMounted`); it appears on the next reload.

Rebased onto current main (no conflicts). Verified locally with `NODE_OPTIONS=--no-experimental-webstorage` as in CI: 4/4 tests, `vue-tsc` and `moeru-lint` clean.

@nekomeowww This is rebased on today's main and the bot's findings are addressed; the approval predates the follow-up commit, so a quick re-look would be appreciated whenever you have a moment. Happy to adjust if you'd prefer a different shape.

PR #1911, opened 2026-05-31
fix(stage-ui): preserve Kokoro worker on abort

wuyua9:
## Summary

- keep Kokoro generation cancellations out of the worker failure recovery path
- restore the loaded adapter state to `ready` when an in-flight `generate()` call is aborted
- add regression coverage that generation abort posts a worker `cancel` message without terminating/restarting the worker

## Testing

- conflict-marker scan on changed files
- `corepack pnpm -F @proj-airi/stage-ui test:run -- src/libs/inference/adapters/kokoro.test.ts` *(blocked locally: this clean worktree has no `node_modules`, so `vitest` was not available)*

### DIFF
--- a/packages/stage-ui/src/libs/inference/adapters/kokoro.test.ts
+++ b/packages/stage-ui/src/libs/inference/adapters/kokoro.test.ts
@@ -166,6 +166,45 @@ describe('kokoro adapter - device loss resilience', () => {
     }))
   })
 
+  it('should not restart the worker when generation is aborted', async () => {
+    const { createKokoroAdapter } = await import('./kokoro')
+    const adapter = createKokoroAdapter()
+
+    const loading = adapter.loadModel('q4', 'webgpu')
+    await vi.waitFor(() => expect(enqueueMock).toHaveBeenCalled())
+
+    const worker = MockWorker.instances.at(-1)!
+    const loadRequest = worker.postMessage.mock.calls.find(([message]) => message.type === 'load-model')?.[0]
+    expect(loadRequest?.requestId).toBeDefined()
+    worker.dispatch('message', {
+      data: {
+        type: 'model-ready',
+        requestId: loadRequest.requestId,
+        device: 'webgpu',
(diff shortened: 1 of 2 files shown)


sebastian-zm:
Basically what codex said. The reason why I made the kokoro worker restart was because the worker cannot interrupt an in-flight transformers.js inference. With your PR the worker would queue the next request, and it wouldn't happen until the first one finishes.

My concern is the wedged-worker case: if someone cancels _because_ generation is stuck, resetting to ready means recovery is deferred to the 120s `GENERATE_TIMEOUT` on the next request (then restart delay + reload) — whereas restart-on-cancel recovered immediately.

I think we could keep the warm-worker fast path but add a short watchdog so a worker that doesn't drain the cancelled request within a few seconds still gets restarted? That keeps both cheap normal cancels and fast recovery from a real hang.

sebastian-zm:
> By Claude 4.8 Opus

Separate from the worker-overlap discussion — the `generate()` abort path this PR adds is never exercised in production.

The only production caller is the Kokoro provider's `fetch` shim (`providers.ts:2073`):

```ts
const buffer = await (await workerManagerPromise).generate(text, voice)
```

It passes only `text`/`voice` and never forwards `init.signal`, so in the real TTS path `options?.signal` is always `undefined`:
- `throwIfAborted` is a no-op,
- `waitForWorkerMessage` never registers an abort listener,
- no `cancel` is ever posted to the worker.

So `generate()` can only reject with `notReadyError`, the `GENERATE_TIMEOUT`, or a worker error — never `AbortError`. The new *"should not restart the worker when generation is aborted"* test only passes a signal because it calls `generate()` directly; it can't catch a real regression in the TTS flow until `init.signal` is wired through at `providers.ts:2073`.

(The load path is fine — `use-inference-preload.ts` does forward a signal to `loadModel`, so that abort guard is genuinely reachable.)

If cancelling stuck speech is the goal, forwarding the request signal into `generate()` is the missing piece — and it ties into the wedged-worker concern above, since only then would a real cancel reach the worker.

PR #1914, opened 2026-05-31
fix(stage-ui): restore default stage model when saved model is missing

NashChennc:
## Summary

Fixes the tamagotchi app getting stuck on the loading screen when `localStorage` points to a custom stage model that no longer exists in IndexedDB.

## Changes

- Adds a shared default stage model id, `preset-live2d-1`.
- Falls back to the default Live2D model when the saved stage model cannot be resolved.
- Persists the restored default selection back to `localStorage`.
- Adds a regression test for missing persisted model data.

## Root Cause

The renderer could become `disabled` when `settings/stage/model` referenced a stale custom model id. Since no valid stage renderer was mounted, the app stayed on the loading screen.

## Testing

- `pnpm -F @proj-airi/stage-ui exec vitest run src/stores/settings/stage-model.test.ts`
- `pnpm exec moeru-lint packages/stage-ui/src/stores/settings/stage-model.ts packages/stage-ui/src/stores/settings/stage-model.test.ts`
- Verified `pnpm dev:tamagotchi` loads successfully with the default Hiyori model.

Note: root `pnpm lint` still reports existing unrelated repo-wide issues.

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/App.vue
+++ b/apps/stage-tamagotchi/src/renderer/App.vue
@@ -197,9 +197,8 @@ function createFullStageRuntime() {
       cardStore.initialize()
 
       await displayModelsStore.loadDisplayModelsFromIndexedDB()
-      await settingsStore.initializeStageModel()
-      await settingsAudioDeviceStore.initialize()
-
+      // NOTICE: Restore the Godot renderer before model recovery so deleted custom VRM ids
+      // fall back to a VRM preset instead of the built-in Live2D preset.
       if (isGodotStageRoute()) {
         try {
           syncGodotStageRenderer(await getGodotStageStatus())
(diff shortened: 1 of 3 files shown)


0xSelenicDove:
Thanks for this fix. The core stale-stage-model recovery is now covered by #2033, which restores `preset-live2d-1` when a persisted custom model is missing.

This PR also appears to include an additional Godot-specific case: restoring the Godot renderer before resolving the saved model, then falling back to `preset-vrm-1` rather than the Live2D preset. Could you confirm whether that scenario still reproduces on current `main`?

If it does, please open a small rebased follow-up containing only the Godot-specific ordering/fallback change and its regression test. Otherwise, this PR can be closed as superseded by #2033.

0xSelenicDove:
Closing as superseded by #2033, which merged the stale persisted-stage-model recovery.

#1914 may include a separate Godot-specific fallback case (`preset-vrm-1`). If that still reproduces on current `main`, please open a small rebased follow-up containing only that case and its regression test.

PR #1921, opened 2026-06-01
feat(stage-ui): upgrade MiniMax default model to M3

octo-patch:
## Summary

Upgrade MiniMax chat provider configuration to make MiniMax-M3 the new default model.

## Changes

- Add `MiniMax-M3` to the model selection list as the new default (first entry)
- Retain `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` for backward compatibility
- Remove deprecated older variants from the list: `MiniMax-M2.5`, `MiniMax-M2.5-highspeed`, `MiniMax-M2.1`, `MiniMax-M2.1-highspeed`, `M2-her`, `MiniMax-M2`

Scope is limited to `packages/stage-ui/src/libs/providers/providers/minimax/index.ts`. The TTS provider (`minimax-speech`) is untouched, and the volcengine-coding-plan provider which exposes a `minimax-m2.5` proxy entry is also untouched (it lives behind that vendor's ARK gateway).

## Why

MiniMax-M3 is the new flagship model with 512K context, 128K max output, and image input support. The older M2.5/M2.1/M2 lines are no longer the recommended path; M2.7 remains available as a stable previous-flagship alternative.

## Testing

- Static verification: scanned the repo for references to removed model IDs (`MiniMax-M2.5`, `M2-her`, etc.) — none remain outside this provider file
- Model list shape unchanged (still `ModelInfo[]`); only the entries inside the array changed
- Both `minimax` (CN) and `minimax-global` providers share the same `minimaxModels` array, so the upgrade applies to both

### DIFF
--- a/packages/stage-ui/src/libs/providers/providers/minimax/index.ts
+++ b/packages/stage-ui/src/libs/providers/providers/minimax/index.ts
@@ -31,52 +31,22 @@ type MinimaxGlobalConfig = z.input<typeof minimaxGlobalConfigSchema>
 
 const minimaxModels: ModelInfo[] = [
   {
+    id: 'MiniMax-M3',
+    name: 'MiniMax M3',
+    provider: 'minimax',
+    description: 'Latest flagship model with 512K context, 128K max output, and image input support',
+  },
+  {
     id: 'MiniMax-M2.7',
     name: 'MiniMax M2.7',
     provider: 'minimax',
-    description: 'Latest flagship model with enhanced reasoning and coding',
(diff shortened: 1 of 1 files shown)


nekomeowww:
Rebase is required.

octo-patch:
Closing this — #1930 just landed and moved model selection to fetch dynamically from `v1/models`, so the hardcoded list this PR edits is gone and M3 now shows up automatically once it's live on the API. No rebase needed, thanks for the heads-up!

PR #1925, opened 2026-06-01
chore(ci): use personal access token for hash update

Weathercold:
## Description

Unfortunately the method of manually calling CI on the PR branch used by #1907 doesn't count as status check of that PR (#1924), probably because the called workflow inherits the context of the calling workflow, which is not triggered by the PR. I read the docs and the most straightforward solution is using a PAT, which does trigger the pull_request event. Alternatively you can open PR with a custom github app, which is more complicated but allows the app to bypass ruleset and commit directly to main. Or just disable mandatory status checks.


.github/workflows/update-nix-assets-hash.yaml
```diff
@@ -32,6 +30,8 @@ jobs:
       - uses: actions/checkout@v6
         with:
           ref: main # Use main regardless of workflow_dispatch branch
+          # Authenticate git with PAT so that pushing to head retriggers CI
+          token: ${{ secrets.HASH_UPDATE_TOKEN }}
```
lietblue, line 34: `actions/checkout` persists credentials by default (`persist-credentials: true`; see https://github.com/actions/checkout/blob/v6/action.yml#L52-L54). Passing `HASH_UPDATE_TOKEN` here leaves the PAT available for authenticated git commands in later steps, including the Nix hash update scripts. Since this is a long-lived PAT, please set `persist-credentials: false` and only use the token in the PR creation/push step.

Weathercold: Is there a problem with using PAT to push the commit? It's needed to retrigger pull_request when it's already created. Also, Nix build doesn't use the PAT.

lietblue: The overall approach looks fine, but passing `HASH_UPDATE_TOKEN` to `actions/checkout` means checkout will keep it available for later authenticated git commands until the post-job cleanup. Since we only need the PAT for the final push / PR step, maybe set `persist-credentials: false` and use the token there instead.

Weathercold: The easiest way to make git push use the PAT is by configuring it with actions/checkout. The git commands are only run in the last step, so it doesn't matter in which step we configure git.

lietblue: ok lgtm rebase and merge

lietblue (approved):

PR #1927, opened 2026-06-02
fix(stage-ui-live2d): normalize Live2D archive paths for import and display

NashChennc:
## Summary

- Ignore macOS/Windows metadata files when validating and loading Live2D zip archives.
- Sanitize malformed `model3.json` optional references such as `Physics: null`.
- Normalize non-ASCII Live2D resource paths to match `pixi-live2d-display` FileLoader validation.
- Prevent OPFS-reconstructed settings files from double-encoding already-normalized paths.
- Add OPFS cache versioning so stale cached Live2D settings are rebuilt.
- Register `packages/stage-ui-live2d` in the root Vitest project list.

## Testing

- `pnpm exec vitest run packages/stage-ui-live2d/src/utils/live2d-archive-paths.test.ts packages/stage-ui-live2d/src/utils/live2d-validator.test.ts`
- `pnpm -F @proj-airi/stage-ui-live2d typecheck`
- `pnpm exec moeru-lint <touched files>`
- `git diff --cached --check`

### DIFF
--- a/packages/stage-ui-live2d/src/utils/live2d-archive-paths.test.ts
+++ b/packages/stage-ui-live2d/src/utils/live2d-archive-paths.test.ts
@@ -1,0 +1,342 @@
+import JSZip from 'jszip'
+
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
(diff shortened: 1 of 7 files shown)


NashChennc:
The root cause of the bug has not been identified.

PR #1949, opened 2026-06-05
fix(stage-ui): robust spark:command result + TTS session isolation

FlowerWater1019:
## Summary

Two small, generic robustness fixes, split out of the Minecraft desktop integration (#1916) so they can land independently of that rework — as suggested in @shinohara-rin's review:

> The generic fixes in this PR, like the TTS session cleanup and `spark:command` result handling, seem separable and worth keeping in a smaller PR.

Both are in shared `stage-ui` runtime code and contain **no Minecraft-specific logic**.

## What's included

- **`spark:command` result handling** (`tools/character/orchestrator/spark-command.ts`): `command.destinations` may be `undefined` — the channel sender (`stores/llm.ts` `sendSparkCommand`) deletes it to broadcast to all authenticated peers. The success message's `.join()` therefore surfaced `Cannot read properties of undefined (reading 'join')` back to the LLM even though the send had already succeeded. Guard it and report a broadcast instead.
- **TTS session isolation** (`components/scenes/Stage.vue` `openTtsSession`): a completing/erroring session now only clears the module-level `currentSession` if it **is** that session. The previous code cleared it whenever any `stream-` session completed, which becomes unsafe once sessions exist that are not assigned to `currentSession` (e.g. one-off read-aloud sessions) — one of those finishing would null a still-active chat session and drop the rest of the reply.

## How tested

```bash
pnpm -F @proj-airi/stage-ui exec vitest run src/tools/character/orchestrator/spark-command.test.ts  # 9 passed (+1 new)
pnpm -F @proj-airi/stage-ui typecheck   # 0 errors
pnpm exec eslint <changed files>        # 0 problems
```

- Regression test for the broadcast-destinations result message.

## Context

This is the first, low-risk slice of reworking the Minecraft↔desktop integration around the neutral Context Flow architecture (per the #1916 review). The Minecraft relay/read-aloud behavior will be reintroduced through a Minecraft-owned adapter in a follow-up.


FlowerWater1019:
**Merge-order note (part of a 3-PR Minecraft rework split).**

This work is split into three PRs:
- **#1949** (this) — generic `stage-ui` robustness fixes.
- **#1950** — restore `services/minecraft` to neutral Minecraft semantics.
- **#1916** — reintroduce desktop relay & read-aloud via a Minecraft adapter.

This PR and **#1950** are **independent of each other** (different areas) and can merge in **any order**. Both should land **before #1916**, which is stacked on them. This PR has no dependency on #1950 — it's purely the generic fixes you noted were separable.

FlowerWater1019:
@shinohara-rin This one is rebased onto the latest `main` and green — `MERGEABLE` (the `BLOCKED` state is just the fork-PR workflow gate). It's an independent generic fix (robust `spark:command` result handling + TTS session isolation) with no ordering dependency, so it can merge whenever you've had a look.

Merge-order note for the stack: **#1949 and #1950 are independent and can merge in any order; #1916 must merge last** (it's stacked on both — once they land I'll rebase #1916 so their commits drop by patch-id and its diff reduces to just the Minecraft adapter). Please take a look and merge when you're happy with it. 🙏

shinohara-rin (approved):

PR #1956, opened 2026-06-07
fix(nix): missing runtime libs & small changes

Weathercold:
## Description

Add the missing runtime libs to LD_LIBRARY_PATH. Also pin pnpm and electron versions, use getExe and update flake lock.


Weathercold:
I'll just merge this since the changes are small

PR #1965, opened 2026-06-11
feat: quiet screen observation & task progress tracking via screenpipe

Fire-friend:
## Summary

This PR teaches AIRI to follow through on goals you give it. You set a task or schedule in plain language in chat ("finish the quarterly report by 5pm, keep an eye on my progress"), and AIRI:

1. confirms what it understood — and exactly **which apps it will watch** — in a task card,
2. quietly observes your screen through a local [screenpipe](https://github.com/mediar-ai/screenpipe) instance, scoped strictly to the apps you whitelist,
3. tracks progress against the goal and plans next steps, and
4. reaches you through the least intrusive channel that works — by default just a small character gesture, plus an optional daily wrap-up.

Everything stays local: screenpipe stores recordings on the user's machine, and only task-scoped summaries — never screenshots, never raw OCR text — leave the capture layer.

## Product decisions baked in

- **Default touch level = L1** (character gesture; no windows, no notifications). One deliberate exception: a brand-new user's first task delivers its first progress update at L2 once, so they can see the feature actually works, then falls back to L1.
- **Observation scope = whitelist.** An empty whitelist means observation is OFF, and the UI says so explicitly ("No apps selected — not currently observing"). There is no silent state where the switch is on but nothing is watched.
- **Daily summary default ON** (18:00 local, configurable). Zero tasks that day → no summary is sent.
- **L3 (system notification) is reserved** for user-set hard deadlines or clearly blocked tasks; never focus-stealing, never modal.
- **Escalation rules are constants, not model judgment**: at most one L2+ touch per task per 30 minutes; two consecutively ignored touches downgrade a level (never auto-upgrade); fullscreen/presentation/meeting suppresses everything to L0, summarized afterward. An unclicked L3 system notification does not count as "ignored" — Electron's notification close events are unreliable cross-platform, so the deterministic ignore signal comes from the L2 notice path only.
- **No bare percentages anywhere.** Progress copy must say what's left in actionable units, the projected completion at the current pace, and whether the task is off-track. "75%" is mechanically intercepted, not style-guided away.

### Shared contract & decision core — `packages/server-sdk-shared`

- New `screen-observation.ts`: task/schedule model, `ScreenObserverSummary`, L0–L3 `TouchEventPayload`, and the Eventa event contracts.
- All behavior decisions are **pure functions in a single place** (`resolveObservationPrivacyState`, `normalizeScreenObserverSummary`, `decideScreenObservationTouch`, `decideDailySummary`, …), consumed by both the server and the Electron main process — no mirrored copies anywhere.
- The escalation constants above live here, alongside the progress-copy guard (`isBarePercentage` + `normalizeTouchMessage`): bare-percentage or empty progress strings are replaced with humane fallback copy at the decision layer, and presentation surfaces reuse the same guard.
- `apps/server` keeps only a thin wrapper (id stamping), per its no-background-loop deployment constraint; screenpipe polling lives in the Electron main process.

### Electron main process — `apps/stage-tamagotchi`

- `ScreenObserver` runtime (`src/main/services/airi/screen-observer/`): polls the local screenpipe REST API (30s tick; auto-degrades and retries when `/health` fails) and aggregates OCR into contract summaries.
- **Whitelist enforced as a capture-side hard boundary, at three layers**: OCR queries are always app-scoped — `SearchOcrParams.appName` is required at the type level, so an unscoped OCR query is unrepresentable; aggregation drops non-whitelisted entries before any output; and meeting/fullscreen detection uses a `focusedWindow()` metadata probe that reads only `app_name`/`window_name` and never materializes OCR text.
- Pause/privacy state machine: master switch off by default; pause for 15 min / 1 h / rest of today, persisted across restarts; meeting/browser-fullscreen auto-suppression; global pause hotkey (default `CmdOrCtrl+Alt+P`).
- Tray: the first menu item is always the current observation state verbatim (with resume time while paused); quick pause/resume; direct link to the privacy settings page.
- L3 system notifications with humane progress copy; clicking routes to task details. A persistent ledger drives the 30-minute throttle and ignore-downgrade rules across restarts.
- i18n for all main-process strings across 9 locales.

### Renderer

- `ChatTaskCard`: natural-language task setup ends in a confirmation card — goal / time window / **what AIRI will watch** / touch level — tweakable inline ("make it 6pm" just updates the card; no form).
- Dashboard (previously a stub): observation status bar, task list (status dot + name + progress phrase + time remaining), and the daily summary view — per task: progress, a one-line observation, and a suggestion for tomorrow.
- Settings → new `screen-observation` privacy module: master switch, whitelist editor, an unskippable three-screen first-run explainer (what it watches → where data goes → how to pause), auto-pause rules, and a "what it saw" log showing time window + app + purpose + the actual summary consumed (not screenshots), with one-click deletion of today's data.
- Notice window: one-line humane progress toast reusing the existing fade-on-hover behavior, expanding to ack / details / mute-this-task.
- A renderer store mirrors main-process state over the IPC snapshot/event contract; empty-whitelist copy is sourced from one place across settings, dashboard, and the task card. Renderer i18n ships en + zh-Hans; other locales fall back to en.

## Privacy model (worth reviewing closely)

- screenpipe runs and stores everything **locally**; only task-scoped summaries leave the capture layer.
- Non-whitelisted apps' OCR never enters the process — enforced at the query (type level), aggregation, and detection-probe layers, with tests asserting that non-whitelisted app names, window titles, and OCR text appear nowhere in the output.
- Observation is off by default and pausing is always one click away (tray, character context menu, hotkey), with state always visible in the tray.

## Testing & verification

- Targeted suites all green at HEAD: server domain tests 11/11, stage-tamagotchi screen-observer tests 24/24, stage-ui store/progress-language tests 10/10; build + typecheck pass for `server-sdk-shared`, `server`, `i18n`, and `stage-ui`.
- An independent QA gate reviewed the branch twice. The first pass found 2 HIGH issues — a whitelist bypass through the meeting-detection OCR query, and L3 notifications bypassing the bare-percentage guard. Both were fixed and re-verified closed at HEAD `ec4837b09`; no CRITICAL/HIGH remain.
- The frozen defaults, escalation/downgrade rules, empty-whitelist state, zero-task summary suppression, and L3-ignore semantics were each verified against the spec.
- Branch history is linear: `bc0c37f61` (shared/server contract baseline) → `b60a78586` (desktop + renderer integration) → `a76c88b5e` (shared touch-message guard) → `ec4837b09` (capture-boundary hardening + L3 formatter guard).

## Known follow-ups (non-blocking)

1. **LOW** — `normalizeTouchMessage`'s fallback strings are English literals in shared; they should be aligned to the established `{ key, params }` + presentation-layer-translation pattern. Only reachable on degraded input (empty or bare-percentage upstream copy), and the L3 surface already localizes its own fallback.
2. **Environment** — a full Electron build is currently blocked by pre-existing missing workspace dists (`crossws/server` via `packages/server-runtime`), unrelated to this change. Targeted builds/typechecks/tests above all pass, and the typecheck error set has zero additions versus the pre-existing baseline.

3. **Scope** — Natural-language task creation in chat (chat → parse → TaskCard confirm → `upsert-task`) is not wired yet — the `upsert-task` IPC contract exists but has no caller and TaskCard isn't mounted in InteractiveArea. Tasks currently enter via the main-process registry and flow to the dashboard. Separate chat-orchestration follow-up.

Also noted for later: "any-app fullscreen" L0 suppression needs a native probe Electron can't provide (a `setSuppressionSignals` external hook is in place; meeting/browser-fullscreen heuristics cover the common cases), and the macOS tray state glyphs (●/⊘) are placeholders pending design assets.

## What this means for users

Tell AIRI what you're working on in plain language, confirm the card it shows you — including exactly which apps it may watch — and get on with your day. It observes only the apps you whitelist, keeps all recordings on your machine, and by default never pops a window: your character just gestures when there's news. At the end of the day you get a short wrap-up — what got done, what AIRI noticed, and what to pick up tomorrow. Pausing is always one click away in the tray, and the tray always tells you whether it's watching.


Fire-friend:
Per repo owner — retargeting to a fork-internal PR for their own development; not upstream.

PR #1827, opened 2026-05-12
ci(stage-tamagotchi): add Steam desktop deployment workflow

Neko-233:
## Description

Adds Steam as a distribution target to the existing Tamagotchi release workflow.

- Runs Steam builds from the existing Release Tamagotchi workflow_dispatch entry.
- Builds Windows x64, Apple Silicon macOS, and Linux x64 Steam depot content.
- Uploads Steam builds to the selected branch, defaulting to internal-test for validation.
- Serializes Steam build and deploy runs per Steam branch.
- Disables custom providers, Flux purchasing, and the GitHub updater in Steam builds.
- Shows store-managed update messaging and hides the inactive update-channel selector.

## Linked Issues

None.

## Additional Context

Ordinary GitHub Release, scheduled build, GitCode mirror, and itch.io publishing paths remain separate from the manual Steam distribution path. The standalone Steam workflow was removed so the release pipeline has one maintained entry point.

Validation:

- actionlint v1.7.12
- pnpm exec vitest run apps/stage-tamagotchi/src/main/services/electron/auto-updater.test.ts (31 tests)
- pnpm -F @proj-airi/stage-tamagotchi typecheck
- pnpm typecheck
- pnpm lint (0 errors; 6 unrelated existing warnings)
- Steam-configured electron-vite production build

PR #1966, opened 2026-06-13
feat(steam): add desktop silent sign-in on top of Steam depot CI (#1827)

lulu0119:
## How to review this PR

Large surface, but three layers. Prefer reading **top → bottom**:

1. **Feature** — Steam-anchored desktop auth + packaging (the product)
2. **Review harden** — follow-ups from PR review (safety / contract / hygiene)
3. **Regression fixes** — bugs found while verifying those hardenings on `server-dev` / Steam `internal-test`

Commit history is long (debug markers, autofix, Railway boot fixes). Focus on the layers below, not every intermediate commit.

---

## 1) Feature — Steam desktop silent sign-in

**Goal:** Steam launch either logs in silently (linked SteamID) or opens enrollment in the system browser (unlinked). No unlinked AIRI account from Steam alone.

### Client (`stage-tamagotchi`)
- Steamworks FFI ticket → `POST /api/auth/steam/desktop-sign-in`
- Linked → OIDC tokens into renderer (`fetchSession`)
- Unlinked → open auth UI `/enroll` with single-use `enrollToken` + OIDC authorize `continue`
- Steam redistributables / depot packaging; macOS SDK path + entitlements so packaged Steam builds actually init

### Server
- Ticket validation, ownership check, link lookup
- Unlinked → `403 STEAM_NEEDS_ENROLLMENT` + enroll token (no placeholder user)
- OIDC authorize choke point: consume token → `linkSteamToUser` → issue code

### Auth UI (`ui-server-auth`)
- `/enroll` reuses email/OAuth step forms
- Email verify resume for enroll (Electron owns PKCE on loopback)
- Steam shown as read-only connected sign-in method (no web link/unlink)

### CI / release
- Unified `Release Tamagotchi` with `distribution=steam`
- Steam `internal-test` packages point at Railway **server-dev** API

---

## 2) Review harden — PR comment follow-ups

Main batch: `55573c56e` and nearby. Intent is fail-closed / hygiene, not new product scope.

| Area | What changed |
|------|----------------|
| Enrollment token | Fail closed on invalid/expired/consumed token before issuing codes; type-guarded deletes |
| Login redirect | Re-attach `enrollToken` on trusted Better Auth login redirects so mid-enroll session loss can retry |
| Continue URL | Trusted-origin checks for enroll / verify-email resume |
| Steam ticket | Cancel Web API ticket after exchange |
| Desktop Steam | Isolate Steam `userData` from direct builds; disable GitHub updater on Steam builds |
| Enroll UX | Hide forgot-password during enroll; debounce concurrent Steam exchanges |
| Silent login contract | `d2250a1ad` — server issues short-lived OIDC **code**; client does PKCE → `/oauth2/token` (no server-minted full tokens) |
| Tests | `3def8388c` — trim overlapping enroll/link coverage |

---

## 3) Regression fixes — after harden / verify on server-dev

Caught while exercising the review follow-ups (not new features):

| Commit | Symptom | Fix |
|--------|---------|-----|
| `3555f131c` | After email verify, browser showed `STEAM_ENROLLMENT_TOKEN_INVALID` while Electron still logged in | Success tab + pending tab both hit authorize; first consumes token, second 403'd. Idempotent retry if Steam already linked; Steam enroll pending tab no longer double-navigates |
| `9e3e11b48` | Onboarding “Sign in / register” closed the window but never opened `/enroll` | Silent startup exchange held `steamSignInInFlight`; user gesture was dropped. User path now waits for the in-flight attempt, then opens enroll |

---

## Out of scope / noise (ignore when reviewing logic)

- Temporary `chore(debug)` instrument / remove pairs
- `[autofix.ci]` formatting commits
- Railway migrate lock / migration 0017 idempotency (boot reliability while iterating server-dev)
- Intermediate Steam SDK path reverts / entitlement debugging trail (final behavior: resolve SDK from exe + library-validation entitlement)

---

## Test plan

- [x] Unit: Steam web API, desktop-sign-in, enrollment authorize, enroll/link domain, Steam client, auth race (`wait for silent → open enroll`)
- [ ] Steam `internal-test`: cold launch, unlinked → onboarding Sign in opens `/enroll`
- [ ] Complete enroll (email create/verify) → Electron receives tokens; no `STEAM_ENROLLMENT_TOKEN_INVALID` page
- [ ] Relinked / already-linked: silent login on next Steam launch
- [ ] Direct (non-Steam) build: Sign in still uses plain OIDC when Steam unavailable
- [ ] Confirm Steam build does not prompt GitHub auto-update


lulu0119:
When logging into Steam for the first time, the system currently uses a placeholder email address in the format `steamId@steam.local` and sets `emailVerified: true`. This creates several practical issues: a fake email is displayed on the settings page, email-based workflows such as password reset and account deletion become unavailable, and seamless account merging with OIDC identities like Google is broken.

I plan to add an explicit step during the initial Steam login flow: users will be required to verify a valid real email address or link an existing AIRI account (instead of silently creating an account and finishing the process immediately).

apps/stage-tamagotchi/src/main/services/airi/auth.ts
```diff
@@ -69,6 +89,161 @@ export function createWindowAuthManagerService(): WindowAuthManager {
...
+async function startEnrollmentFlow(
+  windowAuthManager: WindowAuthManager,
+  params: { enrollToken: string, authUiUrl: string },
+): Promise<void> {
+  const authUiBase = params.authUiUrl.replace(/\/+$/, '')
+  await startOidcLoopbackFlow(windowAuthManager, {
+    promptLogin: false,
+    buildBrowserUrl: (authorizeUrl) => {
+      const enrollUrl = new URL(`${authUiBase}/enroll`)
+      enrollUrl.searchParams.set('token', params.enrollToken)
```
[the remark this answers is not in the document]
lulu0119: `ui-server-auth` keeps PostHog; we will not skip init just because the URL may contain a token (including password reset). The Steam-side "skip init on sensitive enrollment URLs" mitigation can be reverted as well.

apps/server/drizzle/0017_nappy_dagger.sql
luoling8192, line 1: Do not modify existing migration files, you should regenerate them after modifying the schema, as this will cause migration failures.

apps/server/src/libs/db.ts
luoling8192, line 1: Why do we need this?

apps/server/src/libs/steam-oidc-tokens.ts
luoling8192, line 1: I think it is better to write a better-auth plugin, such as https://gist.github.com/Whats-A-MattR/5bce5574e568e8d8e6be55cf692df3a1 or https://github.com/Luc4sguilherme/better-auth-steam-demo/blob/main/plugins/better-auth-steam/index.ts

apps/server/src/utils/error.ts
```diff
@@ -29,15 +29,15 @@ export function createBadRequestError(message: string, errorCode = 'BAD_REQUEST'
...
+  return new ApiError(401, errorCode, message, details)
 }
 
 /**
  * Creates a forbidden error (403)
  */
-export function createForbiddenError(message = 'Forbidden', details?: unknown) {
-  return new ApiError(403, 'FORBIDDEN', message, details)
+export function createForbiddenError(message = 'Forbidden', errorCode = 'FORBIDDEN', details?: unknown) {
+  return new ApiError(403, errorCode, message, details)
```
luoling8192, line 40: You should create a new error handler function instead of rewriting the existing implementation

apps/stage-tamagotchi/src/main/services/steam/client.ts
luoling8192, line 1: Why do we not use https://github.com/dustinlacewell/steamworks-ts

luoling8192 (changes requested):

apps/stage-tamagotchi/src/main/services/steam/client.ts
lulu0119: We’re already on a third-party binding — `steamworks-ffi-node` — not a hand-rolled Steam FFI. `client.ts` is just a thin Electron wrapper (lazy load for non-Steam builds, macOS SDK path, init / GetAuthTicketForWebApi / CancelAuthTicket / shutdown).

We evaluated `steamworks-ffi-node` vs `steamworks.js` / greenworks (and a custom addon) during design and locked `steamworks-ffi-node` for SDK 1.64 + Web API ticket support without tying us to Electron NAPI prebuilds.

`steamworks-ts` wasn’t in that shortlist (very low adoption), and after a quick check it also doesn’t expose `GetAuthTicketForWebApi`, so it wouldn’t cover silent sign-in or shrink this file. Happy to revisit if that API lands and Electron packaging looks solid.

PR #1968, opened 2026-06-13
Local main

Suzumiya-SOS:


0xSelenicDove:
Thanks for the contribution. The PR description is still the empty template, which makes this difficult to triage and review.

Could you please add a concise description that covers:

- the user-facing goal (local GPT-SoVITS TTS / voice-cloning support for Stage Tamagotchi);
- the supported GPT-SoVITS server/API versions and expected setup;
- the added configuration fields and how the reference audio is used;
- why the Electron-wide localhost CORS/preflight interception is necessary, its intended scope, and any security considerations;
- validation performed, plus any known limitations.

Please also resolve the current merge conflicts and rebase before requesting final review.

PR #1971, opened 2026-06-13
feat(mcp): progressive tool disclosure — awareness catalog + on-use native promotion

FlowerWater1019:
## Summary

MCP tools are exposed only through two meta-tools (`builtIn_mcpListTools` + `builtIn_mcpCallTool`), which has two costs:

1. **Zero upfront awareness** — the model must call `list` before it knows any MCP tool exists, and it often doesn't, so configured MCP capabilities go unused.
2. **Awkward invocation** — a `server::tool` reference is a string, not a callable function (the model is tempted to emit it as a tool name and the call fails), and tool inputs are double-JSON-encoded.

This PR adds a Skills-style **progressive disclosure** layer: the cheap part (name + one-line description) is always in context for awareness; the expensive part (the full input schema) is loaded only when a tool is actually used.

## Changes

- **Awareness catalog** (`packages/stage-ui/src/tools/mcp.ts`): a generated `name — one-liner` list of the cold (not-yet-promoted) MCP tools, built from the MCP descriptors (schema dropped, description normalized to one line) and injected into the system prompt via the existing toolset-prompt channel — awareness of every capability without loading its schema.
- **On-use native promotion**: a tool the model successfully runs (once, via `builtIn_mcpCallTool`) is reported back and promoted to a native first-class tool — registered with its own JSON Schema under a sanitized `mcp__<server>__<tool>` name — so the model calls it directly from the next turn (no `::`, no double-encoding). The promoted set is persisted in `localStorage` (survives restart); a watch re-derives the live toolset on any change.
- **Settings UI** (`apps/stage-tamagotchi/.../settings/modules/mcp.vue`): lists the activated tools with per-tool and bulk **deactivation** (demotes one back to the catalog).
- The two meta-tools remain as the "cold path" (discover + run); their descriptions now steer the model to a directly-listed `mcp__` tool when one exists.

Context cost scales with **usage**, not with the number of installed tools. `createMcpToolset()` returns `{ tools, catalog }`; the pure helpers (`sanitizeMcpToolName`, `buildMcpNativeNames`, `normalizeOneLine`, `renderMcpCatalog`) are unit-tested.

## Testing

- `pnpm -F @proj-airi/stage-ui exec vitest run src/tools/mcp.test.ts` — 12 passed
- `pnpm -F @proj-airi/stage-tamagotchi exec vitest run src/renderer/stores/mcp-tools.test.ts` — 1 passed
- `pnpm -F @proj-airi/stage-ui typecheck` and `pnpm -F @proj-airi/stage-tamagotchi typecheck` — clean
- `pnpm lint` — clean

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/pages/settings/modules/mcp.vue
+++ b/apps/stage-tamagotchi/src/renderer/pages/settings/modules/mcp.vue
@@ -267,6 +267,9 @@ async function restartServers() {
   try {
     const result = await invokeApplyAndRestart()
     await refreshRuntime()
+    // Re-derive the live MCP toolset so the awareness catalog and native tools reflect the new
+    // server set; otherwise added/removed server tools stay stale until an app reload.
+    await mcpToolsStore.refresh()
     infoMessage.value = tn('messages.restarted', {
       started: result.started.length,
       failed: result.failed.length,
--- a/apps/stage-tamagotchi/src/renderer/pages/settings/modules/mcp.vue
+++ b/apps/stage-tamagotchi/src/renderer/pages/settings/modules/mcp.vue
@@ -14,6 +14,7 @@ import {
   Checkbox,
   TransitionVertical,
 } from '@proj-airi/ui'
+import { storeToRefs } from 'pinia'
 import { computed, onMounted, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 
@@ -30,6 +31,7 @@ import {
   electronMcpWriteConfigText,
 } from '../../../../shared/eventa'
 import { parseElectronMcpConfigText } from '../../../../shared/mcp-config'
+import { useTamagotchiMcpToolsStore } from '../../../stores/mcp-tools'
 import {
   buildConfigFile,
   buildServerConfig,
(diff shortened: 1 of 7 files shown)


FlowerWater1019:
Thanks for the review. All three addressed in the latest push:

- **P2 — handle bad JSON inside MCP call errors** (`mcp.ts`): `builtIn_mcpCallTool` now wraps `JSON.parse` in a try/catch and returns a structured `isError` result for malformed `arguments`, instead of throwing and aborting the turn. Added regression tests (malformed JSON → `isError`, never reaches the runtime; successful call → promotion).
- **P2 — register MCP tools before awaiting discovery** (`mcp-tools.ts`): `refresh()` now registers the meta-tools with the live runtime *before* awaiting descriptor discovery, so a chat sent during the startup window no longer sees only the unavailable fallback.
- **P2 — refresh after server restarts** (`settings/modules/mcp.vue`): restarting MCP servers now calls `mcpToolsStore.refresh()`, so added/removed server tools no longer stay stale in the awareness catalog until an app reload.

typecheck + lint clean; mcp tests pass.

FlowerWater1019:
Closing this one.

PR #1972, opened 2026-06-13
feat(stage-ui): configurable chat response length limits (max tokens + reply-length hint)

FlowerWater1019:
## Summary

Two opt-in controls in the consciousness module settings to keep replies snappy (inspired by how galgame-style companions keep replies short):

- **Max output tokens** — a hard cap forwarded through `StreamOptions.maxTokens`, serialized by xsAI as the OpenAI-compatible `max_tokens` body field. Omitted entirely when disabled.
- **Reply length guideline** — a soft character-budget instruction appended to the system-prompt supplement.

Both default to `0` (off), so existing behavior is unchanged.

## Changes

- `core-agent`: `StreamOptions.maxTokens` plumbed into the `streamText` request (lands as `max_tokens`).
- `stage-ui`: a `response-length` prompt helper (the soft guideline), consciousness-module settings, and chat-store wiring.
- `stage-pages`: the two controls on the consciousness settings page.
- i18n (en / zh-Hans / zh-Hant).

## Testing

- `pnpm -F @proj-airi/stage-ui exec vitest run src/constants/prompts/response-length.test.ts` — 3 passed
- `pnpm -F @proj-airi/core-agent exec vitest run src/runtime/llm-service.test.ts` — 12 passed
- typecheck (feature files) + lint — clean

### DIFF
--- a/packages/core-agent/src/runtime/llm-service.test.ts
+++ b/packages/core-agent/src/runtime/llm-service.test.ts
@@ -112,6 +112,56 @@ describe('streamFrom tool error capture', () => {
   })
 })
 
+describe('streamFrom maxTokens forwarding', () => {
+  function mockImmediateFinish() {
+    streamTextMock.mockImplementationOnce((options: { onEvent: (event: unknown) => Promise<void> }) => {
+      const steps = Promise.resolve([]).then(async (value) => {
+        await options.onEvent({ type: 'finish', finishReason: 'stop' })
+        return value
+      })
+      return createMockStreamResult(steps)
+    })
+  }
+
+  /**
+   * @example
+   * await streamFrom({ model, chatProvider, messages, options: { maxTokens: 1024 } })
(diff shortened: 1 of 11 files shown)


FlowerWater1019:
Thanks for the review.

- **P2 — use the completion-token field for reasoning models** (`llm-service.ts`): fixed. `streamFrom` now detects OpenAI reasoning families (o1/o3/o4…, gpt-5, matched on the final path segment so provider-prefixed ids like `openai/gpt-5-mini` resolve) and routes the response-length cap to `maxCompletionTokens` for them, keeping `maxTokens` for everything else (non-reasoning OpenAI + non-OpenAI providers that only understand `max_tokens`). Added regression tests asserting the field selection for both a `gpt-5` and an `o3` model.

typecheck + lint clean; llm-service tests pass.

FlowerWater1019:
Closing this one.

PR #1973, opened 2026-06-13
feat(memory): opt-in long-term memory (local IndexedDB store + recall)

FlowerWater1019:
## Summary

Opt-in **long-term memory** (LTM): the companion remembers facts across sessions. Conversations are distilled into memory items, embedded, and stored locally; relevant memories are recalled into the chat context on later turns.

Fully local (IndexedDB + in-browser embeddings + cosine-similarity search) and gated behind a settings toggle — off by default, no behavior change when disabled.

## Changes

- **Store** (`database/repos/memory.repo.ts`, `stores/modules/memory/store.ts`): IndexedDB-backed memory store with cosine-similarity search.
- **Embedder** (`composables/memory/use-memory-embedder.ts`): embeds memory text for retrieval.
- **Recall** (`composables/memory/use-memory-recall.ts`, `stores/chat/context-providers/memory.ts`, `stores/modules/memory/recall-scoring.ts`): scores and injects relevant memories into the chat context.
- **Write path** (`composables/memory/use-memory-write.ts`, `stores/modules/memory/extraction.ts`): extracts memory items from conversation.
- **Browser + settings** (`composables/memory/use-memory-browser.ts`, `stage-pages/.../memory-long-term.vue`): a settings page to view/manage stored memories.
- Design doc: `docs/design/long-term-memory.md`.

## Testing

- `pnpm -F @proj-airi/stage-ui exec vitest run src/composables/memory src/database/repos/memory.repo.test.ts src/stores/modules/memory` — 39 passed
- `pnpm -F @proj-airi/core-agent exec vitest run src/messages/context-prompt.test.ts` — 7 passed
- typecheck (feature files) + lint — clean


FlowerWater1019:
Thanks for the review. Addressed in the latest push:

- **P2 — embedding dimension setting** (`memory-long-term.vue`): the settings page now binds `embeddingDimension` (numeric input) with a note that a mismatch silently fails recall/write. The store field already existed; it just wasn't exposed.
- **Recall hook gating** (`chat.ts`): the before-compose recall hook is now skipped entirely when memory is off (mirroring the write hook), so it no longer ingests a blank clearing context every turn in the common memory-off case.
- **CI (Unit Test)**: `chat.contract.test.ts` mocks the new memory composables/store (disabled); `exports.contract.test.ts` adds the `./stores/modules/memory` subpath. This fixes the 10 contract failures.

**P2 — scope memory hooks to the target session**: valid, but left as a follow-up. The recall/write hooks read `activeSessionId` because the core-agent hook callbacks (`onBeforeMessageComposed`/`onChatTurnComplete`) only receive `(message, context)` with no target session id — threading it through would change the hook contract across `@proj-airi/core-agent` and every consumer. Since it only affects sends to a *non-active* session (chat-sync retries / `ingestOnFork`), I'd rather not expand this PR into the hook contract; happy to do it as a dedicated follow-up.

FlowerWater1019:
Closing this one.

PR #1975, opened 2026-06-13
feat(vision): headless background screen vision as a silent chat-context signal

FlowerWater1019:
## Summary

Headless background screen vision for the desktop pet. When enabled, a resident driver captures the primary screen on an interval, runs it through the configured vision (VLM) provider, and feeds the resulting description into the chat backbone as an **auxiliary, silent** context signal — the persona understands what you're doing without narrating your screen.

## Changes

- **Resident background capture** (`BackgroundVisionCapture.vue`) — an always-mounted, off-screen driver that runs the vision ticker outside the devtools Vision page, gated purely by a persisted `settings/vision/background-capture-enabled` toggle. Previously capture only ran while the devtools Vision page was open.
- **Cross-window capture-leader election** — every Electron window mounts the driver, so a localStorage lease elects exactly one capture leader; the others stand down. Prevents N windows from hammering the (serialized) VLM queue at once.
- **Local context injection** (`chat/context-providers/vision.ts`) — the renderer both produces and consumes this context, and `context:update` isn't echoed back to the sender, so the latest screen description is read directly from the orchestrator at compose time, with a freshness gate (stale frames are dropped).
- **Screen-awareness prompt** (`VISION_AWARENESS_PROMPT`) — tells the backbone that `vision:`/`screen:` context entries are background perception to stay quiet about unless directly asked, injected into the system prompt only while background capture is on.
- **Cross-window status** — capture counters / last-capture timestamps and the latest result move to localStorage so the settings UI (usually a different window than the capture leader) reflects real activity.

## Notes

- Provider/model are runtime-configured through the existing vision settings — no hardcoded endpoints.
- Feature is opt-in and defaults to off.
- UI / Electron + multi-window coordination feature — exercised manually; no unit tests added.

## Testing

- typecheck + lint — clean on `@proj-airi/stage-ui` and `@proj-airi/stage-tamagotchi`.


FlowerWater1019:
Thanks for the review. Addressed in the latest push:

- **P1 — vision context behind the opt-in** (`context-providers/vision.ts`): `createVisionContext` now returns `null` unless `backgroundCaptureEnabled`, matching the `VISION_AWARENESS_PROMPT` gate in `chat.ts`. A devtools (publish-off) capture or a frame left over after the toggle is turned off can no longer leak the persisted `lastResultText` into the next chat turn.
- **P1 — mock `generateText` in the inference tests**: the implementation moved from `llmStore.stream` to `@xsai/generate-text`; `use-vision-inference.test.ts` now mocks `generateText` (and dropped the dead `useLLM` mock), so it no longer hits a real provider. This was also the Unit Test CI failure.
- **CI (Unit Test)**: `chat.contract.test.ts` now mocks the new `createVisionContext` export from `./chat/context-providers`, fixing the 9 contract failures caused by the unmocked import.

Green locally: typecheck + lint clean, contract + inference tests pass.

FlowerWater1019:
Closing this one.

PR #1978, opened 2026-06-14
feat(stage-tamagotchi): reset to center

Neko-233:
## Summary

- add an Electron invoke to move the main AIRI desktop window back to the display center
- expose the recovery action from the controls island and Data settings page
- add localized labels/status copy and regression coverage for multi-display and oversized-window bounds
- clean up shared event-source metadata formatting and add router virtual-module shims needed for typecheck

## Validation

- pnpm typecheck
- pnpm lint
- pnpm -F @proj-airi/stage-tamagotchi exec vitest run src/main/windows/main/window-position.test.ts
- pnpm -F @proj-airi/core-agent exec vitest run src/runtime/context-registry.test.ts src/agents/spark-notify/handler.test.ts
- git diff --check

### DIFF
--- a/apps/stage-tamagotchi/src/main/windows/settings/rpc/index.electron.ts
+++ b/apps/stage-tamagotchi/src/main/windows/settings/rpc/index.electron.ts
@@ -16,6 +16,7 @@ import { createContext } from '@moeru/eventa/adapters/electron/main'
 import { ipcMain } from 'electron'
 
 import {
+  electronCenterMainWindow,
   electronOpenDevtoolsWindow,
   electronOpenSettingsDevtools,
   electronSpotlightShortcutGet,
@@ -26,6 +27,7 @@ import { createGodotStageService } from '../../../services/airi/godot-stage'
 import { createMcpServersService } from '../../../services/airi/mcp-servers'
 import { createWidgetsService } from '../../../services/airi/widgets'
 import { createAutoUpdaterService } from '../../../services/electron'
+import { centerWindowOnDisplay } from '../../shared/display'
 import { setupBaseWindowElectronInvokes } from '../../shared/window'
 
 export async function setupSettingsWindowInvokes(params: {
@@ -33,6 +35,7 @@ export async function setupSettingsWindowInvokes(params: {
   widgetsManager: WidgetsWindowManager
   autoUpdater: AutoUpdater
   devtoolsWindow: DevtoolsWindowManager
+  getMainWindow?: () => BrowserWindow | undefined
   serverChannel: ServerChannel
   godotStageManager: GodotStageManager
   mcpStdioManager: McpStdioManager
@@ -59,6 +62,7 @@ export async function setupSettingsWindowInvokes(params: {
   // Register the global shortcut service for the settings window.
   params.globalShortcut.registerWindow({ context, window: params.settingsWindow })
 
+  defineInvokeHandler(context, electronCenterMainWindow, () => centerWindowOnDisplay(params.getMainWindow?.()))
   defineInvokeHandler(context, electronSpotlightShortcutGet, () => params.spotlightWindow.getShortcutAccelerator())
   defineInvokeHandler(context, electronSpotlightShortcutSet, (payload) => {
     if (payload?.accelerator === undefined)
--- a/apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
+++ b/apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
@@ -135,9 +135,7 @@ function refreshWindow() {
  * Requests the main process to move the AIRI desktop window back to screen center.
  */
 function resetMainWindowPosition() {
-  void centerMainWindow().catch((error) => {
-    console.error(error)
-  })
+  centerMainWindow().catch(console.error)
 }
 </script>
 
--- a/apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
+++ b/apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
@@ -19,6 +19,7 @@ import IndicatorMicVolume from './indicator-mic-volume.vue'
 import {
   electron,
   electronAppQuit,
+  electronCenterMainWindow,
   electronOpenChat,
   electronOpenSettings,
   electronStartDraggingWindow,
@@ -38,6 +39,7 @@ const openChat = useElectronEventaInvoke(electronOpenChat)
 const isLinux = useElectronEventaInvoke(electron.app.isLinux)
 const closeWindow = useElectronEventaInvoke(electronAppQuit)
 const setAlwaysOnTop = useElectronEventaInvoke(electronWindowSetAlwaysOnTop)
+const centerMainWindow = useElectronEventaInvoke(electronCenterMainWindow)
 
 const expanded = ref(false)
 const islandRef = ref<HTMLElement>()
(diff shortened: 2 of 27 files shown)


apps/stage-pocket/src/auto-routes.d.ts
```diff
@@ -0,0 +1,5 @@
+import type { RouteRecordRaw } from 'vue-router'
+
+declare module 'vue-router/auto-routes' {
+  export const routes: RouteRecordRaw[]
+}
```
nekomeowww: Why this is needed? It should generate by itself?

apps/stage-tamagotchi/src/main/windows/settings/rpc/index.electron.ts
```diff
@@ -59,6 +62,21 @@ export async function setupSettingsWindowInvokes(params: {
...
+  function centerMainWindow() {
+    const mainWindow = params.getMainWindow?.()
+    if (!mainWindow || mainWindow.isDestroyed())
+      throw new Error('Main AIRI window is not available.')
+
+    return centerWindowOnDisplay({
+      getDisplayMatching: bounds => screen.getDisplayMatching(bounds),
+      window: mainWindow,
+    })
+  }
```
nekomeowww: Why proxying the function from centerWindowOnDisplay here with another layer?
If you need to ensure the window exists, shouldn't you use utils to achieve this?

apps/stage-tamagotchi/src/renderer/pages/settings/data/components/desktop-reset-section.vue
```diff
@@ -20,6 +27,19 @@ async function resetDesktopState() {
...
+ */
+async function resetDesktopWindowPosition() {
+  try {
+    await centerMainWindow()
+    emitStatus(t('settings.pages.data.status.desktop_window_centered'))
+  }
+  catch (error) {
+    handleActionError(error)
+  }
+}
```
nekomeowww, line 45: Another wrapper again?

apps/stage-web/src/auto-routes.d.ts
```diff
@@ -0,0 +1,5 @@
+import type { RouteRecordRaw } from 'vue-router'
+
+declare module 'vue-router/auto-routes' {
+  export const routes: RouteRecordRaw[]
+}
```
nekomeowww: Why writing those?

apps/stage-tamagotchi/src/main/windows/main/window-position.ts
nekomeowww, line 1: Would you mind checking how https://github.com/moeru-ai/airi/blob/main/apps/stage-tamagotchi/src/main/windows/shared/display.ts organize and write code for window size & display management?

packages/i18n/src/locales/zh-Hans/settings.yaml
```diff
@@ -428,6 +428,7 @@ pages:
       desktop:
         title: 重置桌面设置和状态
         description: 清除 AIRI 桌面设置和运行状态。
+        center: 将桌宠移回屏幕中心
```
nekomeowww: 桌宠？

Can we not say it's "桌宠" but instead call it AIRI.

packages/i18n/src/locales/en/settings.yaml
```diff
@@ -447,6 +447,7 @@ pages:
       desktop:
         title: Reset desktop settings & states
         description: Clear AIRI desktop settings and runtime state.
+        center: Move desktop window to center
```
nekomeowww, line 459: As it seems English don't have such issue.

apps/stage-pocket/src/auto-routes.d.ts
Neko-233: Removed both generated route declaration files from this PR. They were unrelated typecheck workarounds and are not part of desktop window recovery.

apps/stage-tamagotchi/src/main/windows/settings/rpc/index.electron.ts
Neko-233: Moved the window availability check into the shared display utility and reduced this to a direct one-line handler registration. The main and settings registrations still both exist because Eventa contexts are scoped to the sender window; the controls island and settings page invoke through different contexts.

apps/stage-tamagotchi/src/renderer/pages/settings/data/components/desktop-reset-section.vue
Neko-233: Renamed this to handleCenterMainWindow to make its UI-handler role explicit. It intentionally owns the async success status and shared error reporting; binding the invoke function directly would drop that user feedback.

apps/stage-web/src/auto-routes.d.ts
Neko-233: Removed both generated route declaration files from this PR. They were unrelated typecheck workarounds and are not part of desktop window recovery.

apps/stage-tamagotchi/src/main/windows/main/window-position.ts
Neko-233: Moved the centering calculation, Electron window operation, and regression tests into the existing windows/shared/display.ts module and its test file.

packages/i18n/src/locales/zh-Hans/settings.yaml
Neko-233: Updated the Simplified Chinese copy to call it the AIRI window instead of 桌宠.

packages/i18n/src/locales/en/settings.yaml
Neko-233: Agreed. The existing English wording did not have that naming issue, so it remains unchanged apart from the new feature labels.

apps/stage-tamagotchi/src/shared/eventa/index.ts
```diff
@@ -33,6 +33,20 @@ export const electronStartTrackMousePosition = defineInvokeEventa('eventa:invoke
...
+export interface ElectronMainWindowBounds {
+  /** Logical screen x-coordinate of the window origin. */
+  x: number
+  /** Logical screen y-coordinate of the window origin. */
+  y: number
+  /** Current window width in logical pixels. */
+  width: number
+  /** Current window height in logical pixels. */
+  height: number
+}
```
nekomeowww: Can we use `Rectangle` from type of `electron`?

apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
```diff
@@ -128,6 +130,15 @@ const startDraggingWindow = !isLinux() ? defineInvoke(context.value, electronSta
...
   window.location.reload()
 }
+
+/**
+ * Requests the main process to move the AIRI desktop window back to screen center.
+ */
+function resetMainWindowPosition() {
+  void centerMainWindow().catch((error) => {
+    console.error(error)
+  })
```
nekomeowww:
```suggestion
  centerMainWindow().catch(console.error)
```

Neko-233: Updated. The click handler now calls `centerMainWindow().catch(console.error)` directly.

apps/stage-tamagotchi/src/shared/eventa/index.ts
Neko-233: Updated. `electronCenterMainWindow` now uses Electron's `Rectangle` directly through a type-only import, removing the duplicate local bounds interface.

PR #1998, opened 2026-06-20
feat(stage-ui): add character card import and export

BeanDz:


packages/stage-ui/src/services/airi-card-import-export.ts
```diff
@@ -0,0 +1,277 @@
...
+}
+
+async function loadZip(file: File) {
+  try {
+    return await JSZip.loadAsync(await file.arrayBuffer())
+  }
+  catch {
+    throw error('invalid-file', 'Invalid zip file')
+  }
+}
```
LemonNekoGH, line 155: It just re-throw the error and we will never know the details of origin error.

packages/stage-ui/src/services/airi-card-import-export.ts
```diff
@@ -0,0 +1,277 @@
...
+
+async function readJsonFile<T>(zip: JSZip, path: string, guard: (value: unknown) => value is T): Promise<T> {
+  const file = zip.file(path)
+  if (!file)
+    throw error('missing-file', `Missing ${path}`)
+
+  const value = parseJson(await file.async('string'))
+  if (!guard(value))
+    throw error('invalid-file', `Invalid ${path}`)
```
LemonNekoGH: The same above

packages/stage-ui/src/services/airi-card-import-export.ts
```diff
@@ -0,0 +1,277 @@
...
+  const source = record(value)
+  return { provider: text(source.provider), model: text(source.model) }
+}
+
+function parseJson(content: string) {
+  try {
+    return JSON.parse(content) as unknown
+  }
+  catch {
+    throw error('invalid-file', 'Invalid JSON')
```
LemonNekoGH: The same above

packages/stage-ui/src/services/airi-card-import-export.ts
```diff
@@ -0,0 +1,277 @@
...
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
+
+function isSupportedDisplayModelFormat(value: unknown): value is SupportedDisplayModelFormat {
+  return value === DisplayModelFormat.Live2dZip || value === DisplayModelFormat.SpineZip || value === DisplayModelFormat.VRM
+}
+
+function isSpawnMode(value: unknown): value is NonNullable<AiriExtension['modules']['artistry']>['spawnMode'] {
```
LemonNekoGH, line 247: Why so many helpers, are those necessary, are those already exists?

LemonNekoGH (changes requested):
We should also export model file even it was built-in in AIRI

packages/stage-ui/src/services/airi-card-import-export.ts
LemonNekoGH: You can check this [docuemnt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) from mdn about `cause`, also, you can send this link to your AI agent.

LemonNekoGH (changes requested):

packages/stage-ui/src/services/airi-card-import-export.ts
LemonNekoGH: You can use console in devtools to try any JavaScript code like this:

Image

In this example, I tried to re-throw an error, you can see that we need to pass `cause` as object.

LemonNekoGH (changes requested):

PR #1988, opened 2026-06-17
Improve desktop voice input transcription reliability

Neko-233:
## Summary

- Fix Windows packaged voice input by tightening Electron media permission handling and recovering stale or missing microphone streams before listening starts.
- Move recorder-backed voice input into a shared `useVoiceInputSession` pipeline used by the main stage and the Settings speech-to-text test, so both paths share recording, transcription, stale-result gating, and error handling.
- Add a volume-based recorder fallback alongside VAD for record-then-transcribe providers. This lets the main stage still segment speech when VAD initializes but does not emit speech events on some Windows devices.
- Keep assistant-playback suppression and stale segment guards around both streaming ASR setup and recorder-backed ASR results, so microphone audio and old transcriptions do not leak into active assistant playback.
- Add a temporary main-stage voice input log panel for diagnosing field reports; it records stream, VAD/fallback, recorder, ASR, and chat-ingest boundaries.

## Root Cause

The Settings test only recorded a manual audio segment and called the transcription provider directly. The main stage had an extra continuous-listening layer: microphone stream startup, VAD speech segmentation, recorder segment creation, transcription, then chat ingestion. On Windows, the microphone stream and provider could be healthy while VAD never emitted the expected speech-start/speech-end events, so the main stage never produced recorder-backed ASR requests.

The fix keeps the continuous-listening behavior but adds a shared session layer and volume fallback, while preserving the shorter manual Settings test as the same session in manual segment mode.

## Test Plan

- `vitest run packages/stage-ui/src/composables/audio/voice-input-segment.test.ts packages/stage-ui/src/composables/audio/voice-input-vad-startup.test.ts --config packages/stage-ui/vitest.config.ts`
- `eslint apps/stage-tamagotchi/src/renderer/pages/index.vue apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue packages/stage-pages/src/pages/settings/modules/hearing.vue packages/stage-ui/src/composables/audio/index.ts packages/stage-ui/src/composables/audio/voice-input-session.ts packages/stage-ui/src/composables/audio/voice-input-segment.ts packages/stage-ui/src/composables/audio/voice-input-segment.test.ts packages/stage-ui/src/composables/audio/voice-input-vad-startup.ts packages/stage-ui/src/composables/audio/voice-input-vad-startup.test.ts`
- `pnpm -rF @proj-airi/stage-ui run typecheck`
- `pnpm -rF @proj-airi/stage-pages run typecheck`
[...]

PR #2002, opened 2026-06-22
fix(stage-tamagotchi): permission issue

Neko-233:
## Summary

- Register paired Electron permission request and check handlers on the default AIRI session.
- Grant audio-only media access only when every requesting, security, and embedding identity belongs to a packaged file page or loopback development page.
- Preserve reviewed local display-capture and sanitized clipboard-write behavior while denying unreviewed permission categories.
- Wire the existing systemPreferences Eventa service so renderer windows can query and request macOS-level microphone consent.

## Split from

Split from #1988. This PR now contains only Electron media permission work.

The packaged auth callback follow-up moved to #2042.

## Test plan

- pnpm exec vitest run apps/stage-tamagotchi/src/main/services/electron/media-permissions.test.ts
- pnpm typecheck
- pnpm lint


apps/stage-tamagotchi/src/main/services/electron/media-permissions.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+  catch {
+    console.warn('[Media Permissions] Ignoring invalid requester URL:', rawURL)
+    return false
+  }
+}
+
+function isUsableRequesterURL(rawURL: string | undefined) {
+  return !!rawURL && rawURL !== 'null'
+}
```
nekomeowww: I don't quite understand why we need local app url checking and parse & validating this by ourselves. Would you mind clarify this?

apps/stage-tamagotchi/src/main/services/airi/http-server/http/auth/index.ts
```diff
@@ -7,6 +9,8 @@ export interface LoopbackCallbackResult {
   state: string
 }
 
+const AIRI_AUTH_ORIGIN = 'https://accounts.airi.build'
```
nekomeowww: Why this is needed? Why local http server has anything to do with auth domain?

apps/stage-tamagotchi/src/main/services/airi/http-server/http/auth/index.ts
```diff
@@ -39,13 +43,20 @@ export async function startLoopbackServer(): Promise<{
 
   const app = new H3()
   const loopbackServer = createH3Server({ app, host })
-  const corsOptions = {
-    origin: '*',
+  const corsOptions: CorsOptions = {
+    origin: [AIRI_AUTH_ORIGIN],
```
nekomeowww: Why is this no-longer `*`?

apps/stage-tamagotchi/src/main/services/airi/http-server/http/auth/index.ts
```diff
@@ -39,13 +43,20 @@ export async function startLoopbackServer(): Promise<{
...
     },
-  } as const
+  }
+
+  function appendPrivateNetworkAccessHeaders<T extends { headers: Headers }>(event: Parameters<typeof handleCors>[0], response: T) {
+    if (getRequestHeader(event, 'origin') === AIRI_AUTH_ORIGIN)
+      response.headers.set('Access-Control-Allow-Private-Network', 'true')
+
+    return response
+  }
```
nekomeowww: Why this is needed?

nekomeowww: I see that in the server this header wasn't even returned, how come this part of code would work?

apps/stage-tamagotchi/src/main/services/electron/media-permissions.ts
```diff
@@ -0,0 +1,102 @@
...
+    return false
+
+  if (details?.mediaTypes?.length)
+    return details.mediaTypes.includes('audio') && !details.mediaTypes.includes('video')
+
+  if (permission !== 'media')
+    return true
+
+  return details?.mediaType === 'audio'
+}
```
nekomeowww: I don't get this. Should we check how electron official docs says about the permission checks?

- https://www.electronjs.org/docs/latest/api/session
- https://www.electronjs.org/docs/latest/api/system-preferences

And why don't use https://www.electronjs.org/docs/latest/api/system-preferences? Would you mind checking how https://github.com/moeru-ai/airi/tree/main/packages/electron-screen-capture works?

apps/stage-tamagotchi/src/main/services/electron/media-permissions.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+export function setupMediaPermissionHandlers(targetSession: Session) {
+  targetSession.setPermissionRequestHandler((webContents, permission, callback, details) => {
+    callback(shouldGrantElectronPermission(webContents, permission, undefined, details))
+  })
+
+  targetSession.setPermissionCheckHandler((webContents, permission, requestingOrigin, details) => {
+    return shouldGrantElectronPermission(webContents, permission, requestingOrigin, details)
+  })
+}
```
nekomeowww: What is the purpose of those?

apps/stage-tamagotchi/src/main/services/electron/media-permissions.ts
Neko-233: The URL check protects a global Electron session handler: it distinguishes AIRI-owned packaged file pages and loopback dev pages from remote frames loaded in the same session. The policy parses URLs so only exact file/localhost/loopback identities pass, and it now also rejects a local child frame when Electron reports a remote embeddingOrigin.

apps/stage-tamagotchi/src/main/services/electron/media-permissions.ts
Neko-233: Updated this to derive its permission and detail types directly from Electron's official setPermissionCheckHandler/setPermissionRequestHandler contracts, and both handlers are installed as the session docs require. systemPreferences is still used through the existing Eventa service for macOS OS-level consent/status; the session handlers cover Chromium renderer authorization. Existing local display capture remains allowlisted and OS-level screen permission continues to use packages/electron-screen-capture.

apps/stage-tamagotchi/src/main/services/electron/media-permissions.ts
Neko-233: These install AIRI's allowlist on Electron's default session. Audio-only media is granted only to packaged file or loopback frames, while the existing local display-capture and sanitized clipboard-write paths are preserved; every unreviewed permission category remains denied.

apps/stage-tamagotchi/src/main/services/airi/http-server/http/auth/index.ts
```diff
@@ -47,6 +47,11 @@ export async function startLoopbackServer(): Promise<{
     },
   } as const
 
+  function appendPrivateNetworkAccessHeaders<T extends { headers: Headers }>(response: T) {
+    response.headers.set('Access-Control-Allow-Private-Network', 'true')
```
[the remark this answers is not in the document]
Neko-233: Moved the packaged callback work to #2042. The loopback server now receives the generated OIDC state and rejects a mismatch before settling or closing, so a forged request cannot consume the one-shot listener. The obsolete PNA response header is not added.

apps/stage-tamagotchi/src/main/services/airi/http-server/http/auth/index.ts
Neko-233: Removed the auth callback changes from this media-permission PR. The focused follow-up is #2042, and it does not hardcode accounts.airi.build into the loopback server.

apps/stage-tamagotchi/src/main/services/airi/http-server/http/auth/index.ts
Neko-233: Restored standard wildcard CORS in this PR and moved auth to #2042. The relay origin can vary by deployment, and CORS is not used as authorization because it does not stop a simple GET; #2042 validates the high-entropy OIDC state inside the loopback server instead.

apps/stage-tamagotchi/src/main/services/airi/http-server/http/auth/index.ts
Neko-233: Removed this header from #2002. Chrome 142 Local Network Access replaced the paused PNA preflight approach, so the focused #2042 follow-up keeps ordinary CORS and relies on the browser LNA permission plus server-side OIDC state validation.

apps/stage-tamagotchi/src/main/services/electron/media-permissions.ts
```diff
@@ -0,0 +1,83 @@
...
+  return isLocalAppURL(webContents?.getURL())
+}
+
+/**
+ * Lets AIRI-owned Electron windows use the microphone on platforms where Chromium asks the
+ * session for media permission instead of relying on macOS systemPreferences APIs.
+ */
+export function setupMediaPermissionHandlers(targetSession: Session) {
+  targetSession.setPermissionRequestHandler((webContents, permission, callback, details) => {
+    callback(shouldGrantAudioCapturePermission(webContents, permission, undefined, details))
```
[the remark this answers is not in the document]
Neko-233: Addressed. `clipboard-sanitized-write` is now preserved for AIRI-owned local pages with the same requester-origin validation, while unknown or remote permissions remain denied. Regression coverage now protects the existing copy behavior.

apps/stage-tamagotchi/src/main/services/electron/media-permissions.ts
```diff
@@ -0,0 +1,154 @@
...
+      return false
+
+    return url.hostname === 'localhost'
+      || url.hostname === '127.0.0.1'
+      || url.hostname === '[::1]'
+  }
+  catch {
+    return false
+  }
+}
```
nekomeowww: Move this to either `shared/` for electron or `libs/` for electron.

Neko-233: Moved. `isLocalAppURL` now lives in `src/main/libs/electron/url.ts` as the shared Electron main-process trust helper, and the media permission policy imports it. The 14 permission policy tests still pass.

apps/stage-tamagotchi/src/main/libs/electron/url.ts
```diff
@@ -0,0 +1,34 @@
...
+    const url = new URL(rawURL)
+    if (url.protocol === 'file:')
+      return true
+
+    if (url.protocol !== 'http:' && url.protocol !== 'https:')
+      return false
+
+    return url.hostname === 'localhost'
+      || url.hostname === '127.0.0.1'
+      || url.hostname === '[::1]'
```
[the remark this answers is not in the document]
Neko-233: Fixed in 3667a64d2. HTTP(S) requester URLs now have to match the exact ELECTRON_RENDERER_URL origin instead of accepting any loopback host and port, so extension static-asset frames and plugin development servers remain denied. I added regression coverage for both cases; the media permission tests pass (16/16), package and root typechecks pass, and lint reports 0 warnings/errors.

nekomeowww (approved):

PR #2026, opened 2026-07-01
feat(pipelines-audio): Refine event types and error handling

Iro96:
## Description

- Tightened `llm-streaming-control` controller typing and removed `any`.
- Added safe turn ID fallback when `crypto.randomUUID()` is unavailable.
- Made `createPlaybackManager` satisfy the pipeline playback contract by adding `stopByIntent` and `stopByOwner`.
- Normalized playback manager options and error handling with `errorMessageFrom`.
- Added `rejectedAt` to playback reject events, matching emitted data.
- Prevented stranded playback waiters if `schedule()` throws.
- Removed debug leakage from `tts-chunker`.
- Released the internal byte stream reader lock in `createTtsSegmentStream`.
- Improved `processNarrative` range removal from repeated range scans to a sorted linear walk.
- Replaced event payload `any` usage with `unknown`.


packages/pipelines-audio/src/llm-streaming-control/payloads.ts
```diff
@@ -24,77 +34,135 @@ export interface NormalizedActPayload {
   motion?: string
 }
 
-function normalizeEmotionName(value: string): StreamingControlEmotion | undefined {
+/** Runtime-safe object detection. */
+function isObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
```
nekomeowww: Can we remove this?

Iro96: of course

nekomeowww (approved):

PR #2069, opened 2026-07-16
fix(stage-tamagotchi): stabilize plugin auto-reload test

0xSelenicDove:
## Summary

- keep polling while the auto-reload transition has no active session
- prevent the test from failing during the intentional unload/load gap

## Root cause

Auto-reload stops the old plugin session before starting the replacement. The test exited its polling loop when an inspect snapshot observed that temporary empty state, then asserted that the missing session had an ID.

## Validation

- targeted auto-reload regression test
- full `apps/stage-tamagotchi/src/main/services/airi/plugins/index.test.ts`
- `pnpm typecheck`
- `pnpm lint` (0 errors; 6 existing warnings)


0xSelenicDove:
Superseded by #2068. The auto-reload test stabilization from this PR is included as the second commit in the VRM interaction PR, so #2068 now contains the complete change set.

PR #2090, opened 2026-07-19
fix(stage-ui): apply display model on character card activation

### LINKED ISSUE #2089: bug: character card activation not applying its display model
## Problem
Activating a character card does not reliably apply its configured VRM/Live2D display model to the stage runtime.

## Cause
The AIRI card store derives activation from `activeCardId`, but the runtime synchronization was attached to a debounced `activeCard` object watcher. This makes model application dependent on object-level reactivity rather than the activation key itself.

## Expected behavior
Switching character cards should immediately apply the selected card's `displayModelId` to the stage model runtime.

## Scope
Add an explicit `activeCardId` synchronization path and regression coverage.

0xSelenicDove:
## Summary
- apply active character-card settings from the stable `activeCardId` activation key
- ensure the selected VRM/Live2D display model reaches the stage runtime when switching cards
- add regression coverage for Issue #2089

Fixes #2089

## Validation
- `pnpm typecheck`
- `pnpm lint` (0 errors; existing warnings remain)
- `pnpm exec eslint packages/stage-ui/src/stores/modules/airi-card.ts packages/stage-ui/src/stores/modules/airi-card.test.ts`
- focused Vitest: 4 tests passed; Node reports the existing `localforage` storage warning in this environment

### DIFF
--- a/packages/stage-ui/src/stores/modules/airi-card.test.ts
+++ b/packages/stage-ui/src/stores/modules/airi-card.test.ts
@@ -1,8 +1,25 @@
+import type { AiriCard } from './airi-card'
+
 import { createPinia, setActivePinia } from 'pinia'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { useSettingsStageModel } from '../settings/stage-model'
 import { useAiriCardStore } from './airi-card'
+
+// NOTICE:
+// Vitest runs these store tests in Node, where localforage cannot select a
+// browser storage driver. The stage-model watcher legitimately asks the
+// display-model store to resolve IDs, so provide the storage boundary with a
+// deterministic no-op instead of allowing rejected driver initialization to
+// escape as an unrelated test error.
+vi.mock('localforage', () => ({
+  default: {
+    getItem: vi.fn(async () => undefined),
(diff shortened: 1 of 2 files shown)


0xSelenicDove:
Addressed the review feedback in commit `2491acc7f`. The synchronization now watches both `activeCardId` and the active card object, so switching cards and saving edits to the currently active card both update the runtime model/provider settings. Added regression coverage for active-card edits; the review thread is resolved.

0xSelenicDove:
Addressed the reset-ordering feedback in commit `2fc177d60`. Card data is now cleared before `activeCardId`, preventing the synchronous watcher from reapplying the old default card model during a full settings reset. Added a regression test. Also mocked the browser-only localforage boundary in the Node unit test; the focused suite now passes 6/6 without unhandled errors.

0xSelenicDove:
Rebuilt this PR from origin/main to keep it scoped to character-card/model synchronization. The inherited chat-import commit and all chat-related files have been removed from this PR. It now contains only airi-card.ts and airi-card.test.ts; the chat-import work remains separate.

luoling8192 (approved):

PR #2097, opened 2026-07-22
fix(discord): bound and fairly schedule retained input work

Woww603:
## Summary

- bound Discord-origin retained input before preprocessing, Web Lock acquisition, and the core orchestration queue
- enforce admission limits of 2 retained items per session, 4 per verified Discord principal, and 8 globally
- schedule admitted work round-robin across principals while preserving FIFO order within each session
- release capacity exactly once across success, provider/preprocessing failure, rejection, timeout, cancellation, disconnect, and shutdown
- reject missing or malformed Discord principal identifiers instead of falling back to display, channel, or session identifiers

## Security impact

The Discord `MessageCreate` path previously forwarded every retained input into an unbounded core FIFO. A stable Discord user ID was present in `event.data.discord.guildMember.id`, but admission ignored it, so one principal could retain work across multiple sessions and monopolize scheduling.

This change adds a bounded admission layer before expensive preprocessing and orchestration. Only non-zero decimal uint64 Discord IDs are accepted as principals. Non-Discord inputs retain their existing behavior. The Discord adapter also no longer falls back to display names as identifiers.

## Validation

- pre-fix regression reproduction failed as expected: all 6 retained inputs reached ingest instead of the principal cap of 4
- focused Node tests: 28 passed
- focused browser CSR-09 tests: 3 passed
- full `@proj-airi/stage-ui` suite: 99 files, 627 tests passed
- `@proj-airi/stage-ui` typecheck passed
- `@proj-airi/discord-bot` typecheck passed
- root typecheck passed across 49 projects
- `pnpm lint` passed with zero errors; seven pre-existing warnings remain only in untouched files
- `git diff --check` passed
- negative controls verified the regression coverage:
  - disabling the principal cap admitted 6 items instead of 4
  - disabling fair scheduling produced `A1, A2, B1, B2` instead of `A1, B1, A2, B2`
- the staged diff and committed diff matched SHA-256 `c0dd59f35c4550e8e76df0301d2ebaac9058690313c5774e9caca517a3aa6348`

`pnpm -F @proj-airi/discord-bot test:run` was also attempted, but that package does not define a `test:run` script. Its typecheck passed.

### DIFF
--- a/packages/stage-ui/src/stores/chat.ts
+++ b/packages/stage-ui/src/stores/chat.ts
@@ -87,6 +87,7 @@ export const useChatOrchestratorStore = defineStore('chat-orchestrator', () => {
 
   const sending = ref(false)
   const pendingQueuedSendCount = ref(0)
+  const pendingSendCancellationHooks = new Set<(sessionId?: string) => void>()
   let ownedActiveTurnSpan: typeof activeTurnSpan.value
 
   async function streamWithStageAdapters(
@@ -404,6 +405,13 @@ export const useChatOrchestratorStore = defineStore('chat-orchestrator', () => {
 
   function cancelPendingSends(sessionId?: string) {
     runtime.cancelPendingSends(sessionId)
+    for (const hook of pendingSendCancellationHooks)
+      hook(sessionId)
+  }
+
+  function onPendingSendsCancelled(hook: (sessionId?: string) => void) {
+    pendingSendCancellationHooks.add(hook)
+    return () => pendingSendCancellationHooks.delete(hook)
   }
 
   function getPendingQueuedSendSnapshot() {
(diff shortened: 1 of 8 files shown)


Woww603:
@codex address that feedback

Please fix the P2 cancellation issue, add a regression test for a hung ingest/provider, run the relevant tests and typechecks, and push the fix to this PR branch.

Woww603:
@codex address that feedback

Please fix the P2 cancellation issue, add a regression test for a hung ingest/provider, run the relevant tests and typechecks, and push the fix to this PR branch.

Woww603:
@codex review

Please verify that the P2 cancellation deadlock is resolved and check the new cancellation race behavior for regressions.

PR #2118, opened 2026-07-26
fix(stage-tamagotchi): restore chat list actions

### LINKED ISSUE #2085: fix(stage-tamagotchi): chat-list add, switch, and delete actions fail
### Describe the bug

In the latest stable desktop app, the conversation list's add, switch, and delete actions do not behave reliably. Buttons can close the sheet without applying the requested operation, and mutations can briefly occur before disappearing.

### Steps to reproduce

1. Open the standalone desktop chat window.
2. Open Conversations.
3. Try each action:
   - create a new conversation
   - select another conversation
   - delete a conversation
4. Reopen the list or wait for chat synchronization.

### Expected behavior

- Add creates and activates a new conversation.
- Selecting a row switches to that conversation.
- Delete removes only the targeted conversation and leaves a valid active fallback.

### Actual behavior

The sheet can close before the click action completes. Create/delete mutations made in the chat window can also be replaced by the main window's next authority snapshot.

### Verified root cause

Two independent failures overlap:

1. The Vaul DrawerContent pointer-release lifecycle can close and unmount the mobile sheet before child button clicks complete.
2. The standalone chat window is a chat-sync follower but the drawer directly mutates its local session store. The authority window then broadcasts its unchanged snapshot and rolls back follower-side create/delete changes.

### Fix

PR #2086:

- replaces the Vaul interaction boundary with a responsive Reka dialog/sheet
- dispatches persistent create/delete commands to the chat-sync authority
- keeps selection local to the requesting window
- adds UI and authority/follower regression coverage

### Validation

- [x] Reproduced from the supplied stable desktop recording.
- [x] Reproduced the state rollback in the repo Electron app.
- [x] Verified add, switch, and delete through real pointer clicks.
- [x] Waited beyond the authority heartbeat and confirmed state remained correct.
- [x] Added automated regressions.
- [x] Full monorepo typecheck passed.
- [x] Lint passed with no errors.

luoling8192:
## What changed

- force the standalone Electron chat window to use the dialog interaction boundary even at narrow widths
- route session creation and deletion from follower windows through the authority window
- keep each follower window's selected session when the authority knows its metadata but has not loaded its messages
- add sync-store and browser regression coverage

## Root cause

The standalone chat window is narrow enough to select the Vaul drawer path. Pointer release could close and unmount that drawer before its child click completed. In addition, the standalone window is a chat-sync follower: creating or deleting sessions directly in its local store was replaced by the next authority snapshot. Session switching could also be rolled back when the authority snapshot contained the session metadata but not its unloaded messages.

The fix keeps persistent mutations on the authority while preserving per-window session selection locally, so creating a session in the standalone window does not switch the main window.

## Validation

- `pnpm typecheck`
- `pnpm exec eslint apps/stage-tamagotchi/src/renderer/pages/chat.vue apps/stage-tamagotchi/src/renderer/stores/chat-sync.test.ts apps/stage-tamagotchi/src/renderer/stores/chat-sync.ts packages/stage-ui/src/components/scenarios/chat/components/sessions-drawer.vue packages/stage-ui/src/components/scenarios/chat/components/sessions-drawer.browser.test.ts`
- `vitest run src/renderer/stores/chat-sync.test.ts` — 12 tests passed
- `vitest run --config vitest.config.ts --project browser src/components/scenarios/chat/components/sessions-drawer.browser.test.ts` — 2 tests passed

Closes #2085

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/pages/chat.vue
+++ b/apps/stage-tamagotchi/src/renderer/pages/chat.vue
@@ -5,7 +5,10 @@ import { shallowRef } from 'vue'
 import InteractiveArea from '../components/InteractiveArea.vue'
 import WindowTitleBar from '../components/Window/TitleBar.vue'
 
+import { useChatSyncStore } from '../stores/chat-sync'
+
 const sessionsDrawerOpen = shallowRef(false)
+const chatSync = useChatSyncStore()
 </script>
 
 <template>
@@ -20,7 +23,12 @@ const sessionsDrawerOpen = shallowRef(false)
       class="interaction-area block"
       h-full w-full p-4 transition="opacity duration-250"
     />
-    <ChatSessionsDrawer v-model="sessionsDrawerOpen" />
+    <ChatSessionsDrawer
+      v-model="sessionsDrawerOpen"
+      presentation="dialog"
+      :create-session="chatSync.requestCreateSession"
+      :delete-session="chatSync.requestDeleteSession"
+    />
   </div>
 </template>
 
--- a/apps/stage-tamagotchi/src/renderer/stores/chat-sync.test.ts
+++ b/apps/stage-tamagotchi/src/renderer/stores/chat-sync.test.ts
@@ -1,6 +1,6 @@
 // @vitest-environment jsdom
 
-import type { ChatSessionsExport } from '@proj-airi/stage-ui/types/chat-session'
+import type { ChatSessionMeta, ChatSessionsExport } from '@proj-airi/stage-ui/types/chat-session'
 import type { Tool } from '@xsai/shared-chat'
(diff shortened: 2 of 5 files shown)


0xSelenicDove:
This PR overlaps the original chat-list fix in #2086: both route follower create/delete through the authority, preserve follower session selection, force Electron onto the dialog interaction boundary, and add regression coverage. #2086 has since accumulated the follow-up correctness fixes from review (targeted sends, strict hydration, stream scoping, deletion cancellation, and unloaded-session error recovery). I also restored the shared mobile Drawer in #2086 while keeping Electron on Dialog (`0e661691e`). Please treat #2086 as the complete fix for #2085.

PR #2126, opened 2026-07-27
feat(stage): persist speech mute and desktop window bounds

luoling8192:
## Summary

- add a persistent speech mute control to the Web, Pocket, and Electron chat surfaces
- stop active synthesis, streaming, queued playback, and current audio when mute is enabled
- prevent new TTS sessions while muted without suppressing emotion, motion, delay, or plugin signals
- persist the final Electron main-window bounds after native move/resize gestures and flush pending config writes during shutdown
- add localized mute/unmute labels for all current stage locales

## Root cause

The existing speech output control only exposed one-shot stop commands, so it could not persist a user preference or gate future TTS sessions.

For the desktop position issue, bounds updates were throttled. Electron main-process HMR terminates the old process directly, so the latest bounds could remain queued and the next process would restore an older saved position. The macOS native click-drag bridge also required an explicit completed-gesture persistence point.

## Validation

- `pnpm exec vitest run apps/stage-tamagotchi/src/main/libs/electron/persistence.test.ts` — 2 tests passed
- `pnpm exec vitest run --config packages/stage-ui/vitest.config.ts packages/stage-ui/src/stores/speech-output-control.test.ts` — 4 tests passed
- `pnpm exec vitest run --config vitest.config.ts src/composables/useStopSpeakingButton.test.ts` from `packages/stage-layouts` — 4 tests passed
- `pnpm run typecheck` — all 49 selected workspaces passed
- `pnpm lint` — 0 errors; 7 existing warnings outside this change
- Electron runtime check — mute state, accessible button state, and local persistence survived a chat renderer reload; test state was restored afterward

## Remaining verification

A physical macOS window drag followed by a real main-process HMR was not manually performed. The behavior is covered by completed native move/resize event handling, shutdown flushing, and the persistence regression test.

### DIFF
--- a/apps/stage-tamagotchi/src/main/libs/electron/persistence.test.ts
+++ b/apps/stage-tamagotchi/src/main/libs/electron/persistence.test.ts
@@ -77,7 +77,10 @@ describe('createConfig', () => {
       app: appMock,
     }))
     vi.doMock('es-toolkit', () => ({
-      throttle: (handler: (...args: unknown[]) => unknown) => handler,
+      throttle: (handler: (...args: unknown[]) => unknown) => Object.assign(handler, {
+        cancel: vi.fn(),
+        flush: vi.fn(),
+      }),
     }))
     vi.doMock('node:fs', () => ({
       existsSync: () => false,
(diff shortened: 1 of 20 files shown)


luoling8192:
Superseded by two scoped pull requests:

- #2127 — Electron window-bounds persistence fix
- #2128 — persistent speech mute feature across Web, Pocket, and Electron

Closing this combined draft so the bug fix and feature can be reviewed and merged independently.

PR #2145, opened 2026-07-28
fix(electron): stop opening remote inspector

luoling8192:
## Summary

- keep Electron remote debugging available during development but only log the inspector endpoint
- stop calling shell.openExternal on every dev launch, which caused the browser inspector page to open or steal focus
- rename the helper around its remaining responsibility and add focused regression coverage
- remove APP_REMOTE_DEBUG_NO_OPEN from automation launchers because the application never read it and the new default behavior makes the flag obsolete

## Verification

- pnpm build:packages
- targeted ESLint on all changed TypeScript files
- 1 targeted Vitest regression assertion


nekomeowww:
No, by default we should have debugger open.

luoling8192:
Superseded by #2156. The replacement keeps the remote inspector default behavior, honors APP_REMOTE_DEBUG_NO_OPEN=true, and loads the flag from app-local env files.

PR #2156, opened 2026-07-29
fix(electron): make remote inspector opening configurable

luoling8192:
## Summary

Supersedes #2145, which GitHub would not allow to be reopened.

- honor `APP_REMOTE_DEBUG_NO_OPEN=true` before opening Electron's remote inspector in the system browser
- load app-scoped `.env` files into the Electron main process while preserving shell environment precedence
- keep the opt-out enabled in automation launchers so CDP remains available without stealing focus
- preserve the existing default behavior when the flag is absent

Developers can opt out locally with:

```dotenv
APP_REMOTE_DEBUG_NO_OPEN=true
```

in `apps/stage-tamagotchi/.env.local`.

## Verification

- `pnpm exec vitest run apps/stage-tamagotchi/src/main/app/debugger.test.ts`
- `pnpm -F @proj-airi/stage-tamagotchi typecheck`
- `pnpm typecheck`
- `pnpm lint`
- verified `electron-vite` loads the ignored local `.env.local` flag

`pnpm -F @proj-airi/stage-tamagotchi build` completed the main and preload builds, then stopped while downloading the external Cubism SDK because `cubism.live2d.com` reset the TLS connection.


apps/stage-tamagotchi/electron.vite.config.ts
```diff
@@ -32,6 +33,15 @@ export default defineConfig({
       },
     },
     plugins: [
+      {
+        name: 'load-app-process-env',
+        config(_config, { mode }) {
+          // Main-process code reads runtime `process.env`, while electron-vite
+          // normally keeps file-based variables in `import.meta.env`. Forward
+          // only APP_ values before Electron starts; shell values take priority.
+          Object.assign(env, loadEnv(mode, import.meta.dirname, 'APP_'))
```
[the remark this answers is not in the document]
luoling8192: Fixed in 0b3732188: APP_ env loading now happens in the top-level defineConfig callback before the Electron config is returned. Note that electron-vite preview builds by default; only an explicit --skipBuild bypasses config evaluation, and this package start script does not use that option. Verified config evaluation with a temporary mode-specific env file, plus stage-tamagotchi typecheck, Vitest, and ESLint.

apps/stage-tamagotchi/src/main/app/debugger.ts
```diff
@@ -38,8 +45,11 @@ export function openDebugger() {
           }
 
           wsUrl = wsUrl.substring(5)
-          console.info(`Inspect remotely: ${remoteDebugEndpoint}/devtools/inspector.html?ws=${wsUrl}`)
-          shell.openExternal(`${remoteDebugEndpoint}/devtools/inspector.html?ws=${wsUrl}`)
+          const inspectorUrl = `${remoteDebugEndpoint}/devtools/inspector.html?ws=${wsUrl}`
+          console.info(`Inspect remotely: ${inspectorUrl}`)
+
+          if (!/^true$/i.test(env.APP_REMOTE_DEBUG_NO_OPEN || ''))
+            void shell.openExternal(inspectorUrl)
```
nekomeowww, line 52: Again, if you want `APP_REMOTE_DEBUG_NO_OPEN` to work, simply check `APP_REMOTE_DEBUG_NO_OPEN` then prevent `shell.openExternal` then it should work.

apps/stage-tamagotchi/src/main/app/debugger.ts
```diff
@@ -16,7 +17,13 @@ export function setupDebugger() {
...
 }
 
-export function openDebugger() {
+/**
+ * Announces an inspector URL for the first available Electron renderer target.
+ *
+ * Developers may keep CDP enabled without opening the system browser by
+ * setting `APP_REMOTE_DEBUG_NO_OPEN=true`.
+ */
+export function announceDebuggerEndpoint() {
```
nekomeowww: While announce works, but openDebugger is openDebugger, you cannot just replace it with `announceDebuggerEndpoint`

nekomeowww (changes requested):

PR #2159, opened 2026-07-29
fix: deliver X / Twitter configuration to its service

### LINKED ISSUE #2154: bug: X / Twitter settings report configured but are not delivered to the service
## Describe the bug

The shared X / Twitter settings page accepts an API Key, API Secret, Access Token, and Access Token Secret, then displays **“X / Twitter is properly configured!”** as soon as all four local fields are non-empty. However, saving those settings does not deliver a configuration that the X service accepts, so the UI can report success even though the integration is not operational.

This was found while validating the English manual against the AIRI 0.11.3 UI and current source.

## Steps to reproduce

1. Open **Settings → Modules → X / Twitter** in the web or desktop app.
2. Enter any non-empty values in all four credential fields.
3. Select **Save**.
4. Observe that the page reports the integration as properly configured.
5. Try to use the X / Twitter integration.

## Actual behavior

- The success state is derived only from whether the four local fields are non-empty:
  - `packages/stage-ui/src/stores/modules/twitter.ts`
- Saving publishes `ui:configure` with `moduleName: "twitter"`.
- The X service adapter handles the event only when `moduleName === "x"`:
  - `services/twitter-services/src/adapters/airi-adapter.ts`
- Therefore the adapter ignores the settings payload, while the UI still presents a successful configuration state.
- The separate service is also disabled by default unless `ENABLE_AIRI=true`, so the UI currently has no end-to-end connection/health confirmation.

## Expected behavior

AIRI should not claim that X / Twitter is properly configured unless the running service has received and validated the configuration.

At minimum:

- use one consistent module identifier across the UI and service;
- expose a real connection/configuration status to the UI;
- show an unavailable or disconnected state when the service is not running;
- avoid treating four non-empty fields as proof of a working integration;
- add a focused regression test for the configuration event contract.

## Impact

Users may enter sensitive X credentials into a UI that reports success although the integration cannot use them. This creates a misleading setup flow and unnecessary credential exposure.

## System info

Reproduced by source inspection on the current repository state while reviewing AIRI 0.11.3 documentation. The affected settings code is shared by the web and Electron desktop apps.

## Validations

- [x] Checked open and closed repository issues for “Twitter,” “X / Twitter,” the module identifier, and the configured message.
- [x] The only related result, #610, is a closed multi-problem report about empty settings pages and does not track this configuration-contract defect.
- [x] This is a concrete bug rather than a usage question.

0xSelenicDove:
Closes #2154

## Summary
- Route X settings through the consistent `x` module identifier and server-delivered `module:configure` contract.
- Report live X service availability instead of treating locally populated credentials as a working integration.
- Add the configuration-routing regression test and localized status messages.

## Root cause
The UI targeted `twitter`, while the service expected `x`; additionally, the service listened for `ui:configure` although the runtime forwards the event as `module:configure`.

## Validation
- `pnpm exec vitest run packages/server-runtime/src/setupApp.liveness.test.ts`
- `pnpm -F @proj-airi/server-runtime typecheck`
- `pnpm lint`

`pnpm -F @proj-airi/twitter-services typecheck` remains blocked by existing unused variables in `src/parsers/profile-parser.ts` and `src/parsers/tweet-parser.ts`.

### DIFF
--- a/packages/i18n/src/locales/en/settings.yaml
+++ b/packages/i18n/src/locales/en/settings.yaml
@@ -786,7 +786,8 @@ pages:
       access-token-secret: Access Token Secret
       access-token-secret-description: Your X / Twitter access token secret
       access-token-secret-placeholder: Enter your X / Twitter access token secret
-      configured: X / Twitter is properly configured!
+      service-online: X / Twitter service is online. Credentials are sent when saved.
+      service-offline: X / Twitter service is unavailable. Start it before saving credentials.
     web-search:
       title: Web Search
       description: Let AIRI search the web for current information
--- a/packages/i18n/src/locales/es/settings.yaml
+++ b/packages/i18n/src/locales/es/settings.yaml
@@ -749,7 +749,8 @@ pages:
       access-token-secret: Secreto de token de acceso
       access-token-secret-description: Su secreto de token de acceso de X / Twitter
       access-token-secret-placeholder: Ingrese su secreto de token de acceso de X / Twitter
-      configured: '¡X / Twitter está configurado correctamente!'
+      service-online: El servicio de X / Twitter está en línea. Las credenciales se enviarán al guardar.
+      service-offline: El servicio de X / Twitter no está disponible. Inícialo antes de guardar las credenciales.
     web-search:
       title: Web Search
       description: Let AIRI search the web for current information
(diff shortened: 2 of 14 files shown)


0xSelenicDove:
Closing this PR as services/twitter-services is being deprecated in favor of migrating X/Twitter integration to standard OAuth 2.0 (per discussion in #2154).

PR #2185, opened 2026-07-30
fix(ci): pin Node to 24 in CI and autofix workflows

lulu0119:
## Summary
- Pin `actions/setup-node@v6` to Node 24 in `ci.yml` and `autofix.yaml`, replacing `lts/*`.

## Why
Peeled off from #1966 (Steam desktop sign-in) at maintainer request, since it's orthogonal to Steam auth.

Two failures were observed on that branch while `lts/*` was in use:
- `lts/*` intermittently failed `setup-node@v6` with `manifest.filter is not a function`, cancelling PR builds.
- Node 22 (an earlier attempted pin) lacks `CloseEvent`, which breaks `better-ws` unit tests.

Node 24 is the current LTS and has both issues fixed.

## Test plan
- [ ] CI green on this PR (lint/typecheck/unit/build across all four `setup-node` jobs).

### DIFF
--- a/.github/workflows/autofix.yaml
+++ b/.github/workflows/autofix.yaml
@@ -17,7 +17,7 @@ jobs:
       - uses: pnpm/action-setup@v5
       - uses: actions/setup-node@v6
         with:
-          node-version: lts/*
+          node-version: 24
           cache: pnpm
 
       - run: pnpm install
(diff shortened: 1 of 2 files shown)


nekomeowww:
What is manifest.filter is not a function?

PR #2186, opened 2026-07-30
fix(ui-server-auth): preserve apiServerUrl across SPA verify-email navigation

lulu0119:
## Summary
- Carry `api_server_url` on the `/sign-in` -> `/verify-email` router push so the verification page resolves the same API server that created the account.
- Key `getServerAuthBootstrapContext()`'s cache by `window.location.href` instead of a single module-level value, so each SPA route re-resolves its own context.

## Why
Peeled off from #1966 (Steam desktop sign-in) at maintainer request, since it's an independent bug unrelated to Steam.

This auth UI is a client-side SPA: vue-router navigates from `/sign-in` to `/verify-email` via `pushState`, changing `window.location.href` without a page reload. The bootstrap context used a single module-level cache populated on first call, so every later call on the same page load reused the *first* visited route's resolved `apiServerUrl` (or `null`) regardless of the current URL's query string.

Concretely: a user lands on `/sign-in` with no `api_server_url` (cached as `null`), signs up against a local dev server, and gets redirected to `/verify-email?api_server_url=http://localhost:3000` — but the stale cached `null` was returned, so verify-email silently fell back to the production `SERVER_URL` instead of the local server that created the account.

Reproduced via an instrumented debug session before this fix; confirmed resolved after.

## Test plan
- [x] `pnpm exec vitest run apps/ui-server-auth/src/modules/server-auth-context.test.ts` — 7/7 passing, including a new regression test for the SPA-navigation cache staleness.
- [x] `pnpm eslint` on changed files — clean.
- [ ] CI green.

### DIFF
--- a/apps/ui-server-auth/src/modules/server-auth-context.test.ts
+++ b/apps/ui-server-auth/src/modules/server-auth-context.test.ts
@@ -58,4 +58,34 @@ describe('ui-server-auth bootstrap context', () => {
 
     expect(getServerAuthBootstrapContext()?.apiServerUrl).toBe('https://airi-server-dev.up.railway.app')
   })
+
+  // ROOT CAUSE:
+  //
+  // This standalone auth UI is a client-side SPA: vue-router navigates from
+  // /sign-in to /verify-email via pushState, changing window.location.href
+  // without a page reload. getServerAuthBootstrapContext() used to cache its
+  // result in a single module-level variable on first call, so every later
+  // call on the same page load reused the FIRST visited route's resolved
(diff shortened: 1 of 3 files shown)


lulu0119:
Closing: the new Steam PR stack (#2226/#2227/#2228) does not depend on this fix, and current main has no user-visible path that hits the stale-cache bug. If we still want the fix, the cleaner approach is to drop the module cache entirely (stateless getServerAuthBootstrapContext) and carry api_server_url on the /sign-in -> /verify-email push; verified 7/7 tests pass with that version.

PR #2187, opened 2026-07-30
fix(server): serialize schema migration with a Postgres advisory lock

lulu0119:
## Summary
- `migrateDatabase` now takes the `Pool`, checks out one client, and holds a Postgres session advisory lock (`pg_advisory_lock`/`pg_advisory_unlock`) for the duration of the Drizzle `migrate()` call, so only one replica applies schema changes at a time.
- Add migration `0018`, an idempotent repair of `0017`'s objects (`CREATE ... IF NOT EXISTS`, `DROP CONSTRAINT IF EXISTS` before re-`ADD`) for environments where a boot race already created these tables/constraints/indexes without recording `0017` in the migrations journal. This does **not** edit the already-applied `0017` file — per review policy, already-applied migrations should be regenerated/re-applied via a new migration instead.

## Why
Peeled off from #1966 (Steam desktop sign-in) at maintainer request, since this is orthogonal to Steam auth (`db.ts` "why do we need this?" was flagged in review — moving it to its own PR for focused discussion).

Railway (and other multi-replica deploys) boot several processes at once. Without coordination, each replica can race `CREATE TABLE`/`ADD CONSTRAINT` and the loser dies with Postgres `42P07` before reaching HTTP listen, which times out `/livez` healthchecks. This was previously observed on `server-dev`: migration `0017`'s objects appeared to have been created by a losing replica without a corresponding `drizzle.__drizzle_migrations` row, orphaning them for a subsequent clean-replica retry of `0017`.

Note: this necessity has **not** been re-verified with a fresh multi-replica falsification experiment on current `server-dev` as its own PR — happy to run that separately if maintainers want harder proof before merging.

## Test plan
- [x] `pnpm exec vitest run apps/server/src/libs/tests/db-migrate-lock.test.ts` — 2/2 passing (lock acquired/released around success and around a thrown error).
- [x] `pnpm -F @proj-airi/server typecheck` — clean.
- [x] `pnpm eslint` on changed files — clean.
- [x] New migration `0018`'s SQL body is byte-identical to the idempotent-guarded version already exercised on `server-dev` in the original branch history.
- [ ] CI green.

### DIFF
--- a/apps/server/src/libs/db.ts
+++ b/apps/server/src/libs/db.ts
@@ -1,3 +1,5 @@
+import type { Pool } from 'pg'
+
 import type { Env } from './env'
 
 import pg from 'pg'
@@ -14,6 +16,12 @@ const logger = useLogger('db')
 export type Database = ReturnType<typeof createDrizzle>['db']
 
 type DrizzleEnv = Pick<Env, 'DATABASE_URL' | 'DB_POOL_MAX' | 'DB_POOL_IDLE_TIMEOUT_MS' | 'DB_POOL_CONNECTION_TIMEOUT_MS' | 'DB_POOL_KEEPALIVE_INITIAL_DELAY_MS'>
+
+/**
+ * Session-scoped advisory lock key shared by every replica of this service.
+ * Arbitrary stable int4 — must stay identical across deploys.
+ */
+const SCHEMA_MIGRATE_LOCK_KEY = 872_314_059
 
 // NOTICE: pg is imported statically here. The OTEL instrumentation hooks are
 // registered via --import ./instrumentation.ts (preload) which runs before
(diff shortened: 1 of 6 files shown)


apps/server/src/libs/tests/db-migrate-lock.test.ts
nekomeowww, line 1: Why db changes will introduce unit test?

apps/server/src/libs/db.ts
```diff
@@ -36,6 +44,42 @@ export function createDrizzle(env: DrizzleEnv) {
...
 
-export function migrateDatabase(db: Database) {
-  return migrate(db, migrations)
+/**
+ * Runs `run` while holding a Postgres session advisory lock on `client`.
+ *
+ * Lock and unlock must use the same session: advisory locks are session-scoped,
+ * and unlock on another connection would leave the lock held.
+ */
+export async function withSessionAdvisoryLock<T>(
```
nekomeowww, line 53: What is advisory?

PR #2189, opened 2026-07-30
feat(server): add Steam OpenID sign-in and account linking plugin

lulu0119:
## Summary

- Adds a self-contained better-auth plugin (`libs/auth-plugins/steam.ts`) implementing Steam OpenID 2.0 sign-in, account linking, and callback verification via "dumb mode" — Steam's login protocol is not OAuth2/OIDC, so it cannot be registered as a `socialProviders` entry.
- New sign-ups get a placeholder `<steamid64>@steam.placeholder.local` address with `emailVerified: true`, mirroring how Apple Sign In accounts without a real email get `<sub>@apple.placeholder.local`.
- Wires Steam into `apps/ui-server-auth`'s web sign-in page and profile "Connected accounts" section:
  - Extends the shared `OAuthProvider` type and `defaultSignInProviders` list (`packages/stage-ui`) with `steam`.
  - `requestSocialSignInRedirect` now posts to `/sign-in/steam` (no `provider` field) instead of `/sign-in/social` when the provider is Steam.
  - Adds `toLinkedAccountsClient` / `withSteamLinking` to route Steam account linking through the plugin's `/link/steam` endpoint via the client's `$fetch`, since better-auth's `/link-social` validates `provider` against a fixed social-provider enum and would reject `steam`. Unlinking needs no special-casing — `/unlink-account` already takes a free-form `providerId`.

No Steam Web API key is required for this browser-based OpenID flow.

We intentionally do not depend on community Steam packages (e.g. `better-auth-steam`) or the still-unmerged upstream draft ([better-auth#4877](https://github.com/better-auth/better-auth/pull/4877)). Steam never returns an email, and we need sign-up that does not ask the user for one plus first-class account linking; available options either require an email at sign-in, lack linking, or are abandoned / blocked — so shipping a small in-tree plugin is the safer auth dependency for this requirement.

## Test plan

- [x] `pnpm -F @proj-airi/server exec vitest run src/libs/auth-plugins/steam.test.ts` — sign-in, new/existing user, error redirect, account linking (own account + conflict) — 6/6 passing
- [x] `pnpm -F @proj-airi/ui-server-auth exec vitest run` — includes new tests for the Steam sign-in request body and the linked-accounts client wrapper — 34/34 passing
- [x] `pnpm -F @proj-airi/server typecheck`
- [x] `pnpm -F @proj-airi/ui-server-auth typecheck`
- [x] `pnpm -F @proj-airi/stage-ui typecheck`
- [x] `pnpm lint` on touched files

## Follow-ups

- Desktop Steam ticket sign-in (stacked PR): explicit user choice only — no startup silent create; ticket when the user picks Steam, otherwise browser OIDC; link Steam via existing `/link/steam` OpenID (no ticket-bind API in that batch).
- Steam persona name/avatar: if we want display names beyond `Steam User <id>`, fetch `GetPlayerSummaries` **inside the Steam OpenID plugin** on sign-in/link. Not in this PR batch.


server/apps/api/src/libs/auth-plugins/steam.test.ts
```diff
@@ -0,0 +1,258 @@
+import { betterAuth } from 'better-auth'
+import { drizzleAdapter } from 'better-auth/adapters/drizzle'
+import { beforeAll, describe, expect, it, vi } from 'vitest'
+
+import { mockDB } from '../mock-db'
+import { steam } from './steam'
+
+import * as schema from '../../schemas'
+
+const STEAM_ID = '76561198012345678'
```
nekomeowww, line 11: What is this?

server/apps/api/src/libs/auth-plugins/steam.ts
```diff
@@ -0,0 +1,235 @@
...
+import { generateState, parseState } from 'better-auth/oauth2'
+
+import * as z from 'zod'
+
+const STEAM_OPENID_ENDPOINT = 'https://steamcommunity.com/openid/login'
+const STEAM_OPENID_NS = 'http://specs.openid.net/auth/2.0'
+const STEAM_OPENID_IDENTIFIER_SELECT = 'http://specs.openid.net/auth/2.0/identifier_select'
+
+/** Matches `https://steamcommunity.com/openid/id/<steamid64>`. */
+const STEAM_CLAIMED_ID_PATTERN = /^https:\/\/steamcommunity\.com\/openid\/id\/(\d{17})$/
```
nekomeowww, line 12: Do we have existing Steam SDK in Node.js ecosystem to use for easier integration?

apps/ui-server-auth/src/modules/sign-in.test.ts
```diff
@@ -136,6 +136,31 @@ describe('ui-server-auth sign-in flow helpers', () => {
...
+      'https://api.airi.test/api/auth/sign-in/steam',
+      expect.objectContaining({ method: 'POST' }),
+    )
+
+    const init = fetchImpl.mock.calls[0]?.[1]
+    expect(JSON.parse(String(init?.body))).toEqual({
+      callbackURL: 'https://api.airi.test/api/auth/oauth2/authorize?client_id=airi-stage-web',
+    })
+  })
+
```
nekomeowww, line 151: Not necessary.

packages/stage-pages/src/pages/settings/account/account-settings-page.vue
```diff
@@ -145,7 +145,7 @@ const {
   unlink: unlinkLinkedProvider,
   link: linkLinkedProvider,
 } = useLinkedAccounts({
-  client: authClient,
+  client: withSteamLinking(authClient),
```
nekomeowww: Why authClient will have to be `withSteamLink` which leaked the information of `steam` into the Account Settings Page? 

This violates the 

- https://github.com/alysivji/notes/blob/main/software-engineering/philosophy_of_software_design.md
- https://en.wikipedia.org/wiki/SOLID

rules that the consumer side of the module should not have to understand what steam is in order to use authClient.

packages/stage-ui/src/composables/use-linked-accounts.test.ts
```diff
@@ -124,3 +145,49 @@ describe('useLinkedAccounts', () => {
...
+    const result = await wrapped.linkSocial({ provider: 'steam', callbackURL: '/profile', errorCallbackURL: '/profile?error=steam' })
+
+    expect(linkSocial).not.toHaveBeenCalled()
+    expect($fetch).toHaveBeenCalledWith('/link/steam', {
+      method: 'POST',
+      body: { callbackURL: '/profile', errorCallbackURL: '/profile?error=steam' },
+    })
+    expect(result.data?.url).toBe('https://steamcommunity.com/openid/login?...')
+  })
+})
```
nekomeowww: What does this part do?

packages/stage-ui/src/composables/use-linked-accounts.ts
```diff
@@ -93,6 +93,62 @@ export interface UseLinkedAccountsArgs {
...
+      if (args.provider !== 'steam')
+        return client.linkSocial(args)
+
+      return client.$fetch<{ url?: string, redirect?: boolean, status?: boolean }>('/link/steam', {
+        method: 'POST',
+        body: { callbackURL: args.callbackURL, errorCallbackURL: args.errorCallbackURL },
+      })
+    },
+  }
+}
```
nekomeowww: Then why not routing and configure request params through args.provider instead of building your own wrapper?

packages/stage-ui/src/composables/use-linked-accounts.test.ts
```diff
@@ -1,8 +1,29 @@
...
+// Removal condition: vitest ships a `Mock` type that preserves generics.
+function fakeRawClient(overrides: Partial<RawLinkedAccountsClient>): RawLinkedAccountsClient {
+  return {
+    listAccounts: vi.fn(async () => ({ data: [], error: null })),
+    unlinkAccount: vi.fn(async () => ({ data: null, error: null })),
+    linkSocial: vi.fn(async () => ({ data: null, error: null })),
+    $fetch: vi.fn(async () => ({ data: null, error: null })) as unknown as RawLinkedAccountsClient['$fetch'],
+    ...overrides,
+  }
+}
```
nekomeowww: Also, why fake client?

PR #2204, opened 2026-08-03
fix(server): honor mobile analytics consent

Neko-233:
## What changed

- add an explicit mobile analytics-consent contract for chat and speech requests
- fail closed for mobile requests that do not grant analytics consent
- suppress optional PostHog and product-event writes when consent is denied
- never export mobile chat prompts, responses, or TTS text to Langfuse, even after analytics opt-in
- keep operational billing, request logs, and content-free metrics intact

## Why

AIRI Lite exposes an opt-in product analytics setting and promises that analytics excludes chat content. The API previously emitted server-side analytics independently of that setting and could trace prompt or output content.

## Impact

Existing web and Electron behavior remains unchanged unless a client sends an explicit denial. Mobile analytics now follows the in-app consent state, while mobile content remains excluded from Langfuse.

## Checks

- `pnpm exec vitest run server/apps/api/src/routes/openai/v1/route.test.ts server/apps/api/src/services/domain/llm-tracing/index.test.ts` (77 tests passed)
- `pnpm -F @proj-airi/api-server typecheck`
- repository commit hook (`moeru-lint --fix`) passed for all 8 changed files

### DIFF
--- a/server/apps/api/src/routes/openai/v1/operations/chat-completions/index.ts
+++ b/server/apps/api/src/routes/openai/v1/operations/chat-completions/index.ts
@@ -23,6 +23,7 @@ export interface ChatCompletionsOperationRequest {
   sessionId?: string
   roundId?: string
   appSurface?: AiGenerationAppSurface
+  analyticsEnabled: boolean
   abortSignal?: AbortSignal
 }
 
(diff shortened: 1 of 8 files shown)


server/apps/api/src/routes/openai/v1/operations/chat-completions/index.ts
```diff
@@ -71,18 +72,20 @@ export function chatCompletions(deps: V1RouteDeps): GatewayCallback<'chat.comple
...
-      action: 'completion_requested',
-      status: 'started',
-      source: 'openai.chat.completions',
-      model: requestModel,
-      metadata: {
-        stream,
-        message_count: Array.isArray(body.messages) ? body.messages.length : null,
-      },
-    })
+    if (input.analyticsEnabled) {
```
nekomeowww, line 75: This shouldn't be passed and configured by POST body?

server/apps/api/src/routes/openai/v1/operations/chat-completions/index.ts
```diff
@@ -71,18 +72,20 @@ export function chatCompletions(deps: V1RouteDeps): GatewayCallback<'chat.comple
       stream,
       messageCount: Array.isArray(body.messages) ? body.messages.length : undefined,
     }).log('chat completion request')
-    void deps.productEventService.track({
```
nekomeowww, line 74: cc @luoling8192 how come one user using our LLM service will be able to omit tracing and data collection? Should this happen? I thought this part should be able to be handled by internal deployed LLM gateway instead of here.

PR #2271, opened 2026-08-12
fix(stage-ui): restore context broadcast delivery

LauraGPT:
## Summary

- emit Stage context and stream events directly through the BroadcastChannel-backed Eventa context
- remove the local `linkChannel` hop that turns outbound events into inbound events under `@moeru/eventa@1.0.0-beta.13`
- preserve remote-only delivery because listeners continue to accept only events carrying a raw BroadcastChannel message

## Context

After `@moeru/eventa` was pinned from beta.14 to beta.13 in a9906bf13, the existing cross-context test consistently received zero updates. In beta.13, `linkChannel` converts the event to inbound before forwarding it; the broadcast adapter only posts outbound or undirected events, so the event never reaches the native BroadcastChannel.

## Validation

- focused context-channel test: 1/1 passed
- Stage UI Node suite: 106 files, 617/617 tests passed
- Stage UI Chromium suite (`CI=1`): 6 files, 20/20 tests passed
- Stage UI typecheck passed
- changed-file lint: 0 warnings, 0 errors

### DIFF
--- a/packages/stage-ui/src/stores/mods/api/context-channel.ts
+++ b/packages/stage-ui/src/stores/mods/api/context-channel.ts
@@ -1,6 +1,6 @@
 import type { ChatStreamEvent, ContextMessage } from '../../../types/chat'
 
-import { createContext, defineEventa, linkChannel } from '@moeru/eventa'
+import { defineEventa } from '@moeru/eventa'
 import { createContext as createBroadcastChannelContext } from '@moeru/eventa/adapters/broadcast-channel'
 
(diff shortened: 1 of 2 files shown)


LauraGPT:
Included the lockfile normalization produced by this repository's required autofix workflow so the check can complete cleanly. The functional source change remains limited to the Eventa channel-direction fix; all CI checks passed on the prior clean-source head.

nekomeowww:
No longer needed. Eventa bumpped.

PR #2273, opened 2026-08-12
fix: guard kokoro-local provider init timing on direct navigation

stz-2044:
## Summary
When navigating directly to the speech settings page, the provider may not be registered from the catalog yet, so `getProviderConfig()` returns `undefined` and reading `config.model` throws. The model combobox can also write back `v-model` before `onMounted` runs (provider registration is an async synced action).

## Changes
- `packages/stage-pages/.../kokoro-local.vue`: await `ensureProvider()` when config is missing, then re-read; guard `set()`/`watch()` against missing config; fall back to `{}` where a read is unavoidable.
- `pnpm-lock.yaml`: remove iOS `DerivedData` lock entries that were accidentally committed (Xcode build cache), add missing `web-worker` optional marker.

## Verification
- Direct navigation to the page no longer throws.
- Model selection persists after provider registration.
- `pnpm install --frozen-lockfile` passes locally.

### DIFF
--- a/packages/stage-pages/src/pages/settings/providers/speech/kokoro-local.vue
+++ b/packages/stage-pages/src/pages/settings/providers/speech/kokoro-local.vue
@@ -58,10 +58,19 @@ const model = computed({
     return getDefaultKokoroModel(hasWebGPU.value, fp16Supported.value)
   },
   set(val: string) {
-    // The combobox can write back the v-model before onMounted runs, at which
-    // point the provider may not be registered yet and synced actions are still
-    // async. Skip the write in that window; onMounted persists the default model
-    // right after registration, and later user selections always hit a config.
+    // NOTICE:
+    // Why this workaround is needed: the combobox can write back the v-model
+    // before onMounted runs, when the provider may not be registered yet and
+    // synced actions are still async. Skipping the write in that window is
+    // safe because onMounted persists the default model right after
+    // registration, and later user selections always hit a config.
+    // Root cause: direct navigation registers the provider inside onMounted
+    // (ensureProvider), so a pre-mount combobox write would hit
+    // getProviderConfig() === undefined and throw.
+    // Source: PR #2273 review (codex bot), kokoro-local direct navigation path.
+    // Removal condition: when the settings page guarantees the provider is
+    // registered before the combobox can write (e.g. provider pre-registered
+    // at app init), this guard can be deleted.
     const config = providerStore.getProviderConfig(providerId)
     if (config)
       config.model = val
@@ -130,6 +139,9 @@ onMounted(async () => {
       await providerStore.ensureProvider(providerId, providerId, {
         model: getDefaultKokoroModel(hasWebGPU.value, fp16Supported.value),
         voiceId: '',
+        // Match getDefaultProviderConfig() shape (baseUrl: '' for providers
+        // without a base URL) so shouldListProvider() comparison stays clean.
+        baseUrl: '',
       })
     }
     const config = providerStore.getProviderConfig(providerId) ?? {}
--- a/packages/stage-pages/src/pages/settings/providers/speech/kokoro-local.vue
+++ b/packages/stage-pages/src/pages/settings/providers/speech/kokoro-local.vue
@@ -135,6 +135,14 @@ onMounted(async () => {
   hasWebGPU.value = capabilities.supported
   fp16Supported.value = capabilities.fp16Supported
 
+  // Refresh the provider's default config: metadata is selected at store
+  // setup (before capability detection), so Kokoro's default model was
+  // computed from an empty cache. Seeding a capability-accurate default
+  // while the dirty comparison still uses the stale one would make
+  // shouldListProvider() treat this passive seed as user-modified and
+  // list Kokoro as dirty/unconfigured.
+  providersStore.refreshProviderDefaultConfig(providerId)
+
   try {
     voicesLoading.value = true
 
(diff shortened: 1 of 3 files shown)


stz-2044:
Thanks for the review! Fixed in `20e8a83`: `ensureProvider` is now seeded with capability-aware defaults (`model: getDefaultKokoroModel(hasWebGPU, fp16Supported)`, `voiceId: ""`) matching `createProviderConfig`, and the fallback persist also passes `fp16Supported` — so fp16-capable hardware keeps its `fp16-webgpu` default and the saved value always matches the computed getter.

stz-2044:
Both suggestions addressed in `f1d977bb`:

1. **baseUrl seed**: `ensureProvider` now seeds `baseUrl: ""` alongside `model`/`voiceId`, matching the `getDefaultProviderConfig()` shape so the saved object is not dirty for `shouldListProvider()`.
2. **NOTICE block**: the init-timing workaround comment now uses the required `// NOTICE:` format with root cause summary, source reference, and removal condition.

Thanks for the review!

stz-2044:
Thanks for the catch — the race is real. Direct navigation mounts this page before the app-level preload (`useInferencePreload` → `detectWebGPU`, called from `App.vue` `onMounted`) populates the capability cache, so the sync cache read returned `null` and `fp16Supported` fell back to `false` — seeding `fp32-webgpu` on fp16-capable hardware until a reload.

Fixed in `e608514bc`: the page now `await`s `detectWebGPU()` before fetching models and seeding the config. `detectWebGPU()` deduplicates concurrent calls via `pendingDetection`, so when the app preload is already in flight this awaits the same promise instead of triggering a second adapter request — no duplicate detection, no extra cost.

Verified: `vue-tsc --noEmit` passes (0 errors).

stz-2044:
Good catch — the drift was real. `providerMetadata` (including Kokoro's capability-dependent `defaultConfig`) is selected at store setup, before capability detection completes, so seeding a capability-accurate default after `await detectWebGPU()` made `isProviderConfigDirty()` see a user-modified config and `shouldListProvider()` list Kokoro as dirty.

Fixed in `971bb969e`:
1. **`refreshProviderDefaultConfig(providerId)`** added to the providers store — re-selects a provider's `defaultConfig` with current runtime state (same `createProviderConfig` → `getSchemaDefault` path used at metadata selection).
2. **The page now refreshes the default right after `await detectWebGPU()`**, then seeds with `getDefaultProviderConfig(providerId)` — the exact same source the dirty comparison reads. So the passive seed always equals the comparison baseline: `isProviderConfigDirty()` stays `false`, while fp16-capable hardware still keeps its `fp16-webgpu` default (round 1 requirement intact).

Verified: `vue-tsc --noEmit` passes for both `stage-ui` and `stage-pages`; 618/618 node tests pass (the browser-mode test files did not run here — local Playwright browser binary is missing, unrelated to this change).

stz-2044:
Confirmed — the child `SpeechProviderSettings` mounts first and `initializeProvider()` seeds the provider with the stale pre-detection default before the parent page awaits `detectWebGPU()`, so my `ensureProvider()` guard skipped the seed and the stale `fp32-webgpu` config stayed loaded while the refreshed baseline moved to `fp16-webgpu`.

Fixed in `3e37763c2` — `refreshProviderDefaultConfig()` now:
1. Snapshots the previous default before refreshing.
2. Refreshes `defaultConfig` as before.
3. If the saved config still **exactly matches** the stale default (a passive seed, not a user edit), replaces it with the refreshed default via a new synchronous `replaceProviderConfig()` on the config store — a local mutation with no remote round-trip that swaps the config object by reference, so `configs`/`providerCredentials` and the dirty comparison all pick it up immediately.

A real user edit never matches the stale default, so it is never clobbered. Verified: `vue-tsc --noEmit` passes for both packages; 618/618 node tests pass.

stz-2044:
Confirmed — `SpeechProviderSettings` mounts first and its `apiKey`/`baseUrl` computed setters write passive fields (e.g. `apiKey: ''`) into the seeded config, so a full `JSON.stringify` equality check failed and `replaceProviderConfig()` never ran.

Fixed in `b322aff5c`: the untouched-seed check now compares **only the fields that make up the provider default** — if every default field still equals the stale default, the config is the passive seed and is replaced with the refreshed default; any deviation means a user edit and is never clobbered (passive fields like `apiKey` are ignored either way).

Verified: `vue-tsc --noEmit` passes for both packages; the node test suite passes (618 tests; browser-mode files don't run locally due to a missing Playwright binary, unrelated to this change).

stz-2044:
Confirmed — on reload with an empty WebGPU cache, the stale default snapshot is also `fp32-webgpu`, so a user who deliberately picked `fp32-webgpu` on fp16 hardware is value-identical to a passive seed and would get replaced. Value equality alone cannot distinguish the two.

Fixed in `62f1dcea1` — replacement now requires the caller to opt in via `replaceUntouchedSeed`:
- `refreshProviderDefaultConfig(providerId, { replaceUntouchedSeed })` only replaces when the flag is explicitly `true` and the default fields are untouched.
- The kokoro-local page captures `hadConfigBeforeMount` in `setup` (which runs before the child `SpeechProviderSettings` can seed the provider) and passes `replaceUntouchedSeed: !hadConfigBeforeMount` — a persisted config is user-owned and never replaced; only a config created during this mount (a true passive seed) qualifies.

So a manually saved model choice survives reloads, while the direct-navigation seed still gets corrected. Verified: `vue-tsc --noEmit` passes for both packages; node tests pass.

stz-2044:
Confirmed — the child `SpeechProviderSettings` seeding the stale default changes `model` and fires the watcher, loading `fp32-webgpu` while the parent `onMounted` is still awaiting capability detection; the refresh then queues the correct load behind an unnecessary stale one.

Fixed in `bd5fc19e3`: a `initialSeeding` flag gates the watcher until `onMounted` finishes initial seeding. The parent already loads the correct model explicitly after the refresh (in the same `onMounted`), so no model load is lost — the watcher simply no longer races ahead with the stale default. The gate is released in the `finally` block, so a failed seed also resumes the watcher safely.

Verified: `vue-tsc --noEmit` passes for both packages.

stz-2044:
Confirmed — in the synced Electron runtime, `refreshProviderDefaultConfig()` mutated store-local `providerMetadata` and `replaceProviderConfig()` mutated the config store, neither routed through `synced.actions`. A non-leader renderer direct-navigating would refresh the default locally while the leader kept comparing against the stale default.

Fixed in `96943acac`: both actions are now in their stores' `synced.actions` blocks (`refreshProviderDefaultConfig` on the provider store, `replaceProviderConfig` on the config store). With `pinia-plugin-synced`'s leader-execution semantics, the refresh runs on the leader and the resulting config state propagates to every renderer, so the dirty comparison uses the capability-aware default everywhere.

One note: `providerMetadata` itself is a per-renderer setup-time constant in the existing store design (not synced state), so a non-leader window that already ran its own dirty comparison before the refresh broadcast could still show a stale listing until its computed re-evaluates against the synced config — that's a pre-existing architecture trait outside this PR's scope, and the navigation window itself is always consistent.

Verified: `vue-tsc --noEmit` passes for both packages; node tests pass.

PR #2290, opened 2026-08-15
feat(server): stream official ASR over WebSocket

luoling8192:
## What changed

Official ASR now uses one WebSocket per VAD speech segment. PCM frames go upstream in order; interim Aliyun transcript snapshots are sent back while speech is still in progress, then replaced by the final sentence. Stop waits for final text; cancel and failures close both sides.

The browser carries its AIRI bearer in a WebSocket subprotocol, not a URL query parameter. The API validates the session, applies startup/audio limits, and mediates Aliyun NLS. The previous streaming fetch/SSE path is removed. The direct user-configured Aliyun provider is unchanged.

## Verification

- Focused ASR route, session, provider, and hearing tests passed locally.
- Full pnpm typecheck passed (52 tasks).
- pnpm lint passed with 0 errors and existing warnings.
- After removing unrelated desktop-layout changes from this PR, the previously failing 10 controls-island tests pass locally.

## Remaining gates

- Real authenticated microphone and Aliyun streaming smoke has not been run.
- ASR usage and billing accounting still needs a product/backend decision before merge.
- CI for the latest head has one failing Unit Test check in unrelated desktop placement cases; see the CI follow-up comment.


packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
```diff
@@ -0,0 +1,240 @@
...
+  if (!stream)
+    throw new TypeError('Audio stream or file is required for official transcription.')
+  return stream as ReadableStream<AudioChunk>
+}
+
+function toUint8Array(chunk: AudioChunk): Uint8Array {
+  if (chunk instanceof ArrayBuffer)
+    return new Uint8Array(chunk)
+  return new Uint8Array(chunk.buffer, chunk.byteOffset, chunk.byteLength)
+}
```
nekomeowww, line 34: Not necessary of having dedicated wrapper.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
```diff
@@ -0,0 +1,240 @@
...
+
+function toWebSocketURL(baseURL: URL | string, token: string): string {
+  const url = new URL(baseURL)
+  url.protocol = url.protocol === 'https:' ? 'wss:' : 'ws:'
+  url.searchParams.set('token', token)
+  return url.toString()
+}
```
nekomeowww, line 41: Not necessary of having dedicated wrapper.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
```diff
@@ -0,0 +1,240 @@
...
+      throw new Error('Official ASR WebSocket is not open.')
+
+    audioReader = audioStream.getReader()
+    while (true) {
+      const { done, value } = await audioReader.read()
+      if (done)
+        break
+      await waitForSocketCapacity()
+      socket.send(toUint8Array(value))
+    }
```
nekomeowww, line 139: Why not using for await pattern for ReadableStream since it's AsyncIterator?

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
```diff
@@ -0,0 +1,240 @@
...
+      socket.addEventListener('message', (event) => {
+        if (typeof event.data !== 'string') {
+          fail(new Error('Official ASR returned an unexpected binary frame.'))
+          return
+        }
+        handleServerMessage(event.data)
+      })
+      socket.addEventListener('error', () => {
+        // The browser exposes the close code and reason on the following event.
+      })
```
nekomeowww, line 222: Why don't we log for error?

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
```diff
@@ -0,0 +1,240 @@
...
+      case 'session.finished':
+        finish()
+        break
+      case 'error':
+        fail(new Error(`${message.code}: ${message.message}`))
+        break
+    }
+  }
+
+  queueMicrotask(() => {
```
nekomeowww, line 186: Why this requires `queueMicrotask`

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
```diff
@@ -0,0 +1,240 @@
...
+        fullStreamController?.enqueue({ type: 'transcript.text.done', delta: '' })
+        break
+      case 'session.finished':
+        finish()
+        break
+      case 'error':
+        fail(new Error(`${message.code}: ${message.message}`))
+        break
+    }
+  }
```
nekomeowww, line 184: Why don't we respect existing Aliyun NLS protocol?

server/apps/api/src/routes/audio-transcription-ws/index.ts
```diff
@@ -0,0 +1,213 @@
...
+
+function toUint8Array(data: unknown): Uint8Array | undefined {
+  if (data instanceof Buffer)
+    return new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
+  if (data instanceof ArrayBuffer)
+    return new Uint8Array(data)
+  if (ArrayBuffer.isView(data))
+    return new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
+  return undefined
+}
```
nekomeowww, line 57: This wrapper has appeared multiple times. Run `alint .` to discover more duplicates.

server/apps/api/src/routes/audio-transcription-ws/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    function closeWithError(code: string, message: string, closeCode: number = 1011) {
+      if (state === 'finished')
+        return
+      state = 'finished'
+      send({ event: 'error', code, message })
+      upstream?.cancel()
+      try {
+        client?.close(closeCode, code)
+      }
+      catch {}
```
nekomeowww, line 83: Why don't we log out errors?

server/apps/api/src/routes/audio-transcription-ws/session.ts
```diff
@@ -0,0 +1,268 @@
...
+    case 'cn-shanghai-internal':
+    case 'cn-beijing-internal':
+    case 'cn-shenzhen-internal':
+      websocketURL.protocol = 'wss:'
+      websocketURL.hostname = `nls-gateway-${region}-internal.aliyuncs.com:80`
+      break
+  }
+
+  return websocketURL
+}
```
nekomeowww, line 96: Wrappers again? Don't we have Aliyun NLS sdk? If applicable, split into shared module/package.

luoling8192:
Updated against current main in 2216eaf12. Official ASR now publishes interim transcript snapshots during speech, authenticates the browser WebSocket without a URL token, and bounds startup time, session time, and audio size. Aliyun token requests use HTTPS and the documented internal WebSocket endpoint. Focused tests (42), workspace typecheck, and lint completed locally. Remaining before merge: define and implement persistent ASR usage/billing accounting; run a real microphone + Aliyun smoke test. CI is running.

luoling8192:
CI follow-up on head 7fc129b37: all checks except Unit Test pass. Unit Test failed twice, both times only in the unchanged controls-island-root.test.ts desktop placement cases at lines 206 and 249 (expected bottom-left, received top-left). The nine unrelated desktop/docs/i18n files are now identical to PR base and absent from this PR diff. Locally, the isolated 10-case file and the 68-file stage-tamagotchi:node project pass (431 tests). This remains an independent CI blocker; no unrelated placement behavior was changed in this ASR PR.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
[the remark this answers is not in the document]
luoling8192: The AIRI token is carried in the `airi-auth.*` WebSocket subprotocol, not the URL or query string; the route authenticates it before opening Aliyun. The ASR test checks the request URL and handshake protocols. No bearer value is logged.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
```diff
@@ -0,0 +1,240 @@
...
+      })
+      socket.addEventListener('error', () => {
+        // The browser exposes the close code and reason on the following event.
+      })
+      socket.addEventListener('close', (event) => {
+        if (settled || sessionFinished)
+          return
+        const reason = event.reason || `closed_${event.code}`
+        fail(new Error(`Official ASR WebSocket closed before session.finished: ${reason}`))
+      })
```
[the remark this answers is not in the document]
luoling8192: Addressed: the client starts a startup timer before opening the socket and clears it on start, failure, or completion. The default is now 45 s, leaving room for the server's 30 s startup deadline. The no-start regression test passes.

server/apps/api/src/app.test.ts
```diff
@@ -75,4 +86,45 @@ describe('business API app', () => {
...
+        server.once('listening', resolve)
+    })
+    const { port } = server.address() as AddressInfo
+
+    try {
+      const unauthorized = new WebSocket(`ws://127.0.0.1:${port}/api/v1/audio/transcriptions/ws`)
+      const unauthorizedCode = await new Promise<number>((resolve) => {
+        unauthorized.once('close', resolve)
+      })
+      expect(unauthorizedCode).toBe(WS_CLOSE_UNAUTHORIZED)
```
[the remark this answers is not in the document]
luoling8192: Addressed in `app.test.ts`: the unauthorized `ws` client has an error listener so the close-code assertion remains the failure signal.

server/apps/api/src/routes/audio-transcription-ws/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    return new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
+  return undefined
+}
+
+/** Builds one authenticated ASR WebSocket handler set per client connection. */
+export function createAudioTranscriptionWsHandlers(options: {
+  configKV: ConfigKVService
+  envelopeCrypto: EnvelopeCrypto
+  providerCatalogService: ProviderCatalogService
+}) {
```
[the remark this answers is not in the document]
luoling8192: Per product decision, this PR now caps each frame at 64 KiB, forwarded audio at 60 s, startup at 30 s, and active sessions at 75 s. Every terminal path writes a structured usage record with user/request ID, forwarded bytes, derived duration, and outcome. ASR pricing and Flux debit are intentionally deferred; logs are not a billing ledger, so I am leaving this thread open for that follow-up.

server/apps/api/src/routes/audio-transcription-ws/session.ts
```diff
@@ -0,0 +1,268 @@
...
+  format: 'pcm',
+  sample_rate: 16000,
+  enable_intermediate_result: true,
+  enable_punctuation_prediction: true,
+  enable_words: true,
+}
+
+function nlsMetaEndpointFromRegion(region: AliyunNlsCredentials['region']): URL {
+  return new URL(`http://nls-meta.${region}.aliyuncs.com`)
+}
```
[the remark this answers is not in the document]
luoling8192: Addressed: the server token endpoint uses HTTPS; the matching direct-client helper was also corrected to HTTPS. A focused endpoint regression test passes.

server/apps/api/src/routes/audio-transcription-ws/session.ts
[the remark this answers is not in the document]
luoling8192: Addressed: internal regions use `ws://nls-gateway-<region>-internal.aliyuncs.com/ws/v1` (port 80 implied). The matching direct-client helper no longer puts a port inside `hostname` or doubles `-internal`; its regression test passes.

server/apps/api/src/routes/audio-transcription-ws/session.ts
```diff
@@ -0,0 +1,268 @@
...
+  const response = await ofetch<{
+    Token?: { ExpireTime?: number, Id?: string }
+    Message?: string
+  }>(`${endpoint}/?Signature=${signature}&${canonicalQuery}`, { method: 'POST' })
+
+  if (typeof response.Token?.Id === 'string' && typeof response.Token?.ExpireTime === 'number')
+    return { token: response.Token.Id, expiresAt: response.Token.ExpireTime * 1000 }
+
+  throw new Error(`Aliyun NLS token request failed: ${response.Message || 'unknown error'}`)
+}
```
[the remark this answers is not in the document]
luoling8192: Addressed: the token request has a 10 s timeout and upstream readiness is bounded. The server route additionally has a 30 s startup deadline; the no-start regression test passes.

server/apps/api/src/routes/audio-transcription-ws/index.ts
```diff
@@ -0,0 +1,213 @@
...
+          onError(error) {
+            log.withError(error).withFields({ userId }).warn('ASR upstream failed')
+            closeWithError('upstream_error', error.message)
+          },
+        })
+        await upstream.start()
+      }
+      catch (error) {
+        log.withError(error).withFields({ userId }).warn('ASR session failed to start')
+        closeWithError('session_start_failed', errorMessageFrom(error) ?? 'The ASR session failed to start.')
```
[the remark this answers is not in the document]
luoling8192: Fixed in 136131072: client errors and server logs now use stable, sanitized error codes/names. The regression test injects a signed upstream URL and verifies that neither its key ID nor signature reaches the client.

server/apps/api/src/app.ts
```diff
@@ -248,15 +248,28 @@ export async function buildApp(deps: AppDeps) {
...
-    db: deps.db,
-    env: deps.env,
+  // Bidirectional ASR proxy. One client connection maps to one Aliyun NLS
+  // session. The browser carries its bearer in the handshake protocol header.
+  const audioTranscriptionWsSetup = createAudioTranscriptionWsHandlers({
     configKV: deps.configKV,
     envelopeCrypto: deps.envelopeCrypto,
     providerCatalogService: deps.providerCatalogService,
+  })
+  app.get('/api/v1/audio/transcriptions/ws', upgradeWebSocket(async (c) => {
```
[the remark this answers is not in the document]
luoling8192: Fixed in 136131072: the WebSocket receiver applies a 64 KiB frame cap before route dispatch. The transport test sends a 64 KiB + 1 frame and verifies close code 1009. The client splits larger PCM chunks into 32 KiB frames.

server/apps/api/src/routes/audio-transcription-ws/session.ts
```diff
@@ -0,0 +1,296 @@
...
+    catch {}
+  }
+
+  function handleMessage(data: RawData) {
+    if (state === 'finished')
+      return
+
+    let event: AliyunNlsServerEvent
+    try {
+      event = JSON.parse(data.toString()) as AliyunNlsServerEvent
```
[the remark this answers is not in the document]
luoling8192: Fixed in 136131072: Aliyun frames are validated with Valibot before transcript fields are read. A numeric transcript now closes the task, while normal `SentenceBegin` events remain accepted; both paths have regression coverage.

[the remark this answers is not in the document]
luoling8192: Fixed in 136131072: Aliyun frames are validated with Valibot before transcript fields are read. A numeric transcript now closes the task, while normal `SentenceBegin` events remain accepted; both paths have regression coverage.

server/apps/api/src/app.ts
```diff
@@ -248,15 +248,28 @@ export async function buildApp(deps: AppDeps) {
...
 
-  // Realtime ASR proxy. Mounted before the global bodyLimit middleware because
-  // the request body is a live microphone PCM stream rather than a bounded JSON
-  // payload. Auth is resolved manually here for the same reason.
-  app.post('/api/v1/audio/transcriptions/stream', createAudioTranscriptionStreamHandler({
-    db: deps.db,
-    env: deps.env,
+  // Bidirectional ASR proxy. One client connection maps to one Aliyun NLS
+  // session. The browser carries its bearer in the handshake protocol header.
+  const audioTranscriptionWsSetup = createAudioTranscriptionWsHandlers({
```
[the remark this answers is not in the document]
luoling8192: Added `server/docs/ai/adr/2026-09-24-official-asr-websocket.md` with the boundary decision, module graph, affected tree, session sequence, limits, non-goals, and test plan.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
```diff
@@ -0,0 +1,278 @@
...
+        fail(new Error('Official ASR WebSocket URL is missing.'))
+        return
+      }
+      if (options.abortSignal?.aborted) {
+        handleAbort()
+        return
+      }
+
+      options.abortSignal?.addEventListener('abort', handleAbort, { once: true })
+      startupTimer = setTimeout(() => fail(new Error('Official ASR did not start in time.')), options.startupTimeoutMs ?? 15000)
```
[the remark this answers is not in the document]
luoling8192: Addressed: the client default startup deadline is 45 s versus the server's 30 s deadline, leaving handshake and delivery margin. The no-start test remains green.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
luoling8192: I kept `resolveAudioStream` because it owns the three-source input precedence (`inputAudioStream`, then `inputStream`, then `file`) and the missing-input error in one place. The generic deferred-promise wrapper was removed in favor of `Promise.withResolvers`.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
luoling8192: Removed the one-use audio-chunk conversion wrapper. The conversion is now next to the send loop, where ArrayBuffer and typed-array views are handled directly.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
luoling8192: The explicit reader is intentional: abort must cancel a pending `read()` immediately. `for await` hides the reader and does not give this path an explicit handle for cancellation while the source is waiting.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
luoling8192: Added a sanitized browser transport warning in 4b3f58d3d. Browser WebSocket `error` events expose no cause; the subsequent `close` event supplies the code/reason that rejects the transcript promise.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
luoling8192: Agreed; there was no required scheduling boundary here. Removed `queueMicrotask`, so setup runs directly after constructing the result streams, with synchronous failures captured into the returned promise/streams.

packages/stage-ui/src/libs/providers/providers/official/stream-transcription.ts
luoling8192: The official AIRI socket is deliberately provider-neutral; the API adapter owns Aliyun task IDs, token, and NLS framing. The server-boundary ADR documents this contract so an upstream change need not alter Stage UI.

server/apps/api/src/routes/audio-transcription-ws/index.ts
luoling8192: I checked the other audio WebSocket route. This conversion is at the Hono `unknown`-data boundary and accepts Buffer, ArrayBuffer, and views without copying; the speech route has a different outbound payload shape. I kept this local boundary rather than exporting a shared utility for two unlike call sites.

server/apps/api/src/routes/audio-transcription-ws/index.ts
luoling8192: Addressed with structured, sanitized server warnings for startup, upstream, forwarding, and terminal failures. We log error name/code and user/request context, but never the upstream error string because it may contain a signed token URL.

server/apps/api/src/routes/audio-transcription-ws/session.ts
luoling8192: The existing Stage UI Aliyun implementation is a browser-side custom session, not a Node SDK shared with the API. The official route also owns AIRI auth, limits, usage, and cleanup; the adapter remains server-local. The new ADR shows the boundary and non-goals.

dasilva333:
Hey @luoling8192! Thanks for the thorough work here—moving the bearer token into the subprotocol and cleaning up the in-order PCM delivery are great improvements.

I wanted to share a friendly architectural observation regarding removing the streaming fetch/SSE path in favor of a bespoke WebSocket per VAD segment:

1. **Standards Compatibility (`/v1/audio/transcriptions` with `stream: true`)**:
   In the broader ecosystem, standardizing on OpenAI-compatible endpoints (`POST /v1/audio/transcriptions` with chunked transfer and `text/event-stream` SSE when `stream: true` is passed) gives client apps universal flexibility. It allows AIRI (and third-party clients) to transparently point to either the official server or any off-the-shelf local ASR runner (e.g. Faster-Whisper, Whisper.cpp, Parakeet/Citrinet engines, or vLLM audio servers) without maintaining bifurcated transport code.

2. **Per-Utterance Handshake Overhead vs HTTP/2 Multiplexing**:
   Spinning up "one WebSocket per VAD speech segment" requires a full TCP + TLS + HTTP 101 Upgrade negotiation for every utterance. On higher-latency mobile connections, that per-segment handshake delay can visibly bite Time-To-First-Transcript (TTFT). With HTTP/2 or HTTP/3 chunked streaming POST + SSE, the client reuses existing keep-alive connections with zero upgrade roundtrips.

3. **Edge Proxy & Connection Lifecycle Simplicity**:
   Stateful WebSockets can be finicky around edge proxies/reverse proxies (Cloudflare, Envoy, Nginx timeouts), and cancellation races (e.g. browser aborting while the server holds an open upstream socket) require explicit state machine guards (as recently addressed in #2645). A standard HTTP stream naturally closes its reader and aborts the upstream request cleanly via standard `AbortController` / response stream cancellation.

Full-duplex WebSockets are clearly indispensable for bidirectional foundation models (like Gemini Live Bidi or OpenAI Realtime). But for traditional unidirectional speech-to-text, keeping or formalizing the HTTP SSE path (`stream: true`) alongside (or as the primary) transport might save the project from being locked into a proprietary WebSocket protocol.

Just food for thought as you refine the backend gates and billing decisions! Great work on tightening the audio pipeline.

PR #2307, opened 2026-08-17
fix(api): load Drizzle migrations at runtime

luoling8192:
## Summary

- Use Drizzle's Node PostgreSQL migrator to read the checked-in migration journal and SQL files at API startup.
- Remove the virtual migration module, its plugin, and the browser migrator.
- Simplify both API Dockerfiles and Railway watch paths.

## Verification

- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm typecheck`
- `pnpm lint` (0 errors; 7 existing warnings outside this change)
- Official migrator boundary check: PostgreSQL function bodies and string literals retain embedded semicolons.
- `docker build --no-cache -f server/apps/api/production/railway/Dockerfile -t airi-api-drizzle-node-migrator:local .`

## Visual changes

Not applicable. This PR only changes API migration loading and deployment configuration.


luoling8192:
Resolved the SQL statement-boundary concern in 6bc049560. The API now uses Drizzle's official Node PostgreSQL migrator instead of the copied browser migration loader. Its migration reader splits on `--> statement-breakpoint`; a temporary fixture verified that PostgreSQL function bodies and string literals retain embedded semicolons.

PR #2348, opened 2026-08-23
fix(stage-ui): provider config lost after onboarding, causing OpenRouter 401s

Antonbrows:
## Summary

Fixes OpenRouter (and other providers) failing to actually work after completing onboarding, even when validation appears to succeed and the API key is correct. Chat requests end up failing with a `401` auth error from the provider.

Two stacked bugs, both in the onboarding path:

1. **`extractStatusCode()` couldn't read xsai's error shape.** It only checked `error.cause.status` / `error.cause.statusCode` / `error.cause.response.status`, but `@xsai/shared`'s `APICallError` (thrown by `generateText` on a non-ok response) sets `statusCode` and `response` directly on the error instance — `error.cause` is unrelated. This meant the deliberate "a 400 on the validation ping means auth/connectivity are fine, the probe request was just malformed" heuristic in the chat-completions validator never fired. Any provider that rejects the lightweight 1-token validation ping with a 400 — for example OpenRouter enforcing a higher `max_tokens` minimum on certain models — failed the whole validation check even though the key and provider were working correctly.

2. **Provider config entered during onboarding was never actually persisted.** `saveProviderConfiguration()` in `onboarding.vue` wrote through `providerConfigStore.configs`, which is a read-only `computed` projection of the store's real `providers` state (`Object.fromEntries(...)` over the source ref). Mutating the object returned by that computed never reaches the underlying `useLocalStorage` ref, so the write was silently discarded. Routing the write through the store's own `ensureProvider`/`updateProviderConfig` actions doesn't fix it either — both are listed under this store's `synced.actions`, which routes them through `pinia-plugin-synced`'s leader-election RPC. Even a same-window call lands asynchronously with no promise the caller can await to know the write landed before reading the value back (confirmed via debug logging: `updateProviderConfig`'s `if (!provider) return` guard fires because the preceding `ensureProvider` call hadn't landed yet).

Net effect: onboarding *looked* successful — validation passed using the in-memory form data — but nothing was actually saved. The app was left with a provider marked active and no persisted credentials, so real chat requests went out with a missing/stale key and OpenRouter returned `401 User not found`.

## Fix

- `packages/stage-ui/src/libs/providers/validators/openai-compatible.ts`: `extractStatusCode()` now checks `error.statusCode` / `error.status` / `error.response?.status` directly, before falling back to the `.cause`-nested variants.
- `packages/stage-ui/src/components/scenarios/dialogs/onboarding/onboarding.vue`: `saveProviderConfiguration()` now writes straight to the store's raw `providers`/`addedProviders` refs — a plain synchronous local mutation, avoiding the async synced-action indirection — so the config is guaranteed to be in place before `loadModelsForProvider` reads it.

## Test plan

- [x] `pnpm -F @proj-airi/stage-ui exec vitest run src/libs/providers/validators/openai-compatible.test.ts` — 7/7 pass, including a new regression test reproducing the real `APICallError` shape.
- [x] `pnpm lint` on the changed files — clean.
- [x] Manually verified end-to-end in a real browser with a live (now-revoked) OpenRouter key:
  - Reproduced the original failure: onboarding validation failed with "Continue Anyway"/"Retry" before the validator fix.
  - After the validator fix, validation passed and reached model selection.
  - Before the persistence fix, `localStorage['settings/providers/configured']` stayed `{}` after completing onboarding with a working key — confirmed via direct inspection.
  - After the persistence fix, the entered API key correctly landed in `localStorage`.
  - Sent a real chat message through the app; the request went out with the correct model/key and got back a `200` with an actual assistant reply.

## Root cause verification

For the validator bug, confirmed `@xsai/shared`'s `APICallError` shape directly from `node_modules`:

```js
class APICallError extends XSAIError {
  constructor(message, options) {
    super(message, "api_call_error", options);
    ...
    this.statusCode = options.response.status;   // <- not under `.cause`
    ...
  }
}
```

For the persistence bug, added temporary debug logging to `saveProviderConfiguration()` and confirmed that immediately after `providersStore.initializeProvider(providerId)`, `providerStore.getProviderConfig(providerId)` still returned `undefined` — proving the synced-action write hadn't landed synchronously, and that `updateProviderConfig`'s existence guard was silently no-opping as a result.

### DIFF
--- a/packages/stage-ui/src/components/scenarios/dialogs/onboarding/onboarding.vue
+++ b/packages/stage-ui/src/components/scenarios/dialogs/onboarding/onboarding.vue
@@ -42,7 +42,7 @@ const { trackOnboardingCompleted, trackOnboardingStarted, trackOnboardingStepCom
 const providersStore = useProviderStore()
 
 const providerStore = useProviderConfigStore()
-const { configs: providers } = storeToRefs(providerStore)
+const { providers, addedProviders } = storeToRefs(providerStore)
 const { allChatProvidersMetadata } = storeToRefs(providersStore)
 const consciousnessStore = useConsciousnessStore()
 const {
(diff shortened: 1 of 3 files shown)


nekomeowww:
This should be fixed already.

PR #2349, opened 2026-08-23
fix(stage-pages): make Web Speech API selectable as a transcription provider

Antonbrows:
## Summary

Web Speech API (Browser) could never be selected as a transcription provider in Settings > Hearing, even though its settings/test page implied it was ready to use.

Settings > Hearing filters selectable providers by `status === 'configured'` (`configuredTranscriptionProvidersMetadata` in `stores/providers/provider.ts`). Every other transcription provider page transitions that status via `useProviderValidation`/`providersStore.validateProvider` on successful validation, or via an explicit "Continue Anyway" bypass. `browser-web-speech-api.vue` never did either — it only let you tweak local settings (language/continuous/interim results) and run a local demo test, with no path that ever set `status`. So no matter how the page looked, the provider's status stayed `unconfigured` forever, and it could never appear in the Hearing module's provider list.

On top of that, this page's settings writes went through `storeToRefs(providerStore).configs` — a read-only `computed` projection of the store's real `providers` state, not the state itself. Mutating the object returned by that computed never reaches the underlying `useLocalStorage` ref.

I also confirmed the store's own `ensureProvider`/`markProviderAdded`/`forceProviderConfigured` actions aren't a reliable fix here either: they're listed under this store's `synced.actions`, which routes them through `pinia-plugin-synced`'s leader-election RPC. In a single-tab session I watched that RPC go unresolved indefinitely — the provider entry never landed even after 10+ seconds and user interaction.

## Fix

`packages/stage-pages/src/pages/settings/providers/transcription/browser-web-speech-api.vue`:
- Settings (language/continuous/interimResults) now write directly to the store's raw `providers` ref instead of the read-only `configs` computed.
- Added an explicit **"Enable Web Speech API"** button that sets `status: 'configured'` and marks the provider added — both via a plain synchronous mutation of the raw `providers`/`addedProviders` refs, bypassing the unreliable synced-action RPC layer.
- Added a status alert (with a "Disable" action) so it's clear when the provider is actually active and selectable elsewhere.

## Test plan

- [x] `pnpm lint` on the changed file — clean.
- [x] `vue-tsc --noEmit` on `packages/stage-pages` — no new errors.
- [x] Manually verified end-to-end in a real browser:
  - Before the fix: visited the page, entered settings, ran the test successfully — `localStorage['settings/providers/configured']` stayed `{}` regardless, and the provider never appeared on Settings > Hearing.
  - After the fix: clicking "Enable Web Speech API" immediately persists `status: 'configured'` to `localStorage`, the provider appears as a selectable radio card on Settings > Hearing, and selecting it correctly sets it as the active hearing provider with its model populated.

### DIFF
--- a/packages/i18n/src/locales/en/settings.yaml
+++ b/packages/i18n/src/locales/en/settings.yaml
@@ -1459,6 +1459,9 @@ pages:
       browser-web-speech-api:
         description: Browser-native STT (requires Chrome/Edge/Safari)
         title: Web Speech API
+        disable: Disable
+        enable: Enable Web Speech API
+        enabled: Enabled — selectable in the Hearing module
       official:
         title: Official Provider
         description: Official AI provider by AIRI.
--- a/packages/stage-pages/src/pages/settings/providers/transcription/browser-web-speech-api.vue
+++ b/packages/stage-pages/src/pages/settings/providers/transcription/browser-web-speech-api.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import type { RemovableRef } from '@vueuse/core'
+
 import { errorMessageFromValue } from '@proj-airi/stage-shared'
 import {
   Alert,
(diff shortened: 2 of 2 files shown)


nekomeowww:
This should be fixed already.

PR #2350, opened 2026-08-24
fix(stage-ui): speech provider config never persisted or became selectable

Antonbrows:
## Summary

No speech (text-to-speech) provider could ever be selected in Settings > Speech, and any config you entered for one (voice settings especially) could vanish after navigating away. Reported specifically against OpenRouter's speech provider, but the root cause is shared by every provider built on `speech-provider-settings.vue` — ElevenLabs, OpenAI, Google Gemini, Microsoft, Deepgram, Alibaba, Volcengine, MiniMax.

This ended up being three separate, stacked bugs, each only reachable once the one before it was fixed:

**1. Synced actions calling each other without awaiting (`provider.ts`)**

`initializeProvider`/`forceProviderConfigured` call `providerConfigStore`'s synced actions (`ensureProvider`, `setProviderStatus`, `markProviderAdded`) without awaiting them. Each synced store runs its own leader-election `Tab` instance (`pinia-plugin-synced`), so the outer call could resolve before the nested call's RPC round-trip landed — silently dropping the effect, even when the *outer* wrapper itself was awaited by its caller. `stores/modules/default.ts` already `await`s both functions, expecting them to complete, so this was already an implicit contract violation.

Confirmed live via the running Pinia instance: calling `providerConfigStore`'s actions directly and awaiting them lands in a few ms; calling them unawaited from inside another synced action never lands, regardless of whether the outer call is awaited.

**2. Synced actions returning non-cloneable values (`config.ts`)**

Fixing #1 meant these calls actually *complete* now — which surfaced a second, previously-unreachable bug: `ensureProvider`/`fetchProviders`/`updateProviderConfig` could return a live Vue reactive Proxy read straight off the `providers` ref. Synced-action return values get sent back through `postMessage` across the leader/follower `BroadcastChannel`, and a Proxy throws `DataCloneError: could not be cloned` there. This reproduced as an intermittent crash (raced on which of two concurrent `ensureProvider` calls hit the "already exists" branch) — reliably reproducible on a fresh page load before the fix, not reproducible in repeated testing after it. Fixed with `toRaw()` at the return sites, per `AGENTS.md`'s own documented contract ("action results must support structuredClone").

**3. `speech-provider-settings.vue` never validated or persisted correctly**

- Nothing in this component ever transitioned the provider's status to `configured` — no auto-validation, no manual "Continue Anyway" fallback. Settings > Speech only lists providers with `status === 'configured'`, so none could ever appear there no matter what credentials were entered. Fixed by wiring in `useProviderValidation`, the same composable chat/vision provider pages already use.
- `debouncedUpdate` wrote via `providers.value[id] = {...spread, apiKey, baseUrl, voiceSettings}` — a full reassignment through this store's `configs` projection, which only persists *nested* mutations of an existing entry. Assigning a whole new object just replaces a property on the projection's own read-only, recomputed container, not the store's real state — a silent no-op. `apiKey`/`baseUrl` already had separate, correctly-persisting nested-mutation setters, so credentials happened to survive independently, but `voiceSettings` had no other write path and was lost on every reload. Fixed by switching to `Object.assign` onto the existing entry.

**Bonus: unguarded `.trim()` crash surfaced by fixing #1 and #2**

`createProvider(config)` also runs for model/voice listing against a freshly-created, still-credential-less config (right after `ensureProvider`, before the user types a key). `config.apiKey.trim()` crashed on `undefined` in that case — previously masked because the config entry never actually existed when this code ran. Affects OpenRouter, ElevenLabs, Google Gemini, MiniMax, and the unspeech-based Deepgram/Microsoft/Alibaba/Volcengine providers. Guarded with `(config.apiKey ?? '').trim()`, matching the pattern already used elsewhere (e.g. `azure-openai`, `cloudflare-workers-ai`).

## Commits

1. `fix(stage-ui): make synced provider-store actions properly async` — the two `provider.ts`/`config.ts` root-cause fixes.
2. `fix(stage-ui): make speech provider settings actually persist and validate` — the actual user-facing fix.
3. `fix(stage-ui): guard speech providers against a not-yet-configured apiKey` — the surfaced `.trim()` crashes.

## Test plan

- [x] `pnpm lint` on all changed files — clean.
- [x] `vue-tsc --noEmit` on `packages/stage-ui` — no new errors.
- [x] `pnpm -F @proj-airi/stage-ui exec vitest run` — 691/691 tests pass (including tests that previously broke when `initializeProvider` stopped being synchronous, fixed by keeping runtime-state init synchronous and only awaiting the config-store call).
- [x] Manually verified end-to-end against `openrouter-audio-speech` on a clean dev server, repeated across multiple full server restarts and 15+ fresh page loads:
  - Before: the settings page either silently failed to persist anything, or crashed on load with a `DataCloneError`.
  - After: entering an API key transitions the provider to `configured`, it becomes selectable on Settings > Speech, and the full config — including `voiceSettings` — survives navigating away and back.
  - 6/6 clean loads on a fully fresh dev server restart with no crashes (vs. 3/3 reliably crashing before the `config.ts` fix).

### DIFF
--- a/packages/stage-ui/src/components/scenarios/providers/speech-provider-settings.vue
+++ b/packages/stage-ui/src/components/scenarios/providers/speech-provider-settings.vue
@@ -2,8 +2,6 @@
 import { useDebounceFn } from '@vueuse/core'
 import { storeToRefs } from 'pinia'
 import { computed, onMounted, ref, watch } from 'vue'
-import { useI18n } from 'vue-i18n'
-import { useRouter } from 'vue-router'
 
 import ProviderSettingsLayout from './provider-settings-layout.vue'
 
@@ -13,11 +11,11 @@ import {
   ProviderBaseUrlInput,
   ProviderBasicSettings,
   ProviderSettingsContainer,
+  ProviderValidationAlerts,
 } from '.'
-import { selectProviderMetadata } from '../../../libs/providers/metadata'
+import { useProviderValidation } from '../../../composables/use-provider-validation'
 import { useSpeechStore } from '../../../stores/modules/speech'
 import { useProviderConfigStore } from '../../../stores/providers/config'
-import { useProviderStore } from '../../../stores/providers/provider'
 
 const props = defineProps<{
   providerId: string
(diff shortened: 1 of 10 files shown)


Antonbrows:
Withdrawing — the speech-provider fix surfaced more issues than expected (including an intermittent crash) and I'm going to use a different provider instead of continuing to debug this one.

PR #2356, opened 2026-08-24
ci: upgrade actions/setup-java to v6

brunoborges:
## Description

Upgrade actions/setup-java from v4 to v6 in the active Pocket Android release workflow. This adopts the current Node.js 24 action runtime without changing the configured Java distribution or version.

## Linked Issues

None.

## Additional Context

Validated with `git diff --check` and confirmed no pre-v6 actions/setup-java references remain in active workflows.

### DIFF
--- a/.github/workflows/release-pocket-android.yml
+++ b/.github/workflows/release-pocket-android.yml
@@ -33,7 +33,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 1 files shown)


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #2373, opened 2026-08-26
refactor(stage-ui): use public package exports

leaft:
## Description

`@proj-airi/stage-ui` bypassed two public package boundaries:

- `Stage.vue` imported `useSettingsLive2d` from the `stage-ui-live2d` source tree.
- The MCP store imported `useLocalStorageManualReset` from the `stage-shared` source tree.

This PR replaces both source-tree imports with their existing public package entries:

- `@proj-airi/stage-ui-live2d`
- `@proj-airi/stage-shared/composables`

This keeps package consumers independent from another package's internal file layout.

## Linked Issues

None.

## Additional Context

There is no runtime or visual behavior change.

Verification:

- `sem diff --staged --no-cosmetics -v --file-exts .ts .tsx`
- `sem diff --staged --no-cosmetics -v --file-exts .vue`
- `pnpm lint` — 0 errors and 7 existing warnings outside this change
- `git diff --check`

The full workspace `pnpm typecheck` was also run. In the isolated worktree it stopped in unchanged `electron-vueuse` because the generated `electron-eventa` output was not present. The affected package type check above passed.


nekomeowww:
Thanks! I miss you so much, thanks for refactoring so many PRs.

PR #2402, opened 2026-08-29
fix(settings): unblock navigation from settings index

DDomelette:
## Summary

- Remove settings-index route guards that waited for a callback with no caller.
- Let the shared settings page use the existing page-specific transition flow.
- Add a browser regression test for navigation with page-specific transitions enabled.

## Verification

- `pnpm -F @proj-airi/stage-web exec vitest run --config vitest.config.ts src/pages/settings-navigation.browser.test.ts --project browser`
- `pnpm typecheck`
- `pnpm lint` (passes with existing repository warnings)
- `pnpm -F @proj-airi/stage-tamagotchi build`

## Visual changes

The same action opens **System** from the settings index. Before this change, the URL changed but the settings index remained visible. After this change, navigation completes.

| Surface | Before | After |
| --- | --- | --- |
| Web | Web settings index remains visible | Web System settings page opens |
| Electron | Electron settings index remains visible | Electron System settings page opens |

### DIFF
--- a/apps/stage-web/vite.config.ts
+++ b/apps/stage-web/vite.config.ts
@@ -235,7 +235,7 @@ export default defineConfig({
     }),
 
     // https://github.com/webfansplz/vite-plugin-vue-devtools
-    VueDevTools(),
+    ...(env.NODE_ENV === 'test' ? [] : [VueDevTools()]),
 
     DownloadLive2DSDK(),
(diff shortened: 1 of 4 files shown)


apps/stage-web/vite.config.ts
```diff
@@ -235,7 +235,7 @@ export default defineConfig({
     }),
 
     // https://github.com/webfansplz/vite-plugin-vue-devtools
-    VueDevTools(),
+    ...(env.NODE_ENV === 'test' ? [] : [VueDevTools()]),
```
nekomeowww, line 238: Should not be modified.

nekomeowww:
What is the purpose of this change?

PR #2408, opened 2026-08-30
fix(better-ws): publish the package so server-sdk installs

### LINKED ISSUE #2359: bug(server-sdk): published package cannot install because better-ws is missing from npm
### Describe the bug

`@proj-airi/server-sdk@0.12.0-beta.1` cannot be installed in a clean external project from the public npm registry.

The published SDK declares `@proj-airi/better-ws@^0.12.0-beta.1` as a runtime dependency. That package is not available from the registry.

### Steps to reproduce

```bash
mkdir airi-server-sdk-repro
cd airi-server-sdk-repro
pnpm init
pnpm add @proj-airi/server-sdk@0.12.0-beta.1 \
  --registry=https://registry.npmjs.org
```

### Actual result

```text
[ERR_PNPM_FETCH_404] GET https://registry.npmjs.org/@proj-airi%2Fbetter-ws: Not Found - 404

This error happened while installing the dependencies of @proj-airi/server-sdk@0.12.0-beta.1

@proj-airi/better-ws is not in the npm registry, or you have no permission to fetch it.
```

`@proj-airi/server-sdk@0.11.3` fails for the same reason.

### Expected result

A clean external project can install the published Server SDK with the command from its README.

### Additional context

- AIRI main revision: `d5742dbb7a44fbb24e1887230c9f21328261efe4`
- `packages/better-ws/package.json` contains `"private": true`.
- Discussion #2358 contains the original report. This Issue replaces that Discussion.

### Possible fixes

The correct fix depends on the intended package design:

1. Publish `@proj-airi/better-ws`.
2. Bundle `better-ws` into `@proj-airi/server-sdk`.
3. Remove the private runtime dependency, or replace it with a public dependency.
4. Keep `server-sdk` internal. Document another supported external integration path.

After maintainers confirm the package design, I can prepare a focused PR.

I can also add a release regression test. The test will pack the related packages and install the Server SDK in a clean project.

### System Info

The auto-generated host system information below is not the environment used for the reproduction.

Host shell:

- OS: macOS 26.5.1, arm64 Apple M2 Pro
- Node: 20.11.0
- pnpm: 8.14.1

The installation failure was reproduced in a clean environment with:

- Node: 24.19.0
- pnpm: 11.19.0
- Registry: https://registry.npmjs.org

The failure is a registry 404 for `@proj-airi/better-ws`, so it is independent of the host shell's Node and pnpm versions.

### Validations

### Contributions

Jorge-Polanco-Roque:
Closes #2359.

`@proj-airi/server-sdk` cannot be installed from npm by external projects:

```
[ERR_PNPM_FETCH_404] GET https://registry.npmjs.org/@proj-airi%2Fbetter-ws: Not Found - 404
```

**Root cause:** `packages/better-ws/package.json` has `"private": true`, so the release workflow's `pnpm publish -r --access public` skips it. But `server-sdk` (published) declares `@proj-airi/better-ws` as a runtime **dependency**, so the published SDK references a package that never made it to the registry.

`better-ws` is otherwise fully configured for publishing (`files`, `exports`, `dist` build), and its published siblings `@proj-airi/server-sdk` and `@proj-airi/server-shared` are **not** `private`. Removing the flag lets it publish alongside them, so the next release fixes external installs. No changeset is needed (releases use `pnpm publish -r` directly).

### DIFF
--- a/packages/better-ws/package.json
+++ b/packages/better-ws/package.json
@@ -2,7 +2,6 @@
   "name": "@proj-airi/better-ws",
   "type": "module",
   "version": "0.12.0-beta.5",
-  "private": true,
   "description": "Transport-agnostic reliable WebSocket runtime primitives",
   "author": {
     "name": "Moeru AI Project AIRI Team",


Jorge-Polanco-Roque:
Good catch — you're right that this line alone isn't the whole story.

Removing `private: true` is a necessary prerequisite: `pnpm publish -r` skips private packages, so while `better-ws` stays private it can never reach the registry and `@proj-airi/server-sdk` (which declares it as a runtime dependency) keeps failing to install.

But as you note, it isn't sufficient by itself. The release workflow publishes via OIDC trusted publishing (`id-token: write`, no npm token), which npm only accepts once a trusted publisher is configured for an **existing** package. Since `better-ws` has never been published, its first publish needs a one-time bootstrap with authenticated credentials plus configuring the trusted publisher on npm — the same bootstrap `@proj-airi/server-sdk` and `@proj-airi/server-shared` already went through. That's a maintainer/ops step (it needs `@proj-airi` org credentials), so it can't live in this PR.

Happy to add a note flagging the bootstrap, or to sequence it however you prefer. This change is the code-side prerequisite; the SDK stays broken until `better-ws` is both un-private'd and bootstrapped on npm.

PR #2411, opened 2026-08-30
feat(stage-ui): add named custom model connections

JackSmith111977:
## Summary

Add a local Custom Model connection. A user can name the connection, choose OpenAI Chat Completions, OpenAI Responses, or Anthropic Messages, and keep secrets on this device.

The desktop app sends requests from the Main Process. The web app uses browser fetch and shows a CORS limit.

### Follow-up

- Custom Model is on chat provider settings. Welcome, onboarding, Speech, Transcript, and Artistry still need the same entry.
- A Custom Model name can collide with OpenAI Compatible or a hand-typed model id.

## How to test

```bash
pnpm --config.verify-deps-before-run=false typecheck
pnpm exec vitest run packages/core-agent/src/runtime/model-runtime apps/stage-tamagotchi/src/main/services/electron/custom-model-fetch.test.ts apps/stage-tamagotchi/src/shared/eventa/custom-model-fetch.test.ts
pnpm exec vitest run --config packages/stage-ui/vitest.config.ts --project node packages/stage-ui/src/libs/providers/custom-model packages/stage-ui/src/composables/use-custom-model-editor.test.ts packages/stage-ui/src/stores/providers/config.test.ts packages/stage-ui/src/stores/chat.contract.test.ts
```

1. Open Settings, then Providers, then Chat.
2. Open Custom Model.
3. Set a connection name and a Base URL.
4. Discover models, then send a chat message.

## Visual changes

| Before | After |
|---|---|
|  |  |
| Settings / Providers / Chat catalog | Settings / Providers / Chat catalog |

| Before | After |
|---|---|
| _absent_ |  |
| Custom Model editor | Custom Model editor |


nekomeowww:
Thanks for submitting this PR. However, I think that having so many lines of changes to add such feature like this is not necessary. Please resubmit with closer look.

JackSmith111977:
Thanks for the review.                                                                                                                                                                                                                                                                                                                                                                                                                                                 I have used AIRI for some time, and adding a custom model still feels awkward. That is why I opened this PR. I wanted a named custom provider, similar to CC Switch. I think that would improve the product.                                                                                                                                                                                                                                I can polish this and resubmit a smaller change. Would you like me to do that?

JackSmith111977:
Thanks for taking a look.

I have used AIRI for a while. Adding a custom model is still hard to get right, so I built this. The idea is close to CC Switch: a named connection, an explicit protocol, an endpoint, auth, and more than one instance. I think that would make custom providers much easier to use.

The change set is large because this is not only a settings form. The current OpenAI-compatible path cannot express Anthropic Messages or OpenAI Responses, and the web app cannot reach many private gateways because of CORS. This PR adds:

- a protocol-neutral runtime for Chat Completions, Responses, and Anthropic Messages
- Electron Main Process transport, so desktop is not blocked by CORS
- a dedicated editor for protocol, paths, headers, discovery, and a generation test
- tests for those boundaries

I can take another pass, tighten the scope, and resubmit. Is that what you would like me to do?

PR #2412, opened 2026-08-30
fix(stage-tamagotchi): keep expanded controls island reachable on small windows

### LINKED ISSUE #2400: bug(stage-tamagotchi): expanded controls island menu is clipped in small stage areas
### Describe the bug

https://github.com/user-attachments/assets/622187fe-9bdf-41e2-844d-b4d7b97f1e21

If the monitor is small, I would adjust the UI to be smaller, but in this case, it gets cut off. There is no way to make any further adjustments after that.

This issue is occurring in both the development environment and the production deployment.

### System Info

```Shell
System:
    OS: macOS 15.6.1
    CPU: (8) arm64 Apple M1
    Memory: 75.73 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.18.1 - /Users/nayunsang/.nvm/versions/node/v24.18.1/bin/node
    npm: 11.16.0 - /Users/nayunsang/.nvm/versions/node/v24.18.1/bin/npm
    pnpm: 11.24.0 - /Users/nayunsang/.nvm/versions/node/v24.18.1/bin/pnpm
    bun: 1.2.13 - /Users/nayunsang/.bun/bin/bun
  Browsers:
    Chrome: 152.0.7977.64
    Firefox: 137.0.2
    Safari: 18.6
```

### Validations

### Contributions

Jorge-Polanco-Roque:
Fixes #2400.

On a small window, expanding the controls island grows the drawer (auth button + the grid of actions) with no height bound, so it overflows the window and its controls get clipped out of reach — and since nothing scrolls, there's no way to reach them (the reporter's "no way to make any further adjustments").

**Fix:** cap the expanded drawer to the viewport height and let it scroll internally, so every control stays reachable at any window size and dock corner:

```
max-h-[calc(100dvh - 5rem)] overflow-y-auto overscroll-contain
```

I bound the **drawer** rather than the whole island on purpose, so the expand toggle and the main controls docked next to it stay fixed and visible; only the drawer scrolls.

**Heads-up / where I'd love your call:**
- I couldn't reproduce visually end-to-end locally (needs the desktop app on a small window), so this is my best guess — happy to adjust.
- The `5rem` reserve (room for the main controls next to the drawer) is an estimate; easy to tune.
- If you'd rather solve it by **shrinking the icons** on small windows instead of scrolling, the plumbing already exists (`controlsIslandIconSize` auto/small/large, currently forced to `large`) — I can switch `auto` to size by window height instead. Let me know which UX you prefer.

### DIFF
--- a/apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
+++ b/apps/stage-tamagotchi/src/renderer/components/stage-islands/controls-island/index.vue
@@ -4,7 +4,7 @@ import { useElectronEventaContext, useElectronEventaInvoke, useElectronMouseInEl
 import { IS_DEV } from '@proj-airi/stage-shared'
 import { useSettings, useSettingsAudioDevice } from '@proj-airi/stage-ui/stores/settings'
 import { useTheme } from '@proj-airi/ui'
-import { refDebounced, useIntervalFn } from '@vueuse/core'
+import { refDebounced, useIntervalFn, useWindowSize } from '@vueuse/core'
 import { storeToRefs } from 'pinia'
 import { computed, reactive, ref, useTemplateRef, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
@@ -40,6 +40,12 @@ const emit = defineEmits<Emits>()
 const { isDark, toggleDark } = useTheme()
 const { t } = useI18n()
 const { dock, isLeft, isTop, motionPhase } = useControlsIslandPlacement()
+const { height: windowHeight } = useWindowSize()
+
+// Below this window height, `auto` icon sizing switches to the smaller icons so
+// the controls column (and the expanded drawer next to it) still fit; above it,
+// the larger, more legible icons are used.
+const AUTO_LARGE_MIN_HEIGHT = 420
 
 const settingsAudioDeviceStore = useSettingsAudioDevice()
 const settingsStore = useSettings()
@@ -128,9 +134,9 @@ const adjustStyleClasses = computed(() => {
       break
     case 'auto':
     default:
-      // Fixed to large for better visibility in the new layout,
-      // can be changed to windowHeight based check if absolutely needed.
-      isLarge = true
+      // Use the more legible large icons when there's vertical room, and shrink
+      // to the smaller ones on short windows so the controls stay reachable.
+      isLarge = windowHeight.value >= AUTO_LARGE_MIN_HEIGHT
       break
   }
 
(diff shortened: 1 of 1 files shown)


Jorge-Polanco-Roque:
Good catch — you're right, the fixed 5rem reserve underestimates the main-controls column (~5 buttons ≈ 12rem). Reworked it to size from the actual remaining flex space instead of a magic reserve: the island is bounded to the viewport (`max-h-[calc(100dvh - 1rem)]`), the main-controls column keeps its natural height (`shrink-0`), and the drawer shrinks (`min-h-0`) and scrolls within whatever space is left. So at any short viewport the combined island stays within bounds and every control is reachable, regardless of button count/size.

Jorge-Polanco-Roque:
Right again — at the 200px minimum the large-icon controls column alone (~216px) exceeds the viewport, so scrolling the drawer can't help; the controls themselves need to shrink. I've made the `auto` icon size switch to the smaller icons below a window-height threshold (the `windowHeight`-based check the code already anticipated), so the whole controls column fits at the minimum size while the flex cap + drawer scroll keep everything reachable in between. Threshold (420px) is a first estimate and easy to tune to your preference.

nayounsang:
Hi, I’m the person who opened this issue. Although I previously said there was no way to verify it, the issue can be consistently reproduced by reducing the avatar size. 
Could you please provide a demonstration of this behavior? I can help with the review.

Jorge-Polanco-Roque:
Thanks @nayounsang, and good to know it reproduces reliably by reducing the avatar size (which shrinks the stage/window area the island lives in).

Quick recap of what the current fix does: on a small window the expanded drawer (auth button + action grid) grew unbounded and overflowed with nothing scrollable, so the lower controls became unreachable. The fix caps the expanded drawer to the available viewport height with internal scroll (`overflow-y-auto overscroll-contain`), and sizes the action column from the actual remaining space so the controls shrink rather than clip when space is tight — the expand toggle and the main docked controls stay fixed, only the drawer scrolls. So every control stays reachable at any window size / dock corner.

On the demo: I develop without the desktop app on a small window, so I've been reasoning from the layout rather than a recording (noted in the PR). Since you can reproduce it reliably — would you be up for checking out this branch and confirming the drawer now scrolls and every control is reachable with a small avatar? That'd be the most reliable verification, and I'd really appreciate it. Happy to record a short screen capture too if that's easier.

nayounsang:
@Jorge-Polanco-Roque It is convenient to receive videos or GIFs.

Issue #2255, opened 2026-08-11
Support separate display text and TTS pronunciation text


a951753abc:
## Clear and concise description of the problem

As a developer using AIRI with Japanese TTS, I want assistant messages to retain natural Japanese orthography in chat while allowing the speech pipeline to receive an explicit pronunciation form, so that ambiguous kanji, names, and domain-specific terms are pronounced correctly.

Today the same streamed literal is used for both the visible assistant message and TTS. Japanese TTS backends may infer the wrong reading for kanji or proper nouns. Writing the whole response in kana avoids the pronunciation error but makes the chat transcript less natural and less readable.

## Suggested solution / Ideas

Introduce separate display and speech projections for assistant text. One possible input convention is silent ruby metadata:

```text
Raw model output: ｜約束《やくそく》は｜約束《やくそく》です。
Display text:     約束は約束です。
Speech text:      やくそくはやくそくです。
```

The exact syntax is flexible; the important contract is that:

- display text preserves the original kanji;
- TTS receives the explicit reading;
- pronunciation metadata is never displayed, persisted as conversational content, or pronounced;
- parsing remains safe across arbitrary streaming chunk boundaries;
- replay/read-aloud uses the resolved speech text or cached audio;
- unannotated text continues to work unchanged.

A general `displayText` / `speechText` projection in the core-agent-to-speech boundary may be more reusable than implementing this in individual TTS providers.

## Alternative

- Ask the model to output all Japanese in kana: corrects some readings but degrades transcript readability.
- Perform automatic G2P in each TTS adapter: adds backend-specific behavior and still struggles with names and fictional/domain terminology.
- Make a second LLM call to generate kana: increases latency, cost, and synchronization complexity.

## Additional context

This is especially useful for Japanese, but the display/speech split could also support abbreviations, symbols, mathematical notation, and pronunciation overrides in other languages.

Validations completed:

- Followed the Code of Conduct and read the contributing guidance.
- Searched existing open and closed issues for TTS pronunciation, ruby/furigana, and separate display/speech text; no duplicate was found.

PR #2413, opened 2026-08-30
feat(stage-ui): support separate display text and TTS pronunciation via ruby annotations

Jorge-Polanco-Roque:
### Problem

For languages like Japanese, the text you want to **show** and the text you want to **pronounce** differ: the chat should keep the kanji (`約束`), but TTS needs the reading (`やくそく`) to speak it correctly. Today the assistant's raw output feeds both the chat and TTS verbatim, so there's no way to control pronunciation without corrupting the display (and vice-versa).

Closes the design gap described in #2255. As the reporter suggested, a general `displayText` / `speechText` projection at the *core agent-to-speech boundary* is more reusable than doing it per TTS provider — that's the approach here.

### Approach

A small, self-contained streaming projector (`libs/speech/ruby-annotation.ts`) parses Aozora/pixiv-style ruby markup and splits one token stream into two:

```
｜約束《やくそく》は   →   displayText: 約束は     speechText: やくそくは
```

- **`displayText`** keeps the base (markers stripped) for the chat surface.
- **`speechText`** substitutes the reading for TTS.
- The markers themselves are never displayed nor spoken.
- Text without annotations passes through unchanged (`displayText === speechText`) — a no-op for the common case.

**Streaming-safe by construction.** The projector is stateful, so an annotation may be split across *arbitrary* chunk boundaries (`｜約` + `束《やく` + `そく》は`): partial annotations are buffered and only committed once the closing `》` arrives. `flush()` emits any unterminated buffer verbatim, so no text is ever dropped even on malformed input.

### Wiring (speech side)

`Stage.vue` projects assistant tokens at the `onTokenLiteral` boundary before they reach TTS, with the projector reset per message so a partial annotation can't leak across turns, and flushed at `onStreamEnd`. This side-channel is TTS-only, so **the chat display is untouched** (no regression) — it keeps rendering the raw stream today.

Rendering the `displayText` in the chat message (stripping the markup from what the user sees) is a natural **follow-up** in the message renderer; I kept it out of this PR so the boundary change stays minimal and reviewable. Happy to do it here too if you'd prefer.

### Tests

`ruby-annotation.test.ts` (9 cases, `vitest`), including the hard part — an exhaustive sweep asserting the output is identical for **every possible chunk split point**, plus one-character-at-a-time streaming, multiple annotations, malformed/unterminated input, and "a partial annotation is withheld until it completes".

```
Test Files  1 passed (1)
     Tests  9 passed (9)
```

### Notes for reviewers

The issue is labelled `needs-more-info`, so I've kept the **syntax** (`｜base《reading》`) and **placement** as an explicit proposal — both are easy to change: the syntax is three constants in one file, and the projection is a pure function you can drop anywhere along the boundary. Let me know if you'd like a different notation, a different location, or the display-side rendering folded in.

Addresses #2255

### DIFF
--- a/packages/stage-ui/src/components/scenes/Stage.vue
+++ b/packages/stage-ui/src/components/scenes/Stage.vue
@@ -39,6 +39,7 @@ import { live2dMotionMagicProfiles, useLive2DMotionMagic, useLive2DMotionMagicSe
 import { getDefaultStreamingModel, getDefinedProvider } from '../../libs/providers/providers'
 import { OFFICIAL_SPEECH_PROVIDER_ID, OFFICIAL_SPEECH_STREAMING_PROVIDER_ID } from '../../libs/providers/providers/official'
 import { bindSpeakingStateToPlaybackManager } from '../../libs/speech/playback-speaking-state'
+import { createRubyProjector } from '../../libs/speech/ruby-annotation'
 import { createStageTtsSession } from '../../libs/speech/tts-session'
 import { getSpeechBusContext, speechOutputGetPlaybackState } from '../../services/speech/bus'
 import { useLlmStreamingControlStore } from '../../stores/ai/chat-llm/streaming-control'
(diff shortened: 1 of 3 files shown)


nekomeowww (commented):
《 and 》 is bad for this.
you need benchmark it before deciding

PR #2414, opened 2026-08-30
fix(pipelines-audio): preserve multi-code-unit grapheme clusters in TTS chunking

### LINKED ISSUE #2366: Thai (and other space-less script) TTS text gets corrupted when streamed from chat — likely due to space-based text chunking
**Description**
When chatting with a custom OpenAI-compatible provider (both Chat and Speech set to a self-hosted OpenAI-compatible endpoint), the text sent to the TTS provider is corrupted for Thai language responses. The corruption looks like the underlying UTF-8 bytes were split at the wrong boundary and reassembled into different (but still valid) Thai characters, producing gibberish that no longer matches the original sentence.

**Reproduction**
1. Set Chat provider to a custom OpenAI-compatible endpoint that replies in Thai.
2. Set Speech provider to OpenAI-compatible TTS, pointing to the same local server.
3. Send a chat message and let Mia/character reply in Thai, with the reply being spoken automatically.
4. Compare the spoken audio against the actual chat text.

**Evidence**
Original chat reply (confirmed correct, as displayed in the chat UI):
> ดึกป่านนี้แล้วยังจะหาเรื่องกินอีกนะคะเนี่ย! เดี๋ยวก็อ้วนเป็นหมูรอกค่ะ...

Actual text received by the TTS endpoint (captured via server-side logging on my own OpenAI-compatible TTS server) for the same reply, split into two `/v1/audio/speech` calls:
> 'การเบงจงกร เองนละคะเยฯ!'
> 'เยวรน เธอกอะ แกเบยอมเนไ...'

Note this is NOT just missing tone marks — entire character sequences are different from the original, which points to a byte-level splitting issue rather than a text-processing/formatting issue. Thai is a multi-byte-per-character script in UTF-8 (3 bytes/char), so any split that doesn't respect character boundaries will produce this kind of corruption when the resulting byte fragments are reassembled/decoded.

**Isolating the cause**
- I built a minimal Flask server implementing both `/v1/chat/completions` (with SSE streaming, since AIRI's chat UI requires streaming to render messages at all — confirmed non-streaming responses aren't displayed) and `/v1/audio/speech` (OpenAI-compatible TTS, backed by Typecast).
- Pasting the exact same Thai sentence directly into Settings → Speech → Voice Playground and clicking "Test Voice" produces **correct, unaffected audio** — this calls the same `/v1/audio/speech` endpoint and proves both my TTS backend and the endpoint itself work correctly with the full, uncorrupted text.
- The corruption **only happens when AIRI automatically speaks a chat reply**, not when the same text is sent manually via the playground. This points to whatever logic intercepts/chunks the streamed chat text before forwarding it to the speech provider.

**Suspected root cause**
Found an old dev chronicle (v0.0.1, 2024) describing the TTS streaming design: "Split text by spaces to get ['hello ', 'this ', 'is ', 'neuro ']" — chunking streamed text into word groups using spaces as delimiters, presumably for timing/sync purposes. This works for space-delimited languages (English) but **Thai (like Japanese, Khmer, Lao) does not use spaces between words at all**. If the current implementation still relies on this design, or falls back to a fixed-length/byte-length split when no space is found within a reasonable window, it would explain exactly the character-boundary corruption seen here. This would likely also affect other space-less or multi-byte-heavy scripts (Japanese, Chinese, Khmer, Lao, etc.), not just Thai.

**Suggested fix direction**
- If chunking is done for TTS timing sync, chunk by Unicode code points / grapheme clusters (not raw bytes) as a baseline safety measure regardless of language.
- For sentence/phrase-level chunking, use locale-aware sentence segmentation (e.g., `Intl.Segmenter` with `granularity: 'sentence'` or `'word'`, which correctly handles space-less scripts) instead of `String.split(' ')`.

**Environment**
- AIRI Stage Tamagotchi (Windows desktop app)
- Chat provider: Custom OpenAI-compatible (local Flask server)
- Speech provider: OpenAI-compatible TTS (local Flask server, backed by Typecast)
- Language: Thai

Happy to provide the minimal reproduction server code if useful.

Jorge-Polanco-Roque:
### Problem

Fixes #2366. When a chat reply in Thai (or other space-less / combining scripts) is spoken automatically, the text forwarded to the TTS provider is corrupted — the chat UI shows the correct sentence, but the speech uses different, valid-but-wrong characters:

```
chat:   ดึกป่านนี้แล้วยังจะหาเรื่องกินอีกนะคะเนี่ย
spoken: การเบงจงกร เองนละคะเยฯ  (gibberish)
```

### Root cause

Not the space-based splitting the report suspected — that path already segments with `Intl.Segmenter` / grapheme clusters. The real cause is in `chunkTtsInput` (`pipelines-audio/src/processors/tts-chunker.ts`), the chunker behind `createSpeechPipeline` that chat auto-TTS uses:

```ts
if (value.length > 1) {
  previousValue = value
  current = await iterator.next()
  continue   // ← the cluster is dropped, never added to the buffer
}
```

`value` is a grapheme cluster. The guard skips any cluster whose string is longer than one UTF-16 code unit. In Thai/Lao/Khmer/Devanagari a syllable is a **base consonant + combining vowel/tone marks**, i.e. a multi-code-unit cluster — so those syllables were silently discarded, and the bare single-unit consonants left behind reassembled into different words. That is exactly the "entire character sequences are different" corruption in the report.

Reproduced as a unit test: feeding `ดึกป่านนี้แล้วยังจะหาเรื่องกินอีกนะคะเนี่ย` through `chunkTtsInput` yields `กานแวงจะหาเองนกนะคะเย` on `main`.

### Fix

A cluster longer than one code unit is never one of the single-char punctuation markers checked below, so it is ordinary text — append it to the buffer instead of dropping it:

```ts
if (value.length > 1) {
  buffer += value
  previousValue = value
  current = await iterator.next()
  continue
}
```

One line of behaviour change. This also preserves emoji ZWJ sequences (which were being dropped too).

### Tests

Adds regression cases (Thai, Devanagari, ASCII) to `tts-chunker.test.ts` asserting no characters are dropped. The Thai case **fails without the fix** and passes with it:

```
Test Files  1 passed (1)
     Tests  17 passed (17)
```

### Scope

Only the TTS chunking path is touched; the chat display is unaffected (it was already correct). The fix is language-agnostic — any base+combining script benefits, not just Thai.

Fixes #2366

### DIFF
--- a/packages/pipelines-audio/src/processors/tts-chunker.test.ts
+++ b/packages/pipelines-audio/src/processors/tts-chunker.test.ts
@@ -2,7 +2,7 @@
 
 import { describe, expect, it } from 'vitest'
 
-import { isProbablyAngleTag, processNarrative } from './tts-chunker'
+import { chunkTtsInput, isProbablyAngleTag, processNarrative } from './tts-chunker'
 
 describe('tTS Chunker Logic Cleanup', () => {
   describe('isProbablyAngleTag Heuristics', () => {
@@ -88,5 +88,45 @@ describe('tTS Chunker Logic Cleanup', () => {
       expect(isProbablyAngleTag(4, 'café<laugh>')).toBe(true)
       expect(isProbablyAngleTag(6, 'привет<sigh>')).toBe(true)
     })
+  })
+})
+
+describe('chunkTtsInput — space-less / combining scripts (#2366)', () => {
+  async function reconstruct(input: string): Promise<string> {
+    let out = ''
+    for await (const chunk of chunkTtsInput(input))
+      out += chunk.text
+    return out
+  }
+
+  // Chunk boundaries trim surrounding whitespace, so compare with whitespace
+  // removed to assert only that no *characters* were dropped.
+  const stripWs = (s: string) => s.replace(/\s+/g, '')
+
+  it('preserves Thai text with combining vowels and tone marks', async () => {
+    // No spaces between words, and most syllables are multi-code-unit grapheme
+    // clusters (base consonant + combining vowel/tone) — the exact case that
+    // used to be dropped, corrupting the speech text.
+    const input = 'ดึกป่านนี้แล้วยังจะหาเรื่องกินอีกนะคะเนี่ย'
+    expect(await reconstruct(input)).toBe(input)
+  })
+
+  it('preserves Devanagari text with combining matras', async () => {
+    const input = 'नमस्ते दुनिया यह एक छोटा परीक्षण है'
+    expect(stripWs(await reconstruct(input))).toBe(stripWs(input))
+  })
+
+  it('does not drop characters from plain ASCII text', async () => {
+    const input = 'Hello there, this is a normal English sentence.'
+    expect(stripWs(await reconstruct(input))).toBe(stripWs(input))
+  })
+
+  it('treats a CRLF grapheme cluster as a hard boundary', async () => {
+    // `\r\n` is a single grapheme cluster, so it hits the multi-unit branch;
+    // it must still split the chunk like a lone `\n` (not be buffered inline).
+    const chunks: string[] = []
+    for await (const chunk of chunkTtsInput('abc\r\ndef'))
+      chunks.push(chunk.text)
+    expect(chunks).toEqual(['abc', 'def'])
   })
 })
--- a/packages/pipelines-audio/src/processors/tts-chunker.ts
+++ b/packages/pipelines-audio/src/processors/tts-chunker.ts
@@ -70,9 +70,23 @@ export async function* chunkTtsInput(
     let value = current.value
 
     if (value.length > 1) {
-      previousValue = value
-      current = await iterator.next()
-      continue
+      // A grapheme cluster spanning multiple UTF-16 code units — e.g. a base
+      // consonant plus its combining vowel/tone in Thai/Lao/Khmer/Devanagari,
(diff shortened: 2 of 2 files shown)


Jorge-Polanco-Roque:
Good catch from the automated review — thank you. You're right that `packages/stage-ui/src/utils/tts.ts` is not the chunker chat auto-TTS runs through; it's an unused duplicate. I've moved the fix to the active chunker, `packages/pipelines-audio/src/processors/tts-chunker.ts` (the segmenter behind `createSpeechPipeline`), and added the regression tests there. Reverted the change to the dead file so this PR only touches the path that actually fixes #2366.

Jorge-Polanco-Roque:
Good catch on the CRLF case. `\r\n` is a single grapheme cluster (UAX #29 GB3), so it hits the multi-unit branch and my change buffered it, dropping the hard boundary a lone `\n` would trigger. Fixed: the multi-unit branch now normalizes `\r\n` to `\n` and falls through to the punctuation handling, so Windows line endings still split the chunk. Added a regression test (`treats a CRLF grapheme cluster as a hard boundary`). 18/18 passing.

Jorge-Polanco-Roque:
Gentle ping — open ~a week. It preserves multi-code-unit grapheme clusters when chunking text for TTS, so emoji and combining marks aren't split mid-cluster. Happy to tweak on request. Thanks!

PR #2422, opened 2026-08-31
fix(stage-tamagotchi): clamp restored main-window bounds onto an available display

### LINKED ISSUE #2181: Desktop: dragging the main window off-screen persists an unrecoverable position
## Summary

Dragging the AIRI desktop window beyond a screen edge can leave it completely off-screen. The invalid position is persisted and restored unchanged on the next launch, so the on-window recovery control cannot be reached.

A report from the community shows this on Windows; it should be investigated across Electron desktop platforms.

## Steps to reproduce

1. Open the AIRI desktop (Tamagotchi) window.
2. Use the move/drag control and drag the window past a display edge until it is no longer visible.
3. Close and reopen AIRI.

## Actual behavior

The window remains outside the visible work area after restart. Users must discover the tray-icon position command to recover it; the built-in **Center main window** control is unusable because it lives inside the off-screen window.

## Expected behavior

The app should retain a reachable portion of the window at all times, and should restore saved bounds only after validating/clamping them against an available display work area. If a saved display is unavailable or the window has no visible intersection, restore it to a safe position (for example, centered on the matching/primary display).

## Investigation

- The main window persists every `move` and `resize` event, including fully off-screen bounds: `apps/stage-tamagotchi/src/main/windows/main/index.ts` (`handleNewBounds`, lines 119–152).
- Startup passes the saved `x` and `y` directly to `BrowserWindow` without any display/work-area validation (lines 78–98).
- A clamping/recovery pattern already exists for the widgets window, and shared display helpers already center a main window. Applying equivalent validation before main-window creation would prevent this state.
- The existing **Center main window** action is rendered in the main window’s controls island, so it cannot recover a window that is fully off-screen.

## Suggested acceptance criteria

- Dragging cannot persist a main-window position with no visible/reachable portion on any connected display.
- Startup validates persisted bounds against current displays (including removed monitors and changed resolutions/scaling).
- An invalid saved position is replaced with a safe visible position while preserving a reasonable window size.
- Add focused tests for fully off-screen, partially off-screen, and unavailable-display saved bounds.

## Evidence

Community screenshots (provided with this report) show the window dragged off the visible display and the tray-menu workaround for centering it.

Jorge-Polanco-Roque:
### Problem

Fixes #2181. The desktop (Tamagotchi) main window persists its position on every move/resize and restores it verbatim on the next launch, with no validation against the current displays. A window dragged past a screen edge — or saved on a monitor that is later disconnected — reopens completely off-screen, and the in-window **Center main window** control is unreachable (it lives inside the invisible window), so the user has to discover the tray command to recover it.

### Root cause

`main/index.ts` restores the saved position straight into the `BrowserWindow` constructor:

```ts
const mainWindowConfig = getConfig().windows?.find(...)
const window = new BrowserWindow({
  width: mainWindowConfig?.width ?? 450.0,
  height: mainWindowConfig?.height ?? 600.0,
  x: mainWindowConfig?.x,   // ← applied unchecked
  y: mainWindowConfig?.y,
  ...
})
```

There is no clamp against a display work area, so an off-screen `x/y` is faithfully reproduced.

### Fix

A pure helper `sanitizePersistedWindowBounds(bounds, workAreas, primaryWorkArea)`:

- If the saved bounds still overlap a display, clamp them fully into that display's work area (best-overlapping display wins).
- If they no longer intersect **any** display, center on the primary work area (safe fallback for a disconnected monitor or a window dragged fully off-screen).

`resolveMainWindowBounds` applies it before creating the window, feeding it `screen.getAllDisplays()`/`getPrimaryDisplay()`. When there is no saved position, `x/y` stay `undefined` so Electron centers the window exactly as before.

The helper is Electron-free (plain rectangles in, rectangle out), so it is fully unit-tested without launching the app:

```
Test Files  1 passed (1)
     Tests  6 passed (6)
```

covering: on-screen bounds unchanged, clamping from each edge and from negative positions, the off-every-display → center fallback, a window kept on a secondary display it still overlaps, and the no-displays fallback.

### Notes

The caption window already has a private `clampBoundsWithinRect` helper with the same shape; I kept this PR focused on the main-window bug rather than refactoring both onto the new shared helper, but that consolidation would be a natural follow-up.

Fixes #2181

### DIFF
--- a/apps/stage-tamagotchi/src/shared/utils/electron/windows/window-bounds.ts
+++ b/apps/stage-tamagotchi/src/shared/utils/electron/windows/window-bounds.ts
@@ -1,5 +1,17 @@
 import type { Rectangle } from 'electron'
 
+import { clamp } from 'es-toolkit'
+
+/**
+ * The parts of an Electron `Display` this module needs: the full physical
+ * `bounds` (used to decide which monitor a window belongs to) and the
+ * `workArea` (excludes the taskbar/dock, used to place the window).
+ */
+export interface DisplayLike {
+  bounds: Rectangle
+  workArea: Rectangle
+}
+
 /** Area of the overlap between two rectangles (0 when they do not intersect). */
 function intersectionArea(a: Rectangle, b: Rectangle): number {
   const left = Math.max(a.x, b.x)
@@ -15,9 +27,12 @@ function intersectionArea(a: Rectangle, b: Rectangle): number {
 function clampWithin(bounds: Rectangle, rect: Rectangle): Rectangle {
   const width = Math.min(bounds.width, rect.width)
   const height = Math.min(bounds.height, rect.height)
-  const x = Math.min(Math.max(bounds.x, rect.x), rect.x + rect.width - width)
-  const y = Math.min(Math.max(bounds.y, rect.y), rect.y + rect.height - height)
-  return { x, y, width, height }
+  return {
+    x: clamp(bounds.x, rect.x, rect.x + rect.width - width),
+    y: clamp(bounds.y, rect.y, rect.y + rect.height - height),
+    width,
+    height,
+  }
 }
 
 /** Center a window of the given size within `rect`. */
@@ -33,8 +48,8 @@ function centerWithin(size: { width: number, height: number }, rect: Rectangle):
 }
 
 /**
- * Validate persisted window bounds against the currently available display work
- * areas so a reachable portion of the window always stays on-screen.
+ * Validate persisted window bounds against the currently available displays so a
+ * reachable portion of the window always stays on-screen.
  *
  * Use when:
  * - Restoring a window position saved in a previous session, which may now be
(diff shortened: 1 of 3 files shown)


apps/stage-tamagotchi/src/shared/utils/electron/windows/window-bounds.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/** Move `bounds` so they sit fully within `rect` (position only; size preserved, shrunk only if larger than `rect`). */
+function clampWithin(bounds: Rectangle, rect: Rectangle): Rectangle {
+  const width = Math.min(bounds.width, rect.width)
+  const height = Math.min(bounds.height, rect.height)
+  const x = Math.min(Math.max(bounds.x, rect.x), rect.x + rect.width - width)
+  const y = Math.min(Math.max(bounds.y, rect.y), rect.y + rect.height - height)
+  return { x, y, width, height }
+}
```
nekomeowww, line 36: clamp should use es-toolkit

Jorge-Polanco-Roque:
Thanks both. Addressed:
- `clampWithin` now uses es-toolkit's `clamp` (@nekomeowww).
- Display ownership is determined from full display `bounds` and the window is then clamped into that display's `workArea`, so a window over a taskbar/dock strip stays on the monitor it physically sits on instead of jumping to a neighbour whose work area it overlaps more. Added a regression test for that case.

nekomeowww:
Duplicated

PR #2457, opened 2026-09-03
fix(stage-pocket): add missing account settings route

BeanDz:
| Before | After |
| --- | --- |
| Before | After |


LemonNekoGH (approved):
LGTM

PR #2498, opened 2026-09-09
WIP live2d ambient light with normal map and pseudo-3d light remodel

shinohara-rin:
parent: #2391


shinohara-rin:
try https://github.com/huawei-bayerlab/marigold-v2

PR #2545, opened 2026-09-14
fix(stage-ui): clone synchronized provider config

0xSelenicDove:
## Summary

- create plain provider configuration snapshots before follower actions cross `BroadcastChannel`
- cover nested Vue proxies with a structured-clone regression test
- exercise repeated speech-setting edits through a real leader and follower Pinia pair

## Verification

- `pnpm -F @proj-airi/stage-ui exec vitest run src/libs/providers/config.test.ts`
- `pnpm -F @proj-airi/stage-pages exec vitest run src/pages/settings/providers/speech/provider-config-persistence.browser.test.ts`
- `pnpm typecheck`
- `pnpm lint`


nekomeowww (changes requested):
What is the purpose of this PR?

PR #2578, opened 2026-09-17
fix(stage-tamagotchi): hold clicks where a scene is painted

chiba233:
### The problem

#2573 made blank stage pixels pass clicks through to the application behind. A
character card can also pick a scene, and with one set the window looks painted from
edge to edge — yet clicks still fell through everywhere except the character.

### Why the hit test could not see the scene

The scene was a CSS background on a layer the stage paints behind the model, not canvas
content. The pixel test reads the model canvas, which is empty wherever the scene alone
shows, and the document test added in #2573 returns the topmost element, which is that
same canvas sitting above the scene layer. An opaque layer *below* the canvas was the
case neither was built for.

### The approach

The scene now paints into a canvas of its own, applying the same `cover` fit and centre
position by hand, and the stage hit-tests it with the very sampler the model goes
through. No bespoke sampler, no second copy of the geometry, and nothing in the
interaction that knows a scene exists.

The judgement therefore reads the pixels that are displayed rather than recomputing
where they went. Scene art carries its own transparency — a quarter of the built-in
backgrounds is fully clear — so the feathered margins keep passing clicks through
exactly where they are clear.

Reading a canvas that a 2D context owns needs the sampler to take the context as a
declaration rather than discover it. `getContext` creates a context on a canvas that
has none, so probing would lock a canvas out of the kind its renderer needs.

### On the repaint

The paint is synchronous, with the decoded image held ready. Sizing a canvas clears it,
so anything awaited between the clear and the draw shows as a blank frame, once per
frame of a window resize. An earlier revision awaited the decode there and flickered
through the whole drag.

## Linked Issues

Follow-up to #2573, which introduced the behaviour this completes. Supersedes #2576,
which special-cased the scene instead of letting it answer for itself.

## Additional Context

**This changes rendering in a shared component.** `Stage.vue` lives in `stage-ui`, and
`stage-web` and `stage-pocket` render it too, so their scene layer moves from a CSS
background to a canvas as well. Nothing about their behaviour should change — the same
image, the same fit — but the path is new for them and they gain a canvas they do not
read. Worth a look from someone who runs those surfaces.

**Three things the browser used to do are now explicit.** Sizing, device pixel ratio,
and redrawing when either changes. The watchers cover a scene change, an element resize
and a ratio change; a repaint trigger outside those would show as a stale or stretched
scene.

**Verification.** The coordinate mapping was compared against what the browser paints,
by drawing the same image into a canvas at the element's size and reading it back:
72941 sampled points, zero disagreements, corners clear and centre painted. The 2D
reader is covered by two unit tests, one pinning the row it reads — the only
substantive difference from the WebGL reader, and one nothing in the types would catch.

**No screenshots.** The scene should render identically and the change that matters is
where a click lands, which a still frame cannot show. A short recording of clicking
through a feathered margin is the evidence worth having, and it needs a desktop session
to capture.

Verified with:

```sh
pnpm -F @proj-airi/stage-ui typecheck
pnpm -F @proj-airi/stage-tamagotchi typecheck
pnpm -F @proj-airi/stage-ui exec vitest run --project node
pnpm -F @proj-airi/stage-tamagotchi exec vitest run --project node
pnpm lint
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


chiba233:
Superseded by a different approach: the Pixi renderers now paint the scene into the canvas they already draw to, so the pixel test that exists covers it and the click-through logic needs no change at all. This PR's separate sampler and its background branch in the interaction are unnecessary under that shape.

PR #2594, opened 2026-09-18
fix(api): normalize OpenRouter Responses event names

luoling8192:
## Summary

- normalize mismatched OpenRouter SSE event names to the event type in `data`
- keep strict event-name validation for all other providers
- keep the output lifecycle checks from #2593 before recovering a missing terminal event

## Why

After #2593, an OpenRouter stream still failed before EOF recovery with `Responses SSE event name does not match its payload type`. OpenRouter can attach an SSE `event:` name that differs from the Responses event type in the JSON payload.

The gateway now treats `data.type` as authoritative for OpenRouter and forwards a matching SSE event name. It records one warning when it applies this normalization.

## Validation

- `pnpm -F @proj-airi/api-server exec vitest run` (64 files, 623 tests)
- `pnpm typecheck` (52 tasks)
- `pnpm lint` (0 errors; existing repository warnings remain)
- focused OpenRouter regression tests after rebasing on #2593 (4 tests)


luoling8192:
Superseded by #2596, which includes this server fix together with the client-side interrupted-stream handling and retry behavior.

PR #2598, opened 2026-09-19
feat(providers): add Y-API provider

jiweiyeah:
## Description

Add a provider definition for Y-API. Y-API is an OpenAI-compatible gateway that serves models from several vendors behind one endpoint.

The definition follows `n1n`. It needs an API key and a base URL. The default base URL is `https://api.y-api.bestvirtualgoods.com/v1`.

`@proj-airi/lobe-icons` has no Y-API icon. The definition uses `i-lobe-icons:openai`. `n1n`, OpenPaths, and Atlas Cloud use the same fallback.

## Additional Context

I operate Y-API. Read this PR as a provider request from an interested party, not as an impartial review.

### Validator results

I ran the three validators against the live endpoint with a real API key.

| Check | Request | Result |
| --- | --- | --- |
| Connectivity | `GET {baseUrl}/models` | HTTP 200 |
| Model list | `GET {baseUrl}/models` | HTTP 200, 15 models |
| Chat completions | `POST {baseUrl}/chat/completions` | HTTP 200 |

`pickValidationModel` selects `deepseek/deepseek-v4-flash` for Y-API. The chat check sends `max_tokens: 16` to that model. The request passed.

`GET {baseUrl}/models` answers HTTP 401 without an API key. The definition requires an API key, so the check runs only after the user enters one.

### Files

- `packages/provider-inference/src/providers/cloud/y-api/index.ts` — new definition
- `packages/provider-inference/src/providers/index.ts` — register the definition
- `packages/i18n/src/locales/en/settings.yaml` — English source strings
- `packages/i18n/src/locales/zh-Hans/settings.yaml` — Chinese strings

I changed only these two locales. AGENTS.md limits locale edits to the English source and the developer's own locale.

### Schema library

AGENTS.md asks for Valibot. Every provider in `providers/cloud` uses Zod. I used Zod to match them. Tell me if you want Valibot, and I will move the file.


nekomeowww:
Batched, automated without human, and grinding for adding supports of providers to our repository is prohibited.

PR #2607, opened 2026-09-19
perf(stage-tamagotchi): stop re-sending unchanged cursor position and window bounds

FlowerWater1019:
## Summary

`createRendererLoop` defaults to a `1000 / 60` interval, and neither `createScreenService` nor `createWindowService` passes one, so both loops tick 60 times a second and emit unconditionally.

The cursor is stationary for most of those ticks, and window bounds change rarer still, so the renderer paid an IPC deserialize per tick for a value that had not changed. The cursor path costs more than that: `useElectronMouseEventTarget` constructs a synthetic `MouseEvent` and dispatches it through an `EventTarget` for every message, so an unmoved cursor still produced an allocation plus a listener run 60 times a second.

Both loops now emit only when the value actually changes.

```diff
       const dipPos = screen.getCursorScreenPoint()
+      const now = Date.now()
+      const moved = !lastPoint || lastPoint.x !== dipPos.x || lastPoint.y !== dipPos.y
+      if (!moved && now - lastEmittedAt < RESEND_INTERVAL_MS)
+        return
+
+      lastPoint = { x: dipPos.x, y: dipPos.y }
+      lastEmittedAt = now
       params.context.emit(cursorScreenPoint, dipPos)
```

### Why the 1s re-emit

Consumers learn these values solely from the event: `useElectronMouseEventTarget` feeds `useMouse`, and `useElectronWindowBounds` mirrors `bounds` into a ref. A consumer that mounts while the cursor sits still would otherwise hold a stale or absent position until the user happened to move the mouse. Each loop therefore still re-emits once a second, which preserves the eventual-correctness the unconditional emit gave, at 1/60th of the traffic.

## Measurement

With the cursor idle, each loop goes from 60 emits per second to 1. Across the two loops the renderer stops deserializing roughly 120 messages a second, and stops synthesizing roughly 60 `MouseEvent`s a second, all for values that had not changed.

I found this while profiling a Windows desktop build whose renderer sat above one full core with the GPU only at 17%, and whose main process was holding 30-40% of a core with these two loops as the visible work.

I am deliberately not quoting a frame-rate delta: the session I measured had an unrelated problem in flight at the same time, so I could not attribute a before/after difference to this change cleanly. The traffic reduction above is structural and does not depend on that measurement.

## How tested

```
pnpm -F @proj-airi/stage-tamagotchi typecheck
pnpm exec eslint apps/stage-tamagotchi/src/main/services/electron/screen.ts apps/stage-tamagotchi/src/main/services/electron/window.ts
```

Both clean. Also running the patched build on Windows as a daily driver: the cursor-driven surfaces (controls island hover state, ghost-mode click-through) behave as they did before.

No automated test: these two loops are main-process service wiring and this app has no existing harness that drives `createRendererLoop`. Happy to add one if you would like the emit-on-change predicate covered directly.

## Visual changes

None. Both files are main-process IPC wiring with no UI surface, and the change alters *when* an event is published rather than what any component renders. On real movement the delivered values are identical to before.


FlowerWater1019:
Converting to draft while I investigate a possible regression from this change.

The contributor running it reports that the controls-island buttons became intermittently unclickable after it landed locally, which also blocks opening settings. Suspected mechanism: predicates like `isOutside` for the controls island are driven by this cursor event stream, and click-through is computed from them. Emitting only on movement means a geometry change that is not caused by the cursor - dragging the window, the island resizing, a model swap relaying out the overlay - leaves those predicates stale until the next mouse move. The unconditional 60Hz stream was acting as an accidental poll that corrected them within a frame.

That is a hypothesis, not a confirmed diagnosis - I have reverted it locally to test. Drafting rather than closing so it can come back if the cause turns out to be elsewhere, or with the stale-predicate problem addressed properly (a keepalive alone is not enough, since the staleness lasts until the next emit).

Sorry for the noise.

FlowerWater1019:
Confirmed, and closing.

Reverting this change restored the behaviour immediately: the controls-island buttons respond again and the window can be dragged. So the stale-predicate mechanism I described is the actual cause, not just a hypothesis.

The reason it is not a small fix: this event stream has a second, undocumented job. Predicates like `isOutside` for the controls island are derived from cursor events, and click-through is computed from those predicates — but a geometry change that is not caused by the cursor (dragging the window, the island resizing, a model swap relaying out the overlay) also invalidates them. The unconditional 60Hz emit was silently acting as the poll that re-derived them within a frame. Throttling it removes that correction, and a periodic keepalive does not fix it either: the predicate is simply wrong for however long it takes the next emit to arrive.

Doing this properly means making those consumers recompute on geometry change rather than leaning on the cursor stream as a clock, which is a different and much larger change than this one. Closing rather than reworking in place; happy to come back with that version separately if it is wanted.

Apologies for the noise — the idle traffic saving was real, but not worth this.

PR #2603, opened 2026-09-19
 [DO NOT MERGE] perf(stage-tamagotchi): reduce packaged desktop bundle size

nayounsang:
## Description

This PR reduces the packaged `stage-tamagotchi` application size. It removes files that are not required at runtime and loads model preview code on demand.

The optimization methods are:

- Remove duplicate ONNX runtime packages — [`e55204825`](https://github.com/moeru-ai/airi/commit/e55204825).
- Exclude unused CJK font packages from the desktop renderer — [`79fde505`](https://github.com/moeru-ai/airi/commit/79fde505).
- Remove foreign platform native binaries — [`774faed9d`](https://github.com/moeru-ai/airi/commit/774faed9d).
- Load model preview renderers on demand — [`ebe73e980`](https://github.com/moeru-ai/airi/commit/ebe73e980).
- Remove packaging-only dependencies and files — [`a8f0df484`](https://github.com/moeru-ai/airi/commit/a8f0df484).

The measurements use bytes divided by 1,048,576.

| Artifact | Existing | Applied | Improvement |
| --- | ---: | ---: | ---: |
| Renderer output | 333.81 MB | 217.77 MB | 34.8% |
| `app.asar` | 1,297.11 MB | 946.42 MB | 27.0% |
| `app.asar.unpacked` | 212.80 MB | 29.76 MB | 86.0% |
| Unpacked macOS application | 1,803.94 MB | 1,270.22 MB | 29.6% |

The measurements represent an unsigned, unpacked macOS arm64 application.

## Environment

The environment output came from this command:

```sh
npm exec --yes envinfo -- --system --binaries --npmPackages electron,electron-builder
```

| Item | Value |
| --- | --- |
| Operating system | macOS 26.6.2 |
| CPU | Apple M1, 8 cores |
| Architecture | arm64 |
| Node.js | 24.18.1 |
| npm | 11.16.0 |
| pnpm | 11.24.0 |
| Electron | 43.4.1 |
| electron-builder | 26.16.1 |

## Measurement

The baseline uses [`0040675e8`](https://github.com/moeru-ai/airi/commit/0040675e8). The applied state includes these optimization commits:

- [`e55204825`](https://github.com/moeru-ai/airi/commit/e55204825)
- [`79fde505e`](https://github.com/moeru-ai/airi/commit/79fde505e)
- [`774faed9d`](https://github.com/moeru-ai/airi/commit/774faed9d)
[...]

Issue #1719, opened 2026-04-23
desktop application bundle size optimization


drHuangMHT:
analysis on install location of desktop app(tamagotchi) on Windows x64 platform reveals several places for optimization:
- duplication of ONNXRuntime: `onnxruntime-web` and `onnxruntime-node` bundled into desktop
  - `onnxruntime-node` is a dependency of `@huggingface/transformers`
  - `onnxruntime-web` is a dependency of `@huggingface/transformers` and `@ricky0123/vad-web`
  - `onnxruntime-node` bundles both arm64 and x64 binaries, and binaries for linux(.so), windows(.dll), darwin(.dylib) are also bundled and shipped, regardless of the installed platform.
- duplication of out-of-the-box Live2D and VRM models.
- font files for `font-cjkfonts-allseto` and `font-xiaolai` are duplicated in their `src` and `dist` path.

PR #2611, opened 2026-09-20
perf(stage-tamagotchi): exclude duplicate ONNX runtimes

nayounsang:
## Description

Reduce the size of the desktop application package.

Keep the browser ONNX WASM emitted by Vite. Exclude onnxruntime-node and the duplicate onnxruntime-web package tree from the desktop ASAR.

This is minimally affected by the OS(architecture) because it is affected by browser WASM runtime.

| Artifact | Existing | Applied | Improvement |
| --- | ---: | ---: | ---: |
| Renderer output | 350,027,841 bytes | 350,027,846 bytes | 0.0% |
| `app.asar` | 1,360,117,408 bytes | 1,222,515,329 bytes | 10.1% |
| `app.asar.unpacked` | 223,132,648 bytes | 74,647,540 bytes | 66.5% |
| Unpacked macOS application | 1,891,573,024 bytes | 1,605,485,837 bytes | 15.1% |

## Linked Issues

#1719 

## Additional Context

ref & context: #2603


apps/stage-tamagotchi/electron-builder.config.ts
```diff
@@ -81,6 +81,11 @@ export default {
     // fails on non-code resources (for example `locale.pak`) with timestamp/signing errors.
     '!**/node_modules/electron{,/**}',
     '!**/.vscode/*',
+    // The renderer uses the browser ONNX backend. The node package and its
+    // platform binaries are not loaded by the packaged renderer.
+    '!**/node_modules/onnxruntime-node{,/**}',
+    // Vite bundles the browser runtime and its WASM assets into `out/renderer`.
+    '!**/node_modules/onnxruntime-web{,/**}',
```
nayounsang, line 88: Vite bundles the browser runtime into out/renderer. this rule only removes the duplicate package tree from the packaged app.

PR #2627, opened 2026-09-20
fix(stage-ui): initialize Kokoro catalogs before discovery

### LINKED ISSUE #2626: bug: kokoro-local model & voice catalogs return no data → provider can't be activated, chat replies are never spoken
### Describe the bug

The **Kokoro TTS (Local)** speech provider cannot be activated on a fresh install: its **model list and voice list both come back empty**, so there is nothing to select. `activeSpeechModel` / `activeSpeechVoiceId` stay empty, and since the speech store's `configured` is `hasModel && hasVoice`, it stays `false` — the chat→TTS path (`chat_auto_tts`) is never taken, so **chat replies are never spoken**.

The app anticipates providers whose catalog returns nothing (the UI offers a *manual* model/voice entry), but a normal user hits a dead end: **Test Voice is disabled** and the only explanation is "no models available".

### Steps to reproduce

1. Fresh browser profile → https://airi.moeru.ai → complete onboarding with a Chat provider (DeepSeek, `deepseek-flash`).
2. `Settings → Providers → Speech` → open the **Kokoro TTS (Local)** card.
3. Observe: the `Model` section shows *no models available* (`该服务来源的模型列表没有返回任何的数据`) and `Voice Configuration` shows *no voices available* (`没有任何可用的声线`). The manual inputs below them are the only way in.
4. `Settings → Modules → Speech` → select **Kokoro TTS (Local)**: only the **Provider** radio group renders — the **Model and Voice sections never render**.
5. Inspect the speech store: `activeSpeechModel === ''`, `activeSpeechVoiceId === ''`, `configured === false`.
6. On the stage, send a message: the assistant replies in text but **nothing is played**.

### Expected behavior

Selecting the Kokoro speech provider should expose its models (`FP32 (WebGPU) / FP32 (WASM) / FP16 (WASM) / Q8 (WASM) / Q4 (WASM) / Q4F16 (WASM)`) and its voices (`af_heart`, …). Selecting them should set `activeSpeechModel` / `activeSpeechVoiceId` → `configured === true` → replies get spoken.

### Actual behavior

Catalog returns nothing → nothing selectable → `configured` can never become `true` → chat TTS never fires; Test Voice stays disabled.

### Workaround (verified)

On `Settings → Speech`, type the model name by hand (`q8` in *"Enter model name manually when no models are returned"*) and the voice name (`af_heart`). After that **Test Voice becomes enabled, speech is produced, and the local voice works.** This is the only way we could get it working.

### Evidence

Store dump taken from the stage page (DevTools console):

```js
(() => {
  let app = null;
  for (const el of document.querySelectorAll('body *')) if (el.__vue_app__) { app = el.__vue_app__; break; }
  const s = app?.config?.globalProperties?.$pinia?._s?.get('speech');
  console.table({ configured: s.configured, provider: s.activeSpeechProvider,
                  model: s.activeSpeechModel, voice: s.activeSpeechVoiceId });
})();
// → { configured: false, provider: 'kokoro-local', model: '', voice: '' }
```

Also observed once, right after the manual model entry: a toast `Model not loaded. Call loadModel() first.`

### Related source

`packages/stage-ui/src/stores/modules/speech.ts`

```ts
const configured = computed(() => {
  if (activeSpeechProvider.value === 'speech-noop') return false
  if (!activeSpeechProvider.value) return false
  let hasModel = !!activeSpeechModel.value
  let hasVoice = !!activeSpeechVoiceId.value
  return hasModel && hasVoice
})
```

### Environment

- AIRI hosted web build: https://airi.moeru.ai — **Version 0.12.0-beta.5 (dce185ba1d) · Branch main · Commit dce185ba1d · Built on 2026-09-20T16:26:55.766Z**
- Chat provider: DeepSeek (`deepseek-flash`) · Speech provider: Kokoro TTS (Local)
- Chrome, desktop (macOS) — reproduced
- Also reproduced in headless Chromium on Linux with a completely fresh profile
- No failed requests observed while the catalog loads; in one older session `providerModels`/`filteredModels` *were* populated with all 6 models, in a fresh one they were empty → looks state/timing dependent rather than network dependent

### Notes

- Same failure mode as #1056 (`Deepgram TTS Provider Missing listModels - Cannot Select Model`).
- Possibly related: #2273 (guard `kokoro-local` provider init timing on direct navigation), #2525 (load speech provider voices after configuration updates), #2559 (v2 provider catalog ignores `isAvailableBy`).
- The TTS engine itself is fine: with the workaround, Kokoro (`q8`, voice `af_heart`) synthesizes audio correctly.

lorenzozanee:
## Description

On fresh navigation, Kokoro local model and voice catalogs can be requested before the provider configuration exists. The empty catalogs prevent the local speech provider from being configured, so chat replies are not spoken.

Initialize credential-free providers before loading their model catalog and voice catalog. Use the provider default configuration and requested model for voice catalog requests while preserving existing credential-bearing provider behavior.

## Linked Issues

Fixes #2626

## Additional Context

The focused regression coverage exercises both model catalog initialization and voice catalog configuration.

## Test plan

- [x] `./node_modules/.bin/vitest --config packages/stage-ui/vitest.config.ts packages/stage-ui/src/stores/providers/provider.test.ts packages/stage-ui/src/stores/modules/speech.test.ts -t 'credential-free|local'`
- [x] `./node_modules/.bin/vitest --config packages/stage-ui/vitest.config.ts packages/stage-ui/src/stores/modules/speech.test.ts`
- [ ] TypeScript check: blocked in this checkout by missing Vue module files.


leaft:
The full `stage-ui` regression has four failed tests.
1. `official-provider` has `requiresCredentials: false`. The new initialization branch therefore runs for this provider, not only for Kokoro. `fetchModelsForProvider()` awaits initialization before it sets `modelStatus`, so the test reads `undefined` instead of `loading`.
2. The speech store starts with `speech-noop`. The unconditional initialization in `loadVoicesForProvider()` makes the startup watcher create an unnecessary provider record.
3. The official-default flow already initializes `official-provider-speech`. Its speech watcher initializes the same provider again, which increases the call count from four to six.
4. The Voicevox configuration contains a nested reactive `voiceSettings` object. The code spreads the configuration before `toRaw()`, so the nested value remains a Vue proxy. `structuredClone()` rejects this proxy, and the Voicevox catalog remains empty.
PR test command skips these existing tests.

